PR #39, opened 2012-03-14
Git-plugin: Add a "describe style" giving a relative commit name

ColinHebert:
Add a way to get a branch name with approximation.

With that, you can get branch names such as master~3 or origin^2~2 instead of a commit number.

### DIFF
--- a/plugins/git/functions/git-info
+++ b/plugins/git/functions/git-info
@@ -145,6 +145,10 @@ function git-info {
   local deleted=0
   local deleted_format
   local deleted_formatted
+  local describe
+  local describe_cmd
+  local describe_format
+  local describe_formatted
   local dirty=0
   local dirty_format
   local dirty_formatted
@@ -212,6 +216,9 @@ function git-info {
 
   # Gets the commit difference counts between local and remote.
   ahead_and_behind_cmd='git rev-list --count --left-right HEAD...@{upstream}'
+
+  # Use describe for approximative branch naming
+  describe_cmd='git describe --contains --all HEAD'
 
   # Ignore submodule status.
   zstyle -b \
@@ -286,6 +293,13 @@ function git-info {
         zformat -f behind_formatted "$behind_format" "B:$behind"
       fi
     fi
+  else
+    # Format describe.
+    describe=$(${(z)describe_cmd} 2>/dev/null)
+    if [[ -n "$describe" ]]; then
+      zstyle -s ':omz:plugin:git:prompt' describe 'describe_format'
+      zformat -f describe_formatted "$describe_format" "i:$describe"
+    fi
   fi
 
   # Format added.
@@ -349,6 +363,7 @@ function git-info {
       "c:$commit_formatted" \
       "d:$deleted_formatted" \
       "D:$dirty_formatted" \
+      "i:$describe_formatted" \
       "m:$modified_formatted" \
       "R:$remote_formatted" \
       "r:$renamed_formatted" \
--- a/plugins/git/style.zsh
+++ b/plugins/git/style.zsh
@@ -29,6 +29,9 @@ zstyle ':omz:plugin:git:prompt' deleted 'deleted:%d'
 # %D - Indicator to notify of dirty files.
 zstyle ':omz:plugin:git:prompt' dirty 'dirty:%D'
 
+# %i - Closest branch/tag to the commit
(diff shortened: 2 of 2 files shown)


unphased:
I often need to refer the git commit hash (and a friendly relative branch/tag name, although very friendly, is too fuzzy/unclear). I've been trying to grok `prompt_sorin_setup` for a while, but making very little progress. I cannot tell how to modify this to force the `RPROMPT` to always show the commit hash.

belak:
@unphased Yeah, it's a little convoluted. I'll try to give an overview here:

This portion goes over the formatting for git-info:
```
  # Set git-info parameters.
  zstyle ':prezto:module:git:info' verbose 'yes'
  zstyle ':prezto:module:git:info:action' format '%F{7}:%f%%B%F{9}%s%f%%b'
  zstyle ':prezto:module:git:info:added' format ' %%B%F{2}✚%f%%b'
  zstyle ':prezto:module:git:info:ahead' format ' %%B%F{13}⬆%f%%b'
  zstyle ':prezto:module:git:info:behind' format ' %%B%F{13}⬇%f%%b'
  zstyle ':prezto:module:git:info:branch' format ' %%B%F{2}%b%f%%b'
  zstyle ':prezto:module:git:info:commit' format ' %%B%F{3}%.7c%f%%b'
  zstyle ':prezto:module:git:info:deleted' format ' %%B%F{1}✖%f%%b'
  zstyle ':prezto:module:git:info:modified' format ' %%B%F{4}✱%f%%b'
  zstyle ':prezto:module:git:info:position' format ' %%B%F{13}%p%f%%b'
  zstyle ':prezto:module:git:info:renamed' format ' %%B%F{5}➜%f%%b'
  zstyle ':prezto:module:git:info:stashed' format ' %%B%F{6}✭%f%%b'
  zstyle ':prezto:module:git:info:unmerged' format ' %%B%F{3}═%f%%b'
  zstyle ':prezto:module:git:info:untracked' format ' %%B%F{7}◼%f%%b'
```

The bottom-most setting is the overall display:
```
```

Each of those keys can be looked up in https://github.com/sorin-ionescu/prezto/blob/master/modules/git/functions/git-info#L408 where we use this format string to include the actual data. The important thing to note is the 3 space-separated items at the start of the info format (followed by a :). These correspond to branch, position, and commit respectively.

We run git-info in an async task as seen here and use it to print out the overall status:
```
function prompt_sorin_async_git {
  cd -q "$1"
  if (( $+functions[git-info] )); then
    git-info
    print ${git_info[status]}
  fi
}
```

Finally, we parse this in the following by splitting the output into two fields on the `:`:
```
function prompt_sorin_async_callback {
  case $1 in
    prompt_sorin_async_git)
      # We can safely split on ':' because it isn't allowed in ref names.
      IFS=':' read _git_target _git_post_target <<<"$3"

      # The target actually contains 3 space separated possibilities, so we need to
      # make sure we grab the first one.
      _git_target=$(coalesce ${(@)${(z)_git_target}})

      if [[ -z "$_git_target" ]]; then
        # No git target detected, flush the git fragment and redisplay the prompt.
        if [[ -n "$_prompt_sorin_git" ]]; then
          _prompt_sorin_git=''
          zle && zle reset-prompt
        fi
      else
        # Git target detected, update the git fragment and redisplay the prompt.
        _prompt_sorin_git="${_git_target}${_git_post_target}"
        zle && zle reset-prompt
      fi
      ;;
  esac
}
```

There's some additional magic where we use coalesce which takes space separated arguments (which were the branch name, position, and commit) and returns the first one that's not empty:
```
      # The target actually contains 3 space separated possibilities, so we need to
      # make sure we grab the first one.
      _git_target=$(coalesce ${(@)${(z)_git_target}})
```

So, long story short:

* Remove the line which calls coalesce since you only want to return one value.
* Change the info format to the following (removing `%b %p` from the first part):
  ```
      'status' '%c:%s%A%B%S%a%d%m%r%U%u'
  ```

That was probably more than you ever wanted to know, but I hope it helps.

unphased:
Omg I love you

diraol:
@belak this should go to a file and/or a page!!! Pretty awesome explanation!!!!

unphased:
@belak Thank you for the explanation. I was able to use it to very quickly implement what I wanted, which was to move the `%c` from before the colon to after it, and it looks perfect.

PR #271, opened 2012-09-07
Use the init.zsh parent folder as the PREZTO_HOME folder

ColinHebert:
Instead of hardcoding `.prezto` and using that multiple time, we can easily say that the prezto folder is the one where init.zsh was present.

### DIFF
--- a/helper.zsh
+++ b/helper.zsh
@@ -40,7 +40,7 @@ function pmodload {
   pmodules=("$argv[@]")
 
   # Add functions to $fpath.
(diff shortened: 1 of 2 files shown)


marcelo-mason:
I want to be able to have he same prezsto config when I go down to root `sudo -s`.  How can I do that when this forces install onto a user home?

marcelo-mason:
My use case:  I want to have the same prezto setup when I `sudo -s`.  Currently I am unable as presto installs itself on a user's home folder.

PR #310, opened 2012-10-10
make creation of git aliases optional

kljohann:
As I do not use the git aliases provided by the git module but do not want to miss out on the defined functions I propose aliases to be optional (on by default). Is `:prezto:module:git` → `aliases` okay?


pushrax:
@sorin-ionescu is there a particularly compelling reason not to have this?

sorin-ionescu:
No.

PR #407, opened 2013-03-28
DNF first support

firewavemtl:
Just a copycat of yum module to dnf module. Mainly usefull for Fedora 19


sorin-ionescu:
Commited.

PR #413, opened 2013-04-08
Added a new module for 'grc' that defines some aliases for some major co...

lancespeelmon:
...lor-coded output goodness.


sorin-ionescu:
[lnav](http://lnav.org/features/) is a good alternative for log highlighting.

PR #462, opened 2013-07-24
Fixed slow performance with `alias git=hub`

jozefizso:
Prefixed all git calls with `command` so `alias git=hub` won't affect shell performance.

### DIFF
--- a/modules/git/functions/_git-hub-browse
+++ b/modules/git/functions/_git-hub-browse
@@ -8,7 +8,7 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
-if ! is-true "$(git rev-parse --is-inside-work-tree 2> /dev/null)"; then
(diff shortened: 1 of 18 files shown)


devinrhode2:
hey @sorin-ionescu it seems that the 'new hub' is actually the same github/hub repository, still hub on homebrew. At the bottom of the readme on github/hub you can see "hub in ruby" (previous implementation) - also the latest commits (even by the creator of gh) are on the github/hub repo.

(also, sorry I commented on the wrong issue :/ )

sorin-ionescu:
We can replace `git` with `command git` in Prezto. The last remaining code is the Git completion, which is already slow in Zsh and won't be fixed because the Zsh developers care more about correctness over usability and making completion go through `hub` will slow it down even if it's now compiled.

PaBLoX-CL:
Sorry, what means correctnes over usability in this context?

sorin-ionescu:
I don't remember. Google 'slow Zsh Git completion'.

PR #473, opened 2013-08-25
Fix non-working alias gSu

leoj3n:
I'm guessing the current alias `alias gSu='git submodule foreach git pull origin master'` is from the early days of git? The command doesn't work for me, but my proposed change to `alias gSu='git submodule update'` does. It's also possible I've misinterpreted what `gSu` is supposed to do and am using it wrong (hopefully not!)

### DIFF
--- a/modules/git/alias.zsh
+++ b/modules/git/alias.zsh
@@ -165,7 +165,7 @@ alias gSI='git submodule update --init --recursive'
 alias gSl='git submodule status'
 alias gSm='git-submodule-move'
 alias gSs='git submodule sync'
-alias gSu='git submodule foreach git pull origin master'
+alias gSu='git submodule update'
 alias gSx='git-submodule-remove'
 
 # Working Copy (w)


leoj3n:
Submodules on a freshly cloned repo don't respond to `gSu` because they track the SHA:

http://stackoverflow.com/questions/20794979/git-submodule-is-in-detached-head-state-after-cloning-and-submodule-update

`git submodule update --remote` succeeds in this case but checks out the SHA again, ruining any submodule branches manually setup to track origin, for example like this:

http://stackoverflow.com/questions/8642668/how-to-make-submodule-with-detached-head-to-be-attached-to-actual-head

Do we add an alias or create a happy medium? Both of these commands are useful in their respective contexts but don't work as expected in a repo with mixed submodule setup (i.e. some tracking origin and some referencing SHA).

sorin-ionescu:
What's the advantage to checking out the branch instead of leaving it as a
detached head?

leoj3n:
For reference, `branch =` in `.gitmodules` is how I should lock down a branch, [as mentioned here](http://stackoverflow.com/a/18797720).

leoj3n:
I am going to necro this issue, because I continue to believe the `--remote` feature, added in Git 1.8.2, is the correct way of updating submodules.

http://stackoverflow.com/a/21195182

~~~
--remote

This option is only valid for the update command. Instead of using the superproject’s recorded SHA-1 to update the submodule, use the status of the submodule’s remote-tracking branch.
~~~

> This is equivalent to running git pull in each submodule, which is generally exactly what you want.

Because this is how Git wants us to do it since version 1.8.2, and because it has the benefit of updating the superproject’s recorded SHA-1, I will continue to argue this is the correct and expected behavior in contrast to manually executing `git pull` for each submodule repository.

PR #517, opened 2013-12-20
Automatic virtualenv activation and deactivation, virtualenv non lazy mode.

jck:
chpwd hook for automatically activating and deactivating virtualenvs, and added option to load virtualenvwrapper normally(as opposed to lazy).

more info in README.md :)


jck:
I'm deleting this pull request to pursue a more general alternative such as zsh-autoenv.

PR #533, opened 2014-01-12
Additional verbose flag displays upstream branch for git-branch

goneflyin:
As of git 1.8.3:

> -v, -vv, --verbose
> When in list mode, show sha1 and commit subject line for each head, along with relationship to upstream branch (if any). If given
> twice, print the name of the upstream branch, as well (see also git remote show <remote>).


belak:
Thanks for the idea! `gbv` and `gbV` were recently added and it looks like these mostly do what was intended here... I'd rather not change existing aliases, so I'm closing this for now.

PR #552, opened 2014-02-08
git-hub-shorten-url: Support shorthand and code

leoj3n:
My proposed changes enhance the `git-hub-shorten-url` command so it can be used like:

```
~ ❯❯❯ git-hub-shorten-url sorin-ionescu/prezto instantly-awesome-zsh
http://git.io/instantly-awesome-zsh
```

:+1:

### DIFF
--- a/modules/git/functions/git-hub-shorten-url
+++ b/modules/git/functions/git-hub-shorten-url
@@ -2,21 +2,29 @@
 # Shortens GitHub URLs.
 #
 # Authors:
+#   Joel Kuzmarski <leoj3n@gmail.com>
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
 local url="$1"
 
-if [[ "$url" == '-' ]]; then
-  read url <&0
-fi
-
 if [[ -z "$url" ]]; then
-  print "usage: $0 [ url | - ]" >&2
+  print "usage: $0 [ url | - ] [slug]" >&2
+  return 1
+elif [[ "$url" == '-' ]]; then
+  read url <&0
+elif [[ "$url" != *github.com* ]]; then
+  url="https://github.com/$url"
 fi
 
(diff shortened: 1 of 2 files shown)


devinrhode2:
hey @sorin-ionescu it seems that the 'new hub' is actually the same github/hub repository, still hub on homebrew. At the bottom of the readme on github/hub you can see "hub in ruby" (previous implementation) - also the latest commits (even by the creator of gh) are on the github/hub repo. Hey, my comment is almost one year since the last comment.

Regarding the word "slug" for urls - I learned it before I knew how to code thanks to wordpress.

modules/git/functions/git-hub-shorten-url
```diff
@@ -2,21 +2,29 @@
...
+  return 1
+elif [[ "$url" == '-' ]]; then
+  read url <&0
+elif [[ "$url" != *github.com* ]]; then
+  url="https://github.com/$url"
 fi
 
 if (( $+commands[curl] )); then
-  curl -s -i 'http://git.io' -F "url=$url" | grep 'Location:' | sed 's/Location: //'
+  local result="$(curl -s -i 'http://git.io' -F "url=$url" -F "code=$2")"
```
belak: This should also probably use https

belak:
I merged a smaller PR which also modifies this so there are conflicts now. Sorry about that.

leoj3n:
> Sorry, we're no longer accepting new links to shorten.

belak:
Thanks for the update! I'll guess if there are no new links, we might as well drop this function.

leoj3n:
> Thanks for the update! I'll guess if there are no new links, we might as well drop this function.

Shortening a URL (in general) is a nice feature, but probably not something needed on a daily basis.

I would axe it unless you think it's a good example that can be recreated using a still functioning service.

PR #598, opened 2014-04-27
Add support for skipping VIRTUALENVWRAPPER init in modules/python

eukaryote:
This branch just changes modules/python/init.zsh so that it will recognize if the user defined a PREZTO_PYTHON_NO_VIRTUALENVWRAPPER var and will not do the virtualenvwrapper setup and sourcing of virtualenvwrapper_lazy.sh in that case.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -38,7 +38,7 @@ fi
 zstyle -t ':prezto:module:python' skip-virtualenvwrapper-init
 if (( $? && $+commands[virtualenvwrapper.sh] )); then
   # Set the directory where virtual environments are stored.
-  export WORKON_HOME="$HOME/.virtualenvs"
+  export WORKON_HOME="${WORKON_HOME:-$HOME/.virtualenvs}"
 
   # Disable the virtualenv prompt.
   VIRTUAL_ENV_DISABLE_PROMPT=1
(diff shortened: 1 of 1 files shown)


eukaryote:
Bump. Please merge.

mrmachine:
I get bitten by this every time I reset local modifications to prezto, and I am unable to open a new terminal window as a result. Normally I just have this line commented out as a local modification.

The problem is that when using pyenv, you must load virtualenvwrapper with `pyenv virtualenvwrapper` instead of `source "$commands[virtualenvwrapper.sh]"`.

Please merge this so I can configure prezto to skip this step.

N4M3Z:
If I understand correctly, the problem is caused by the fact that `pyenv` registers the `virtualenvwrapper.sh` command through its shims which causes `prezto/modules/python` to automatically detect it and attempt to source it. As far as I can tell, calling `source virtualenvwrapper.sh` on my system crashes the shell if `pyenv` was loaded beforehand. Consequently, isn't the following the easiest solution to the problem?
```
  if (( $+commands[pyenv] )); then
    pyenv virtualenvwrapper
  else
    source "$commands[virtualenvwrapper.sh]"
  fi
```
Also, if using `pyenv` the following enviroment variable should be set:
```
export PYENV_VIRTUALENVWRAPPER_PREFER_PYVENV="true"
```

belak:
This seems really useful. I'll try to take another look at this later today.

belak:
Thanks for your contribution! Sorry it took so long to merge.

PR #599, opened 2014-04-29
Terminal titling improvements

oohlaf:
Simplifies titling functions.

Multiplexer titles can be set independent of window and tab titles.
Adds auto-title always option to have dynamic titling inside terminal multiplexers.
Setting auto-title to yes continues to ignore dynamic titling inside terminal multiplexers.
This patch should not alter behavior under Apple terminals.

### DIFF
--- a/modules/terminal/README.md
+++ b/modules/terminal/README.md
@@ -13,14 +13,19 @@ directory, add the following to *zpreztorc*:
 
     zstyle ':prezto:module:terminal' auto-title 'yes'
 
-Auto titling is disabled inside terminal multiplexers, except inside dvtm, since
-it interferes with window names defined in configuration files and profile
-managers.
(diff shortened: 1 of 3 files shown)


oohlaf:
Hi Sorin,

I squashed my commits to a single commit and rebased against current master. Please accept the pull as it simplifies the titling by separating window, tab and multiplexer titles.

Cheers,
Olaf

sorin-ionescu:
Thank you. I will let you know what I do with it.

dchenbecker:
This looks like it would fix the issue that I opened up #807 for

tjharman:
Thanks, I'd love to see this in the base install.

belak:
Thanks for your contribution! This seems like a nice improvement over the current behavior.

PR #638, opened 2014-07-13
git performance improvement if aliased

teemow:
If git is aliased to hub the performance of the git module was very poor. The prompt was very laggy. 

I replaced all occurences of `git` with `command git` and used it for a while now. Performance is back to normal and all git infos are displayed nicely.

### DIFF
--- a/modules/git/functions/_git-hub-browse
+++ b/modules/git/functions/_git-hub-browse
@@ -8,7 +8,7 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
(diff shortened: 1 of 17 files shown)


ianks:
Slightly off topic, but if you install hub from --HEAD performance is much better as it is written in Go.

sorin-ionescu:
This is a duplicate of #462.

belak:
This still seems like a reasonable change. Is there any reason to avoid merging this @sorin-ionescu ?

belak:
Unfortunately, there are pretty substantial conflicts in this now. If you'd still like this to be merged, please re-submit it rebased off of the master branch.

PR #659, opened 2014-08-23
Allow to load prezto modules from an alternate path

piec:
Hello guys, this change allows you to load prezto modules from a user-specified directory.
It is useful to me when I create modules that I don't want to commit to the prezto repository because they are too specific to be useful to anyone or because they contain confidential data.
It can be used like that

```
zstyle ':prezto:load' user_pmodules_path "$HOME/.zsh/modules"
zstyle ':prezto:load' user_pmodule \
  'mymodule'
```

I have considered making `pmodload` use a search path variable (like $path or $fpath) but it makes things a bit more complicated. What do you think?
Cheers

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -23,15 +23,22 @@ unset min_zsh_version
 
 # Loads Prezto modules.
 function pmodload {
+  pmodload_path "${ZDOTDIR:-$HOME}/.zprezto/modules" "$argv[@]"
+}
+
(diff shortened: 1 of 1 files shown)


piec:
OK so in the end it was implemented in prezto! https://github.com/sorin-ionescu/prezto/commit/ce349dff81c6ca1f636c3ca9118d4d60b56617cd :)

PR #664, opened 2014-08-29
Upstream/issue/663 - update syntax highlighting submodule

chauncey-garrett:


ParthKolekar:
@chauncey-garrett actually... It lists `4.3.11+`. We can lower it to `4.3.11`. Because that is the maximum version that CentOS 6 has got. 

While it is true that CentOS6 needs to be purged, surely it can't hurt to ease it off in style with fancy zsh and colours?

ParthKolekar:
I found this. https://github.com/zsh-users/zsh-syntax-highlighting/commit/a4196eda5e6fea29bf3cc70a17bf3bd2dd2eea95

Apparently the compatibility broke, but now it is back. 

Can we revert to support of 4.3.11 since it works absolutely peachy with my setup.

ParthKolekar:
I created #1491 for this.

chauncey-garrett:
At the time, requiring v4.3.17 was for the syntax highlighting module so seems fine to me to allow v4.3.11 as long as the syntax highlighting module is up to date and there aren’t other modules that require it. 

Glad they’ve fixed the regression.

PR #669, opened 2014-09-02
Python virtualenvwrapper workon crashing when tabbing.

hmorales:
Mac 10.9.4 using iTerm2 

When in a new terminal session tabbing to list virtualevn's using workon crashes the session. Using the same fix found on the oh-my-zsh thread below fixed the issue for me.

https://github.com/robbyrussell/oh-my-zsh/issues/2355

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -35,14 +35,14 @@ if (( ! $+commands[python] && ! $+commands[pyenv] )); then
 fi
 
 # Load virtualenvwrapper into the shell session.
-if (( $+commands[virtualenvwrapper_lazy.sh] )); then
+if (( $+commands[virtualenvwrapper.sh] )); then
   # Set the directory where virtual environments are stored.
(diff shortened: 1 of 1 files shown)


jeffwidman:
Looks like https://github.com/sorin-ionescu/prezto/commit/2794f95d3eb7d5b65be56710d9d9e9ac9d642ff8 reverted the fix for this. Which may be fine if upstream has since fixed the bug.

@belak did you check into this before merging the lazy version?

belak:
I had no idea this bug existed. I'll double check in a bit.

jeffwidman:
> I had no idea this bug existed. I'll double check in a bit.

@belak Did you ever get a chance to test this?

I haven't observed any issues so far on macOS / iTerm2 both latest stable versions.

belak:
I didn't try very hard, but I wasn't able to reproduce it either.

jeffwidman:
Cool, then if neither of us can repro, we are good for now.

PR #715, opened 2014-11-03
Port the oh-my-zsh gitignore plugin to prezto

1989car:
The oh-my-zsh gitignore plugin can get gitignore template from [gitignore.io](http://gitignore.io)  .  It may be usefull for developers.

As the gitignore.io website says, they hosting 191 Operating System, IDE, and Programming Language .gitignore templates.

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -237,6 +237,7 @@ Functions
   - `git-stash-recover` recovers given dropped stashed states.
   - `git-submodule-move` moves a submodule.
   - `git-submodule-remove` removes a submodule.
+  - `git-ignore-template` get gitignore template from [gitignore.io][9]
 
 Theming
 -------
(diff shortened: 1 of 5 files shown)


belak:
I'm fairly hesitant to add this, as the feature is built into github, this depends on an external service, and there are gitignore templates available in https://github.com/github/gitignore.

Does anyone else have opinions on this?

ezintz:
From my perspective this is not needed. I anyways to not like all the Git aliases that are already there. So to get this feature in my prezto/cli environment I would go with a different solution, also regarding to the docs of gitignore.io:

`git config --global alias.ignore '!gi() { curl -L -s https://www.gitignore.io/api/$@ ;}; gi'`

belak:
Thanks for submitting this, but I'm closing this for now as I'm not sure it's a good fit for prezto right now.

belak:
I just realized we also do something similar with github's url shortener... so this may actually be a decent fit... unfortunately it looks like the repo is gone, so I can't reopen this.

PR #734, opened 2014-11-26
Restart Zsh without opening new window or tab

christianbundy:
It seems that you could just use `exec $SHELL` instead of manually opening a new system or tab. Is there any reason not to do this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ version is 4.3.17.
 
   5. Restart Zsh:
   
-        exec $SHELL -li
+        exec $commands[zsh]
 
 ### Troubleshooting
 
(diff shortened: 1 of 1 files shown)


christianbundy:
Just made the requested change – let me know if there's anything else you need from me to merge. Thanks!

christianbundy:
Sorry for the delay on this, I just switched over to the `$commands[zsh]` method you mentioned in #1034. Would you like me to squash, or is there anything else I need to do before merging?

belak:
Sorry for the delay in responding to this. For now, I'd like to keep the existing description. It's a bit simpler, less error prone, and should work even when using bash initially.

PR #740, opened 2014-12-01
Fix git status parsing

eugenk:
The previous behaviour only checked for assumed possible states,
which is not a safe way to go. The behaviour of the git client might
change in what combinations are possible.

The previous behaviour missed some "deleted" lines in `git status`.

With the new behaviour, all combinations one can think of are checked.
This way, also the missed "deleted" lines are caught.


belak:
What's the status on this? Should it be merged or dropped?

belak:
There hasn't been an updated on this in 6 years so I'm closing this for now. Please feel free to re-open (or comment) if this is still an issue. Maybe it's as simple as updating the docs?

PR #749, opened 2014-12-04
Add external module bd.

eugenk:
Since #741 was rejected because it was not a submodule, I made it a submodule now.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "modules/prompt/functions/pure"]
 	path = modules/prompt/external/pure
 	url = https://github.com/sindresorhus/pure.git
+[submodule "modules/bd/external"]
+	path = modules/bd/external
+	url = https://github.com/Tarrasch/zsh-bd.git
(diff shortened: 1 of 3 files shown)


eugenk:
Any new opinions on this pull request?

sorin-ionescu:
Apparently not.

eugenk:
What about the old opinions? You said, you wanted this to be an external module (which it is now) and that you don't see much use in this. However, there were some +1 comments in the old pull request, telling what the use is. Are you still against `bd`?

zmwangx:
@eugenk I find `bd` pretty useful, thanks for the good work. But honestly, one shouldn't argue with the maintainer. Folks can merge this into their own forks if they like it; for instance, I just merged this PR into my own fork: https://github.com/zmwangx/prezto/commit/a50d4d35f925f3f7566b7b5fe8f9aea822b4e43b.

eugenk:
Don't get me wrong, I don't want to argue. Sorry if I came off that way. There was just no comment on this pull request for a long time and it made me believe that there's something left to do.

If bd is not useful enough to get merged into the main repository, I completely understand. However, I don't see any reason for this pull request to stay open for this long, if there is no chance for a merge.

belak:

PR #761, opened 2014-12-15
Allow directories customizations

alem0lars:
This is a simple change.
I've just added the possibility to specify the following paths:
- `PREZTO_DIR`: The prezto directory
- `PREZTORC`: The path for the prezto configuration file

**This allows the user to use custom locations, which is really useful sometimes...**

I've tested the changes, it will work without breaking compatibility. By default, the paths are the current values, so users can just ignore the change if they aren't pro-users and don't want custom locations.

Thanks

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -6,6 +6,15 @@
 #
 
 #
+# Configuration
+#
+
+# Set PREZTO_DIR to the default value if not already set
+if [ -z "${PREZTO_DIR}" ]; then
+  PREZTO_DIR=${ZDOTDIR:-$HOME}/.zprezto
+fi
+
+#
 # Version Check
 #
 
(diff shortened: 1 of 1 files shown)


alem0lars:
Even if Prezto is meant to be installed in the home directory I don't understand why this isn't better since it cleans up *a lot* the codebase without repeating strings everywhere...

Thinks, it's like if in your program you prefer repeating everywhere a constant instead of declaring it...

It doesn't really makes any sense

kaluzki:
+1

ispmarin:
:+1:

belak:
A portion of this was merged. prezto should be able to be loaded from wherever it's cloned to as of #1264. This now conflicts with that PR, so if you update this, we can look into merging it.

belak:
I'm going to close this for now. Please feel free to open another PR if you would like to get the preztorc changes in.

PR #762, opened 2014-12-15
fix(module/ssh): Fix error when enable module ssh

Marthym:
$USER is replace by 'id -un' to avoid error due to the $USER no set

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -25,7 +25,7 @@ if [[ ! -S "$SSH_AUTH_SOCK" ]]; then
   source "$_ssh_agent_env" 2> /dev/null
 
   # Start ssh-agent if not started.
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
`$USER` is used in a few places, and truthfully, neither `$USER` nor `$LOGNAME` can be trusted since they can be arbitrarily changed. `$(id -un)` and `$(logname)` give the correct answer.

Having said that, instead of changing `$USER` to `$LOGNAME` in various places, a better approach, in your case, is to `export USER="${USER:-$LOGNAME}`" in *zprofile*.

sorin-ionescu:
Fixed in 88aee30.

PaBLoX-CL:
Why did you decided to go anyway with `$LOGNAME`?

sorin-ionescu:
`$USER` is deprecated, at least on Mac OS X. `$(id -un)` is better, but it's a process call, which makes it slower.

PR #764, opened 2014-12-18
Added transfer module. Allows command line file sharing.

nl5887:
Usage:
- `transfer file.txt` will transfer a file and return a link
- `transfer directory` will zip the directory, transfer the zip and return a link
- `cat file.log | transfer file.log` will transfer the piped output and return a link

### DIFF
--- a/modules/transfer/README.md
+++ b/modules/transfer/README.md
@@ -1,0 +1,14 @@
+Transfer.sh
+==========
+
+Defines [transfer][https://transfer.sh/] aliases for easy file sharing from the commandline.
+
+Authors
+-------
+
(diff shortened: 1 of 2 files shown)


sorin-ionescu:
Yes, the coding style is different. Is transfer.sh popular? Does anybody actually use it?

nasenatmer:
Never heard of it, I was usually using some CLI programmes to access paste bins, but as this allows to upload whole files or even directories it seems quite useful.

belak:

I'm closing this for now because I think it's a little too specialized for prezto at the moment but thank-you for taking the time to submit this.

PR #772, opened 2015-01-05
allow sourcing zshrc.local config

rothgar:
Allowing .local files to be sourced will allow users to customize prezto without needing to maintain a fork and making sure they are current with upstream. This way prezto can be cloned and custom aliases can remain in users own dotfiles repos if desired.

Of course, they can still fork and update zshrc as before, but this will allow a default option for users who do not want to.

### DIFF
--- a/runcoms/zshrc
+++ b/runcoms/zshrc
@@ -11,3 +11,7 @@ if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
 fi
 
 # Customize to your needs...
+# Source custom configuration
+if [[ -s "${ZDOTDIR:-$HOME}/zshrc.local" ]]; then
+  source "${ZDOTDIR:-$HOME}/zshrc.local"
+fi
--- a/runcoms/zshrc
+++ b/runcoms/zshrc
@@ -10,8 +10,9 @@ if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
   source "${ZDOTDIR:-$HOME}/.zprezto/init.zsh"
 fi
 
-# Customize to your needs...
-# Source custom configuration
+# Source custom configuration if present
 if [[ -s "${ZDOTDIR:-$HOME}/zshrc.local" ]]; then
   source "${ZDOTDIR:-$HOME}/zshrc.local"
 fi
+
+# Customize to your needs...


modules/completion/external
```diff
@@ -1 +1 @@
-Subproject commit 08afea0e230b5a321d584119f55b675990327ef9
```
sorin-ionescu: These should not be here.

runcoms/zshrc
```diff
@@ -11,3 +11,7 @@ if [[ -s "${ZDOTDIR:-$HOME}/.zprezto/init.zsh" ]]; then
 fi
 
 # Customize to your needs...
+# Source custom configuration
```
sorin-ionescu: Move these above the customize to your needs comment.

sorin-ionescu:
Actually, this is more useful for machine-local configuration that for what you suggest. It's much better to modify dotfiles than to override them. From my experience with Oh-My-Zsh's custom directory, overrides go out of sync.

rothgar:
I'm not sure how sourcing an external file is much different than modifying the existing zshrc file. In either case they may override prezto sourced aliases/functions. If modifying the dotfiles is recommended then we should be modifying the modules/**/*.zsh files instead.

Putting the customization in an external file allows for easier pulls from origin because you don't need a fork (and yet can still customize) and you don't have to worry about merge conflicts.

I would also maybe suggest a zpreztorc.local file so modules can be added removed as well but I haven't tested that yet.

I think I fixed the other submodule commits.

xuhdev:
I also find this useful. Avoiding fork is easier to maintain prezto configurations. We can do this for zshenv, zprofile and other configurations as well.

PaBLoX-CL:
I insist that the simpler/straightforward workflow with prezto is:

1. clone prezto
2. copy runcoms to `~/`
3. modify **your** zshfiles, useunalias, or override configurations (createa another glg alias for example)
4. update prezto (it will be fast forward with no conflicts)
5. create a git hook that launches diff tool any time a runcom is updated

Most times, there's **no** need in modifying prezto in place.

rothgar:
@PaBLoX-CL How do you track your copies/changes of runcoms files?  It means you either need to keep a fork of prezto (and track upstream changes) or you need to always pull changes and fix any conflicts. For less technical users I think that might be too much to ask them to manually run diff on any modified files any time they want to update prezto. They probably would just never update prezto

Keeping externally untracked files allows people to pull prezto and add a custom zshrc.local file for any customization they want.

We could add a zpreztorc.local file that they could modify as an alternative to the built in zpreztorc file. This would allow users to pull directly from source and track their customizations however they want (in prezto fork, personal dotfiles repo, or not at all).

leoj3n:
Sourcing .local files if they exist may be trivial but documenting and supporting that functionality is not. Black-boxing prezto adds an unnecessary level of indirection, issue noise, and maintenance overhead; proper APIs would necessitate a lot of complexity (think `/modules`). Shell scripts aren't meant to be that fancy, and if the API isn't robust enough it will be hard if not impossible for users of this method to track down breaking upstream changes. Perhaps a simpler solution is better documentations for the noobs.

PaBLoX-CL:
@rothgar I didn't say it was for beginners, I said it was simpler (as in the KISS principle). In my workflow, I don't find any "merge conflict", since I don't modify prezo in place, because I don't believe it's worth it. I don't use that workflow anymore, so I don't remember how I did it exactly. But I believe there's a way to find which files are being "fast-forward", and therefore, you can run a diff tool to see the changes. There's no need to run a diff tool everytime you update prezto, its a git hook that runs after a merge has happened.

Honestly, I do something even simpler now: I update prezto from time to time, when I see a runcom file has been updated, I run diff between the runcom and my local file. That's it :)

PR #775, opened 2015-01-16
added paradox prompt with ruby_info in rprompt

rocknrollMarc:
I added a new version of the new paradox Prompt as I need a ruby-info on the right, I added it so we have the time and ruby-info now.


sorin-ionescu:
Edit the existing one. There is no need to duplicate it.

rocknrollMarc:
ok

PR #776, opened 2015-01-19
Fixed 'topc' and 'topm' aliases

chirgjn:
The 'topc' an 'topm' aliases are working even when 'htop' is not installed

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -146,8 +146,8 @@ alias du='du -kh'
 if (( $+commands[htop] )); then
   alias top=htop
 else
-  alias topc='top -o cpu'
-  alias topm='top -o vsize'
+  alias topc='top -o %CPU'
+  alias topm='top -o %MEM'
 fi
 
 # Miscellaneous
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
Those aliases work for me on Mac OS X without changing to `%CPU` and `%MEM`.

chirgjn:
They don't work on Linux.
Does the new ones work on OS X?

sorin-ionescu:
They do not. LINUX may have a different `top` than BSD.

chirgjn:
Then an if statement to check the OS,  maybe?

sorin-ionescu:
That would be the best option. There is an example in that file.

chirgjn:
OK, then I will close this pull request then & create a new one with OS detection. Is that ok?

sorin-ionescu:
No. You can force push on top of an existing pull request.

chirgjn:
Ok

chirgjn:
This should work. Please confirm.

sorin-ionescu:
It should work for most users. I'm not sure about Solaris and other UNIXES.

chirgjn:
Pull it, if anyone has a problem they can raise an issue.

chirgjn:
Any other changes you want me to do?

PR #777, opened 2015-01-20
Node is not a requirement for nvm

### LINKED ISSUE #765: Problem in modules/node?
At line #18:

    if (( ! $+commands[node] )); then
      return 1
    fi

If `nvm alias default XYZ` has been run, node is added to PATH:

    ~ echo $+commands[node]                                                                                                                                                         
    0
    ~ source $(brew --prefix nvm)/nvm.sh
    ~ echo $+commands[node]
    1

Prezto's script sources NVM identically but exits early, instead of adding completions.

qur2:
Fixes #765.

### DIFF
--- a/modules/node/init.zsh
+++ b/modules/node/init.zsh
@@ -15,11 +15,6 @@ elif (( $+commands[brew] )) && [[ -d "$(brew --prefix nvm 2>/dev/null)" ]]; then
   source $(brew --prefix nvm)/nvm.sh
 fi
 
-# Return if requirements are not found.
-if (( ! $+commands[node] )); then
-  return 1
-fi
-
 # Load NPM completion.
 if (( $+commands[npm] )); then
   cache_file="${0:h}/cache.zsh"


sorin-ionescu:
Node.js is not a requirement for `nvm`, but is it not a requirement for `npm`?
Should not that check ought to be moved instead of removed?

ohcibi:
As @qur2 and I pointed out in #765 nvm does not add `node` to `$PATH` **unless**

* the user issues `nvm use <a-version>`, or
* a default node version is configured with nvm (e.g. `nvm alias default 0.10.35`)

This is not a problem at all and there is no need to load completions before `node` gets added to the `PATH` (i.e. this is not about the completions for `node`, `npm`).

The problem is:

There is the function `node-info` which by default usage gets issued whenever the prompt is generated (if the user wants to use that information in the prompt). When `node` is not in the `PATH` currently the node-module won't create this function, which results in an error "prompt_myrompt_precmd: command not found: node-info" for every new prompt. It is not clear why that happens as the expectation would be that if the node-module is included `node-info` should actually exist.

So the solution would be to create this function no matter what and

* populate `$node-info` with a version string like "no node available", or
* throw an error, clearly indicating that node in fact *is* a requirement for the node-module

PR #779, opened 2015-01-21
added paradox prompt with ruby_info in rprompt

rocknrollMarc:
Added ruby-info to the existing paradox prompt on the right (RPrompt)


sorin-ionescu:
I do not want to duplicate themes in Prezto. Themes can and ought to be configured.

rocknrollMarc:
ok how would I configure paradox to have nvm and rvm prompt would love to know? Ill do it like that then if its possible.

PR #607, opened 2014-05-09
Adding the ability to specify tmux session name.

guildencrantz:

PR #781, opened 2015-01-21
Option added to name the autocreated tmux session

kuetemeier:
Now you can use 

```
zstyle ':prezto:module:tmux:session' name 'your-session-name'
```

in `.zpreztorc` to name your default session. If this config option is not present, the name still defaults to 'prezto'. So default behavior is preserved.


sorin-ionescu:
This is a duplicate of #607.

belak:
I merged #607, so I'm closing this. Thanks for your contribution!

PR #783, opened 2015-01-22
Use `which` to find correct zsh path

christianbundy:
Some environments (like OS X using Homebrew) don't install applications to /usr/bin, so hardcoding the path to zsh is likely to result in the wrong version of zsh being saved as the default shell.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ version is 4.3.17.
 
   4. Set Zsh as your default shell:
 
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
Mac OS X comes with Zsh. Using Homebrew Zsh will fail unless it is added to */etc/shells*. Also, in Zsh, you can use `$commands[zsh]`.

Issue #656, opened 2014-08-17
Readd z


mpolden:
It seems like fasd is no longer maintained, as last commit was 2 years ago.

Would it be possible to readd z? (which is actively maintained)

VMatrix1900:
[z.lua](https://github.com/skywind3000/z.lua) is faster and more convenient than `fasd`. Maybe we can replace `fasd` with `z.lua`.

peterpme:
+1 for z

fikovnik:
There seems to be a more actively maintained fork https://github.com/whjvenyl/fasd

aliBenhenia:
nice code

aliBenhenia:
din

PR #786, opened 2015-01-28
Add rupa z to modules

Treggats:
I have added rupa/z to the modules, since it's a repo I use regularly.

I believe that the init.zsh I created is sufficient.


sorin-ionescu:
Duplicate of #656.

Issue #231, opened 2012-08-06
Add an installation script


sorin-ionescu:
A one line, `curl` based installation may increase adoption.

ebello:
+1 for this. I also had to use `-f` for the `ln` command in order for Prezto to work.

thallam:
+1

devinrhode2:
It was *strongly* disheartened after uninstalling OMZ at a whim and installing Prezto, only to see an extremely raw prompt:
```
devin%
```

FORTUNATELY I found [this answer](https://www.google.com/search?client=safari&rls=en&q=uninstall+prezto&ie=UTF-8&oe=UTF-8) by googling "uninstally prezto" which led me to this issue. `brew install prezto` and `brew uninstall prezto` would be a **stellar** improvement - of course prezto would need to be accessible via linux package managers too

sorin-ionescu:
Prezto is not in Homebrew. Prezto is meant to be used via Git, not package managers.

tombh:
Just to make it clear, the main issue people are having is *not* that lack of an installation script (whether that is through a package manager or a curl one liner). It is the simple fact that the current installation instructions in the README fail, if you *already have* an existing `.zshrc`. So in @devinrhode2's case, and anyone else coming from OMZ, then they will have a problem because `.zshrc` already exists.

The absolute most simple solution here is just a note in the README that `.zshrc` must not already exist. Or if it does, then the sourcing of `zprezto/init.sh` must be manually added.

PaBLoX-CL:
@tombh :+1:

monfresh:
+1 on adding a quick note in the README that you must move or rename your existing `.zshrc`. I just ran into this as well.

For a longer term solution, would it not be possible for Prezto to append the sourcing lines to any existing .zshrc?

pickfire:
+1 to `curl ... | sh`

MPNussbaum:
@devinrhode2 +1

al-the-x:
I see that PRs have been submitted (and declined) since #240 (circa 2012). See the more refined effort in #395. Are you interested in PRs for this issue, @sorin-ionescu? Happy to continue refining #395 or take another stab at it, if so.

sorin-ionescu:
Of coarse.

I have neglected to add an installation script to discourage beginners from using Prezto to prevent the issues list blowing up with dumb questions and even dumber pull requests in an effort to not replicate the Wild Wild West nature of Oh-My-Zsh.

[...]

PR #790, opened 2015-01-30
adds rudimentary install script

abak:
this PR creates and add an install script in the root folder.
This script then creates symlinks between runcoms/^README.md and the target folder (${ZDOTDIR:-$HOME}).

The main advantage is that prezto can be cloned anywhere, and not necessarily in ${ZDOTDIR:-$HOME}.


paulmelnikow:
Related: #231

PR #791, opened 2015-01-30
makes prezto and its init script location agnostic

abak:
This PR makes prezto location agnostic, so it can be installed anywhere, not just in .zprezto in $HOME or $ZDOTDIR
This is done be retrieveing the full path to init.zsh, and using the fact that the script is in the top level directory of the install, and by getting the full path to zshrc.


sorin-ionescu:
Duplicate of #271.

PR #798, opened 2015-02-09
Add WD module from oh-my-zsh to prezto

mtxr:
This modules is really helpful for me and I think more people would like to use it.

### DIFF
--- a/modules/wd/functions/wd
+++ b/modules/wd/functions/wd
@@ -1,0 +1,444 @@
+# WARP DIRECTORY
+# ==============
+# Jump to custom directories in terminal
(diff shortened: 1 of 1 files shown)


mtxr:
Please, include his README file.

https://github.com/mfaerevaag/wd

sorin-ionescu:
This must be implemented as an external module. See history-substring-search for an example.

PaBLoX-CL:
How is that this is more useful than using `fasd` and `pushd`/`popd`. I don't understand :/

sorin-ionescu:
I don't understand either.

Issue #579, opened 2014-03-24
Do not show ruby version (at ruby-info) on system ruby


wookayin:
After the commit dbeb6d4 (#361), the active ruby version is stored into `ruby_info[version]`.

This is a very cool feature when working across several ruby environments, but I suggest that this information should NOT be displayed when we're on the system ruby (/usr/bin/ruby for example).

This is a little annoying if one uses a prompt theme that displays the ruby version (`peepcode` from builtin or [my own customized one](https://github.com/wookayin/dotfiles/blob/master/zsh/prezto-themes/prompt_wook_setup), for example) The other version informations such as python is not displayed on the prompt if they are from system commands.

PR #802, opened 2015-02-14
Suppress node-info when nvm points system-wide node

wookayin:
On modern versions of nvm, when virtual environment is not activated
(i.e. node comes from system-wide and `nvm version` == 'system'),
redundant node-info could be suppressed from showing.

Example screenshot :

### DIFF
--- a/modules/node/functions/node-info
+++ b/modules/node/functions/node-info
@@ -17,7 +17,7 @@ if (( $+functions[nvm_version] )); then
   version="${$(nvm_version)#v}"
 fi
 
-if [[ "$version" != (none|) ]]; then
+if [[ "$version" != (none|system) ]]; then
   zstyle -s ':prezto:module:node:info:version' format 'version_format'
   zformat -f version_formatted "$version_format" "v:$version"
   node_info[version]="$version_formatted"


sorin-ionescu:
A similar discussion has occurred in #579 about #342 in respect to Ruby. What is system node? If it's just you on one machine, you may know, but if you handle many servers, or the the package manager on a  particular system allows you to install different versions, you do not know. As to why `nvm` shows **system**, which is pretty useless, as opposed to a specific node version, I do not know.

wookayin:
One wouldn't want to be notified redundantly if the node is from system-wide installation, not from `nvm`.

The purpose of `node-info` is, in my humble opinion and guess, to let users know which **virtual environment** provided **by nvm** (as nvm provides different versions as separate environments) version they are currently on, not the **actual version* itself. Frankly, I hardly get the picture when you mention different servers or different versions in a particular system, because in any cases `node` would be always located with respect to `$PATH` variable (such as /usr/bin/node, /usr/local/bin/node), but it doesn't care.

The reason why `nvm` shows `system` in the case, having a something common with that, would be to let one know whether the virtual environment is activated or not.

belak:
This seems like a reasonable change. The current version does not show any version information other than "system" and this is how other integrations work, so it makes sense to ignore that here. Thanks for your submission!

PR #806, opened 2015-02-15
Addded clarification to README. fixes #805

### LINKED ISSUE #805: Clarify globbing magic in README
I know I am probably just dumb...  but I tried to find, and could not.. what the following constructs are doing... (from the README)

`(.N)` and `:t`

    for rcfile in "${ZDOTDIR:-$HOME}"/.zprezto/runcoms/^README.md(.N); do
        ln -s "$rcfile" "${ZDOTDIR:-$HOME}/.${rcfile:t}"

I know this is just simple parameter expansion... but it would be great if these could be elaborated upon, and would provide a good example of what parameter expansion _IS_, and _how to use it_, lol

mralexgray:


sorin-ionescu:
The code will fail to execute when pasted into the terminal due to comments.

PR #807, opened 2015-02-16
Allow the `terminal` module auto-title for tmux

dchenbecker:
Allows the user to selectively enable auto window titles in
tmux. Defaults to not using window titles in tmux.

Because I use many, many windows in tmux, it's helpful to have the (tmux) window show he current path.

### DIFF
--- a/modules/terminal/init.zsh
+++ b/modules/terminal/init.zsh
@@ -74,7 +74,12 @@ function _terminal-set-titles-with-path {
   local truncated_path="${abbreviated_path/(#m)?(#c15,)/...${MATCH[-12,-1]}}"
   unset MATCH
 
-  set-window-title "$abbreviated_path"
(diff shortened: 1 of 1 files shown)


PaBLoX-CL:
Wouldn't make more sense to activate windows titles in tmux.conf? (or is it handled different?)

sorin-ionescu:
It makes more sense to use sessions and session management, which is what most people do; hence why it's disabled since it interferes with window titles set manually via configuration.

dchenbecker:
Perhaps I'm missing something simpler, but from the man page for tmux:

``` A pane's title is typically set by the program running inside the pane and is not modified by tmux.  It is the same mechanism used to set for example the xterm(1) window title in an X(7) window manager.```

I set the window name (and hence my terminal window title) in tmux.conf to be the host I'm on, but the _pane_ title I want to be the truncated path.

oohlaf:
Have a look at #599 which should solve this.

belak:
Closing in favor of #599

PR #810, opened 2015-02-18
Adding install.sh script

JackDanger:
Copying large commands out of a README is slightly more daunting for new
users (or was for me, anyway).


sorin-ionescu:
Duplicate of #790 and #231.

PR #811, opened 2015-02-18
remove double escape sequence

kamou:
Hi in my setup:
- fedora 21 
- byobu (TERM=screen) inside gnome-terminal (TERM=xterm-256color)

ControlLeft and ControlRight are: \e[D and \e[C
I might be wrong but I guess the \e\e escape sequence  should be replaced by a single \e escape sequence.

Cheers,
 Ayman

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -29,8 +29,8 @@ zmodload zsh/terminfo
 typeset -gA key_info
 key_info=(
   'Control'      '\C-'
-  'ControlLeft'  '\e[1;5D \e[5D \e\e[D \eOd'
-  'ControlRight' '\e[1;5C \e[5C \e\e[C \eOc'
+  'ControlLeft'  '\e[1;5D \e[5D \e[D \eOd'
+  'ControlRight' '\e[1;5C \e[5C \e[C \eOc'
   'Escape'       '\e'
   'Meta'         '\M-'
   'Backspace'    "^?"


sorin-ionescu:
Those were taken from Debian as far as I remember.

sorin-ionescu:
Change screen term to `screen-256color-bce`.

kamou:
-bce is not possible with my setup (byobu-tmux)
cat /usr/share/byobu/profiles/tmux
[snip]
# Cannot use:
#  - screen-bce, screen-256color-bce: tmux does not support bce
#  - screen-256color: vim broken without -bce
set -g default-terminal "screen"
[snip]

tried changing screen to screen-256color and it didn't work as well
this is also the case with a bare tmux.

I did the test with scree/byobu-screen and the codes become \e[1;5C and \e[1;5D so it is not a problem with those muxes.

I believe that you should either add \e[D and \e[C in the ControlLeft and ControlRight key_info
or just remove the double escape sequence.
I never saw these kind of codes. but again... I might be wrong :)

belak:
I wasn't able to reproduce this on ArchLinux with byobu 5.116, tmux 2.3, TERM=xterm-256color (in gnome-terminal) and screen-256color inside tmux... it also works with raw tmux. They appear to send \e[5D and \e[5C

I have to admit I've never seen `\e\e` in any instance and I can't find where this came from originally but I'd rather not merge this if I can't reproduce the issue and don't know where the original came from.

belak:
Thanks for submitting this! Unfortunately, as mentioned above I have not been able to reproduce it. If it's still an issue, please feel free to re-open this once the PR has been updated to work with master.

Issue #813, opened 2015-02-20
recursive diff


kontrafiktion:
`diff -r` does not work if colordiff is not installed

After migrating from oh-my-zsh to prezto the recursive diff command did not work anymore since the diff _function_ (https://github.com/sorin-ionescu/prezto/blob/master/modules/utility/functions/diff) mapped `diff` to `git --no-pager diff --color=auto --no-ext-diff --no-index "$@"` and that obviously does not support the `-r` switch.

Workaround: install "colordiff" (than `diff` is mapped to colordiff instead of `git-diff`)
Preferred solution: if `-r` is part of the arguments do not map to git-diff

sorin-ionescu:
`alias diffr='/usr/bin/diff -r'`

PaBLoX-CL:
I ran into the same issue, and I have to agree... it's quite annoying :/. I don't agree that alias is really a solution.

kontrafiktion:
I must admit that I was somewhat put off by @sorin-ionescu simply closing the issue. I would consider the alias a _workaround_
I would either expect a useful error message (instead of "fatal: invalid diff option/value: -r") or simply not switching to git-diff when `-r` is present (as in my pull request https://github.com/sorin-ionescu/prezto/issues/814) or a new command `cdiff` or `colordiff` but not "hijacking" an existing well-known command and breaking its behaviour

sorin-ionescu:
The use of git-diff is a workaround for missing colordiff that is good
enough for most uses.

Pull requests are welcome.

pickfire:
Maybe it is better to add a **DEPENDENCIES.md** to tell the users to install `colordiff`.

sorin-ionescu:
I disagree. Perhaps falling on `git-diff` should be removed since it differs from expected behavior.

pickfire:
Yeah, good idea. Since not everybody assumes that `diff` is `git-diff`.

cemeyer:
Please don't hijack diff like this.  I just spent 5 minutes trying to figure out why my `diff` command didn't match the manual page.  Very frustrating.  If colordiff takes the same options as normal diff, great.  But `git diff` does not.

belak:
@cemeyer You must be on an old version. This was changed almost 2 months ago for this very reason. See https://github.com/sorin-ionescu/prezto/commit/38f0d2f34e84ec3d02c589573a8e097d1807e762 for more info.

cemeyer:
Yes, must be.

[...]

PR #814, opened 2015-02-21
do not use git for recursive diff

kontrafiktion:
see https://github.com/sorin-ionescu/prezto/issues/813

I think this is a better solution than breaking existing code


sorin-ionescu:
There may be other switches that `git-diff` does not support. Adding an ugly hack just for `-r` is not good. Perhaps `git-diff` fallback should be removed instead.

PR #816, opened 2015-02-23
Fix parse error on init

aminb:
Fix parse error on init introduced in 933c61b, by adding missing parentheses.


aminb:
Oh, I just saw b41f485.

Closing this :)

PR #825, opened 2015-02-27
Update zsh-completions to 6cc78eb382a62ad99cb0c900d4ed623af9cee10a

postblue:


sorin-ionescu:
External code is updated periodically, and this module was updated a few days ago.

PR #828, opened 2015-03-01
Skip setting terminal title inside Emacs

jcf:
Hi Sorin,

Might have a quick fix for an Emacs + Prezto + auto-title hiccup I've encountered.

Setting Terminal title in Emacs results in beeping and garbled `2;<cmd>1;<cmd><result>` output, as you can see in the screenshot below.

multi-term-auto-title

I've fixed this in my fork of Prezto by preventing the `_terminal-set-titles-with-path` precmd from being used when `EMACS` is set.

Thanks for Prezto BTW. :smile:

### DIFF
--- a/modules/terminal/init.zsh
+++ b/modules/terminal/init.zsh
@@ -6,7 +6,7 @@
 #
 
 # Return if requirements are not found.
-if [[ "$TERM" == (dumb|linux|*bsd*) ]]; then
+if [[ "$TERM" == (dumb|linux|*bsd*|eterm*) ]]; then
   return 1
 fi
 


sorin-ionescu:
Should not Emacs fail at line 9?

jcf:
@sorin-ionescu unfortunately term, ansi-term, and multi-term identify themselves as `eterm-color`, which won't match line 9.

I could move the conditional logic up to line 9 though. Something like this:

``` sh
if [[ "$TERM" == (dumb|linux|*bsd*) || -z "$EMACS" ]]; then
  return 1
fi
```

Let me know what you think.

sorin-ionescu:
`[[ "$TERM" == (dumb|linux|*bsd*|eterm*) ]]`

jcf:
@sorin-ionescu I like it. Changed as per your recommendation.

jcf:
@sorin-ionescu I've found a similar issue in the prompt module. Is there some way to extract this logic and share it throughout prezto?

sorin-ionescu:
No, but it can be added to all modules, which are not many, that do not work on dumb terminals. It's an easy change; I'll merge it.

sorin-ionescu:
I need to know what works in Emacs. These files check for dumb terminals:

[init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/init.zsh#L95)
[modules/completion/init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/modules/completion/init.zsh#L10)
[modules/editor/init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/modules/editor/init.zsh#L9)
[modules/prompt/init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/modules/prompt/init.zsh#L13)
[modules/spectrum/init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/modules/spectrum/init.zsh#L10)
[modules/terminal/init.zsh](https://github.com/sorin-ionescu/prezto/blob/master/modules/terminal/init.zsh#L9)

jcf:
I can try out each when I'm back at my computer tomorrow. I'll post an
update as aol as I have the necessary info.

Thanks Sorin.

sorin-ionescu:
@jcf, what's the status of this?

jcf:
Hey @sorin-ionescu,

I haven't had time to look at it what with the Easter break and work the last couple of days. I can test each of the modules you listed this weekend.

jcf:
Okay, I've done a quick run through, and tested to the best of my knowledge.

### init.zsh

Both colour and my prompt work as expected. I don't think this check needs to be changed.

### modules/completion/init.zsh

Completion works nicely. You might have to change a few keybindings to let <kbd>&lt;tab&gt;</kbd> get through but that's nothing to do with Prezto.

completion

### modules/editor/init.zsh

I don't know how to test all of this. Basic keystrokes like <kbd>&lt;up&gt;</kbd> work as expected.

### modules/prompt/init.zsh

My prompt looks fine.

### modules/spectrum/init.zsh

Looks reasonable to me.

spectrum

### modules/terminal/init.zsh

Activating this doesn't break anything, but titles aren't updated. You'd have to do some work to get Emacs to support this I think.

Hope that helps. I wonder if there's some way to automate testing this stuff…

sorin-ionescu:
Thank you.

How does the Sorin theme look like in Emacs? In my limited tests, RPROMPT is broken.

jcf:
I make use of RPROMPT too, and it shows up. It doesn't play too well with
evil-mode, but I don't think that's a Prezto problem.

sorin-ionescu:
It does show up, but for me, I see escape characters, which break the line. I'll post a screenshot later.

sorin-ionescu:
RPROMPT is broken for me.

Emacs

jcf:
Perhaps we can merge this small change to address the issue with using the terminal module, and think about a solution to the `RPROMPT` issue separately?

sorin-ionescu:
Merged.

PR #831, opened 2015-03-11
Add sys module

k0lter:
A simple module to make possible to use various system informations (gathered from LSB, uname, etc.) in prompt.


belak:
Hi! Thanks for submitting this! I really like the idea, but as we're not really accepting new modules or prompts in the main repo, if you're still interested in getting this in, please feel free to submit it to the contrib repo (https://github.com/belak/prezto-contrib).

PR #832, opened 2015-03-12
Add aliases to osx module

thetateal:
Include aliases for lock screen, screencapture, and system volume.

`lock` brings up the login screen without a full logout.
`snap` saves a screen shot to current dir following default OS X screen shot naming convention
`mute` and `unmute` are essentially identical to pressing the physical mute key, can be changed via any other volume altering function in OS X. I find it useful for (1) not straying too far from the home row and (2) when function keys and fn are remapped to oblivion.

PS - thanks so much for prezto. OMZ performance was killing me.

### DIFF
--- a/modules/osx/init.zsh
+++ b/modules/osx/init.zsh
@@ -19,3 +19,13 @@ alias cdf='cd "$(pfd)"'
 
 # Pushes directory to the current Finder directory.
 alias pushdf='pushd "$(pfd)"'
+
+# Locks screen.
+alias lock='/System/Library/CoreServices/Menu\ Extras/User.menu/Contents/Resources/CGSession -suspend'
+
+# Saves a screen shot to the current directory with 1s delay for app switching.
+alias snap="screencapture -T 1 "Screen Shot $(date +'%Y-%m-%d') at $(date +"%I.%M.%S %p").png";"
+
+# Mutes/Unmutes the system volume.
+alias mute="osascript -e 'set volume output muted true'"
+alias unmute="osascript -e 'set volume output muted false'"


sorin-ionescu:
Please document your additions in the README.

thetateal:
Updated the README with descriptions for each

modules/osx/init.zsh
```diff
@@ -19,3 +19,13 @@ alias cdf='cd "$(pfd)"'
 
 # Pushes directory to the current Finder directory.
 alias pushdf='pushd "$(pfd)"'
+
+# Locks screen.
+alias lock='/System/Library/CoreServices/Menu\ Extras/User.menu/Contents/Resources/CGSession -suspend'
```
kudos: Seems like overkill, something like `pmset displaysleepnow` is nicer and isn't as slow to show the login prompt since it doesn't initiate fast user switching.

sgtsquiggs: the ```pmset``` approach is a lot more elegant. I like the rest of the additions.

janka102: Unfortunately just doing the `pmset` command by itself is not enough to actually lock the screen, it will only put the display to sleep. Looking at [a question on AskDifferent](http://apple.stackexchange.com/q/80058/55251) it seems the lock command @musophob wrote is the best option to lock the screen regardless of user settings.

andruby:
This seems like a few great additions. Is there something blocking this from being merged in?

sorin-ionescu:
@andruby I haven't reviewed it.

modules/osx/init.zsh
macserv: Looks like the `pmset` approach is the only remaining one on Big Sur/Monterey... `User.menu` is gone, now that the Control Center has supplanted most Menu Extras.

PR #833, opened 2015-03-13
Added rebar completion (from omz)

rjcoelho:
https://github.com/rebar


sorin-ionescu:
Please submit completions to the zsh-completions project.

PR #834, opened 2015-03-13
Module/aws

rjcoelho:
https://github.com/aws/aws-cli


sorin-ionescu:
This module is not matching the Prezto coding style.

The prompt code doesn't play well with Prezto prompts; see `ruby-info` for an example on how to do it.

There is also a lot of unnecessary code to load a completion.

PR #836, opened 2015-03-13
Update README with putty font issue

rjcoelho:
See http://superuser.com/questions/393834/how-to-configure-putty-to-display-these-characters


sorin-ionescu:
Broken pull request of #834.

PR #837, opened 2015-03-15
Add virtualenv info to prezto paradox prompt

milin:
The paradox prompt was missing virtualenv info for python projects, so this pull request adds that.

virtualenv info for paradox theme

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 
 [1]: http://www.zsh.org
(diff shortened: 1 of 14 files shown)


milin:
@sorin-ionescu Any comments would be helpful regarding this. Have personally been using this patch for a few days with virtualenv support and it seems to be pretty stable.

sorin-ionescu:
It looks fine; it has just a few white space issues — missing blank lines. Is `(e)` needed on variables?

Squash the commits.

milin:
@sorin-ionescu looks like e is not needed, so I removed it. Added missing blank lines and squashed the commits into one.

milin:
Anything else needed before merging @sorin-ionescu

PR #846, opened 2015-04-07
fix old/broken submodule

Eriner:
This fixes the failed build by updating the submodule commit.


sorin-ionescu:
Fixed in c98da90. I don't know what's up with that commit, perhaps a forced-push?

PR #854, opened 2015-04-14
Add Arch Linux configuration for node module

ianks:
Currently, only manual and Homebrew locations are checked for NVM. This adds a configuration for Pacman.


sorin-ionescu:
This is AUR, not official.

ianks:
Yes I don't imagine NVM would ever make it into the official repos...

sorin-ionescu:
Just install it in your home directory for now.

PR #856, opened 2015-04-16
Add pattern tweaking to syntax-highlighting

el1t:
Enable setting patterns in *zpreztorc* like styles are set, since patterns are set separately.
Accidentally closed the other pull request—sorry! I had to make a small edit.


sorin-ionescu:
You can update pull requests using forced push.

el1t:
Yeah I did that except I accidentally reused a command that pushed HEAD instead of master; I'll be more careful next time :wink:

PR #858, opened 2015-04-28
omit warning of brew upgrade

tevino:
In newly version of homebrew when you execute `brew upgrade` you will
get the following:

> Warning: brew upgrade with no arguments will change behaviour soon!
> It currently upgrades all formula but this will soon change to require '--all'.
> Please update any workflows, documentation and scripts!

### DIFF
--- a/modules/homebrew/README.md
+++ b/modules/homebrew/README.md
@@ -13,8 +13,7 @@ Aliases
   - `brewi` installs a formula.
   - `brewl` lists installed formulae.
   - `brews` searches for a formula.
-  - `brewU` upgrades Homebrew and outdated brews.
-  - `brewu` upgrades Homebrew.
+  - `brewu [formulae]` upgrades Homebrew and all/specified formulae.
   - `brewx` uninstalls a formula.
 
 ### Homebrew Cask
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -20,8 +20,10 @@ alias brewC='brew cleanup --force'
 alias brewi='brew install'
 alias brewl='brew list'
 alias brews='brew search'
-alias brewu='brew upgrade'
-alias brewU='brew update && brew upgrade'
+brewu() {
+    brew update
+    [[ -z $1 ]] && brew upgrade --all || brew upgrade "$@"
+}
 alias brewx='brew remove'
 
 # Homebrew Cask


janka102:
I did a bit of testing and it seems you can't override the `--all` argument by adding a formula name after it (also the [source](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/cmd/upgrade.rb#L8) seems to back this up too). ie. `brewu git` would still attempt to upgrade all formulae, and I'm not sure that would be the intended outcome.

Perhaps `brewu` should be a function that passes along its arguments to `brew upgrade` unless no arguments were passed, and in that case pass `--all`.

```bash
brewu() {
    [[ $# -eq 0 ]] && brew upgrade --all || brew upgrade "$@"
}
```

sorin-ionescu:
A function is the only way to make it work. Also, I always run `brew update`  before `brew upgrade`. Might as well eliminate `brewU`.

janka102:
@sorin-ionescu I actually personally use `brewU` more often then running those individually, so I do like to have it

sorin-ionescu:
`brewu` will be the new `brewU`.

janka102:
Oh I see what you are saying. Right on, sounds good

tevino:
@sorin-ionescu Eliminate `brewU` is a good idea, but you mean embed `brew update` into new `brewu` or invoke `brew update` manually before `brewu`? The later one seems inconvenient and behaves different from the old `brewU`.

sorin-ionescu:
Don't invoke manually. Merge into `brewu`.

tevino:
I think that's it. @sorin-ionescu

julionc:
bump @sorin-ionescu

sorin-ionescu:
Is there a case where you only want to upgrade just one package instead of all? If you don't want to upgrade a specific package, you just `brew pin` it. The following may be better.

```sh
alias brewu='brew update && brew upgrade --all'`
```

tevino:
Both `brew pin formulae` and `brew upgrade formulae` are essential in different situations, one is usually the opposite to the other, `upgrade` is for fewer packages to upgrade, `pin` is for fewer packages to NOT upgrade.

```
pin formulae
              Pin the specified formulae, preventing them from being upgraded when issuing the brew upgrade --all command.  See  also
              unpin.
```

```
upgrade [--all] [install-options] [formulae]
              Upgrade outdated, unpinned brews.

              Options for the install command are also valid here.

              If  --all is passed, upgrade all formulae. This is currently the same behaviour as without --all but soon --all will be
              required to upgrade all formulae.

              If formulae are given, upgrade only the specified brews (but do so even if they are pinned; see pin, unpin).
```

sorin-ionescu:
I know. My use of brew is as follows: I want to upgrade all packages, always, except for the ones I have pinned. I have never manually updated just one package.

tevino:
My use case is almost the same as you, but I don't see cons of adding support for upgrade a single formulae, what you want is still there, nothing change, `brewu` is equal to `brew update && brew upgrade --all`

tevino:
BTW, the following code is exactly the first version of this PR.
```bash
alias brewu='brew update && brew upgrade --all'`
```

@janka102 proposed the support for upgrade a single formulae, and you seems agreed at first, then I changed the code to what it is now.

sorin-ionescu:
Merged my own.

PR #785, opened 2015-01-26
add support for pyenv-virtualenv

seanzxx:

PR #862, opened 2015-04-29
Use pyenv's virtualenv capabilities.

jszczepanik:
Add pyenv virtualenv or virtualenvwrapper support. I know it's two things for one PR, but they are so closely related that I couldn't resist. Now, I know this is a duplicate of #785 but honestly, I don't think #785 was going to work.

There's a couple of things to mention here:
1. Pyenv's virtualenvs work anyways without this feature, but they aren't activated. Basically this means, that the prompt wouldn't change for them.
2. I've checked prezto's prompt for virtualenvs and it works the same for pyenv's virtualenvs.
3. Pyenv virtualenv messes up with `precmd` hooks, so until they merge https://github.com/yyuu/pyenv-virtualenv/pull/77 or solve it another way, prezto's prompt is going to break after `eval "$(pyenv virtualenv-init -)"`
4. Virtualenvwrapper crashes if pyenv is initialized, unless https://github.com/yyuu/pyenv-virtualenvwrapper is used.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -34,8 +34,28 @@ if (( ! $+commands[python] && ! $+commands[pyenv] )); then
   return 1
 fi
 
+# Load pyenv's virtualenvwrapper into the shell session.
+if [[ $(pyenv commands) == *virtualenvwrapper* ]] 2&> /dev/null; then
+  # Prefer pyenv over virtualenv.
+  export PYENV_VIRTUALENVWRAPPER_PREFER_PYVENV="true"
+
+  # Set the directory where virtual environments are stored.
+  export WORKON_HOME="$HOME/.virtualenvs"
+
+  # Disable the virtualenv prompt.
+  VIRTUAL_ENV_DISABLE_PROMPT=1
+
+  pyenv virtualenvwrapper
+
+# Load pyenv's virtualenvs into the shell session.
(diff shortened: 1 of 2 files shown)


modules/python/init.zsh
```diff
@@ -34,8 +34,28 @@ if (( ! $+commands[python] && ! $+commands[pyenv] )); then
...
+# Load pyenv's virtualenvs into the shell session.
+elif [[ $(pyenv commands) == *virtualenv-init* ]] 2&> /dev/null; then
+  # Disable the virtualenv prompt.
+  VIRTUAL_ENV_DISABLE_PROMPT=1
+
+  eval "$(pyenv virtualenv-init -)"
+
 # Load virtualenvwrapper into the shell session.
-if (( $+commands[virtualenvwrapper.sh] )); then
+elif (( $+commands[virtualenvwrapper.sh] )); then
```
hex4gon: I am not really convinced about these `elif`s. If you got better idea please tell.

hex4gon:
yyuu/pyenv-virtualenv has now fixed it's problems with hooks, so initializing it shouldn't no longer affect prezto.

belak:
Thanks for your contribution. This has been superseded by #1305 so I can consolidate the work from a number of other PRs.

PR #863, opened 2015-04-30
Add support for docker and boot2docker

skarfacegc:
Add some commands to reduce typing when using boot2docker (setting up the correct environmental variables correctly etc).

Add completion for docker commands


modules/docker/init.zsh
```diff
@@ -0,0 +1,35 @@
+#
+# Defines docker and boot2docker commands & aliases
+#
+# Authors:
+#   andrew williams (https://github.com/skarfacegc)
+#
+
+# Only setup the boot2docker commands and aliases if we're on OSX
+if [[ "$OSTYPE" == darwin* ]]; then
```
sorin-ionescu: Other operating systems don't need this?

skarfacegc:
The completions aren't mine, they're out of the docker project (I did check the license before I sent the pull request).  I'll see if I can get the two teams to talk to each other.

modules/docker/init.zsh
skarfacegc: Linux boxes dont since they can run docker natively  (don't need the virtualization layer that boot2docker provides).  Windows likely needs it, but I only have access to windows VMs so I can't test a vm within a vm.

skarfacegc:
Docker (at least when installed via the boot2docker keg) comes with a bash completion script.  I can search for and autoload that if found.  Not sure of the best way to find it if not installed via homebrew.  Thoughts?

sorin-ionescu:
BASH and Zsh completions are not compatible.

pavel-v-chernykh:
@skarfacegc will you work on this PR? If you have no time I can merge it with mine and send. If you don't mind of course

skarfacegc:
The only thing blocking it is having the completions added to the zsh-completions repo.  I've contacted them and have heard nothing back.

pavel-v-chernykh:
As I understood this PR needs to be cleared up from completion code and everything else is ok.

kusold:
https://github.com/zsh-users/zsh-completions/blob/master/src/_docker

Docker completions appears to have been merged in towards the beginning of June with this commit: https://github.com/zsh-users/zsh-completions/commit/73b3d62408ecfcad23576f58e59907549e78b388

Is anything else required?

gerrywastaken:
@kusold Who was that question directed at? You should tag them.

skarfacegc:
Sorry, I've been crushed with work.  I'll get the PR updated to use the docker completions in zsh-completions.

modules/docker/init.zsh
malikoth: There are native Mac and Windows clients in beta.  They still use a VM behind the scenes, but once these clients leave beta, boot2docker will basically be out of a job.

skarfacegc:
switched to docker-machine and removed completions

skarfacegc:
I've got a few more things I want to add to this.  

more docker-machine aliases and automatically attaching to the first running docker-machine instance

jeffwidman:
FYI: https://github.com/felixr/docker-zsh-completion has instructions for integrating with Prezto / completions module

belak:
There was some major work in #1147 to get a bunch of docker aliases in. I'm not sure if there are any aliases here that weren't in the other PR, but feel free to open another with those changes if you want them. Thanks for taking the time to submit this!

Issue #851, opened 2015-04-13
modules/directory vs modules/utility


leoj3n:
I feel there is some organizational inconsistency between the [`directory` module](https://github.com/sorin-ionescu/prezto/blob/c98da90662a11e7ac1e046ccd6a99edd21e68312/modules/directory/init.zsh) and the [`utility` module](https://github.com/sorin-ionescu/prezto/blob/c98da90662a11e7ac1e046ccd6a99edd21e68312/modules/utility/init.zsh) because they both include aliases related to directory switching. My opinion is the directory aliases should all be in the directory module or all in the utility module and not split up between two modules. Thoughts on this?

sorin-ionescu:
You might be right. They remained in utility after the project was forked from Oh-My-Zsh. Utility ought to contain stuff that doesn't fit anywhere else.

ressu:
Both of the directory and utility modules also need a way to disable some of the safeties. Both of the modules contain useful things but for someone who lives in the terminal shell, all the safeties get in the way.

I was just about to send a pull request adding a safety-off option to the utility module which would have disabled the interactivity in cp, mv & friends. But after thinking about it a bit, I realized that utility should be split into parts as well. The dircolors part should be in a separate module. It's not really an utility, but rather a feature that manipulates ls and friends. Same goes for the other coloring options.

As for the directory module, things like unsetting CLOBBER do not fit in to directory behaviour.

sorin-ionescu:
@ressu, submit your changes, and I'll have a look. If you make constant backups, safety is not useful; if you don't, it very much is. I have been burned many times by overwriting or deleting the wrong file; it ceased when I have turned on safety.

ressu:
I'll see if I manage to get back to the changes this weekend. I undid all my changes after I started thinking about the organization a bit.

As for the safety settings, I've found that too much safety makes you careless since you can always rely on the safety net to catch you if you make a mistake. What happens when you don't have the safety net in place?

In any case, my solution was to add a flag that allows the user to disable the interactivity in certain commands, but leave it undocumented. Even if I prefer to not use safeties, I don't want to promote disabling them.

leoj3n:
@belak Should revisit this perhaps?

PR #869, opened 2015-05-04
Add interactivity toggle to utility module

ressu:
Make it possible to remove interactive options from aliases. Disabling these come down to a matter of opinion, so I left them on by default and because it could be argued that it's not always in the best interest of the user to disable these, I left the feature undocumented.

This was briefly discussed in issue #851 as well.

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -28,6 +28,14 @@ To disable `make` highlighting, add the following line to *zpreztorc*:
 
     zstyle ':prezto:module:utility:make' color 'no'
 
+### Interactive aliases
+
+By default `rm`, `cp`, `mv` and `ln` are aliased with the `-i` flag, which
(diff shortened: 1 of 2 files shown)


nasenatmer:
Great pull request! I've change the alias of `rm` to `-I`, but that is individual taste.

However, about the documentation, I disagree. Any feature or config option, especially zstyles should be documented. It is possible to caution users against doing this but we shouldn't prohibit them actively by non-informing them.

ressu:
Hmm.. You have a point with the documentation. I'll add that as well.

nasenatmer:
:+1:

belak:
There was a safe-ops option added a while back in 2a75fba6a80b91a764fbd2e25123f69e22554232 which does something similar to this, so I'm closing this.

Thanks for submitting this and getting the ball rolling on the idea. :)

PR #876, opened 2015-05-09
Update Utility http-serve

plessbd:
start a SimpleHTTPServer with UTF-8 encoding
http://stackoverflow.com/questions/15288891/how-can-i-serve-files-with-utf-8-encoding-using-python-simplehttpserver#24517632

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -158,7 +158,7 @@ fi
 # Miscellaneous
 
 # Serves a directory via HTTP.
-alias http-serve='python -m SimpleHTTPServer'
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
You should probably use a real server.

plessbd:
95% of the time I do, but I had a company send me some stuff that I didnt feel like setting up a webserver for and it needed utf, so thought this might help a couple other people in the rare case that it was needed.

belak:
I'd like to keep this as simple as possible for now. I'm also not sure, but the python3 version may handle unicode better... and it will be used automatically if python3 is installed.

PR #878, opened 2015-05-10
substitute command name when alias is not defined (for ls and grep in the utility module)

zmwangx:
`$aliases[grep]` works with the assumption that an alias is already defined for `grep`. If one comments out the correction-related code, including `alias grep='nocorrect grep'`, then `alias grep="$aliases[grep] --color=auto"` would break. `${aliases[grep]:-grep}` fixes the problem. Same goes for `ls`.

In fact, the `${aliases[command]:-command}` idiom is already seen in the file, just not used throughout.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -71,9 +71,9 @@ if is-callable 'dircolors'; then
       eval "$(dircolors --sh)"
     fi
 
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
Commited.

PR #880, opened 2015-05-19
osx module: add iTerm2 support to the tab function, and create a similar window function

zmwangx:
The `tab` function of the `osx` module currently supports opening a new tab in pwd in Terminal or iTerm. I've added support for iTerm2.

I also adapted the `tab` function a little bit to make a `window` function, which opens a new window rather than a new tab (I personally like windows better than tabs, which I can separately hide and show and drag around to different screens). I didn't include iTerm in the `window` function; it should be trivial to write but since I don't have iTerm installed, I didn't want to risk making a mistake.

The two things are kept in two separate commits so that if you find `window` not worth including, you can easily cherry pick the iTerm2 support in `tab`. Thanks.

### DIFF
--- a/modules/osx/functions/tab
+++ b/modules/osx/functions/tab
@@ -1,5 +1,5 @@
 #
-# Opens a new Terminal.app/iTerm.app tab in the current directory.
+# Opens a new Terminal.app/iTerm.app/iTerm2.app tab in the current directory.
 #
 # Authors:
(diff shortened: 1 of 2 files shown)


zmwangx:
Actually the AppleScript syntax here is for iTerm2 v2.9+. There were breaking changes introduced in v2.9, and I'm not familiar with the syntax of 2.1.1.

PR #890, opened 2015-05-28
modules/browser-home-profile (web-browser profile/cache utility)

tokiclover:
This module, previously known as *firefox-profile* (generalized to many web-browsers), is a utility to maintain home profile/cache directory-ies to tmpfs, or any custom temporary directory by setting up `/tmp` accordingly (e.g. ZRAM backed device on Linux OSs) to improve speed and responsiveness of the said browsers (and maybe anonymize the cache/profile-dir if so desired leaving nothing behind of web-activities but to... ISPs and 3rd parties like Google, Facebook and such.)

Backup tarballs are saved/restored when running `bhp` command in the shell automatically. When loaded, the temporary directory are set (`sudo` is used here to mount directories because the tempoary directories are ramdomly generated.) So an empty and temporary profile/cache-dir are the default. Still, this can be changed to get a default profile setup.

Default compressor is lz4, lzo might be a better compatibility choice, for speed and efficiency; so running `bhp` is seamless and practically instantaneous (even for a hundreds mega bytes, bottleneck is storage... and CPU for this kind of de/compress things.)

**History:**

A previous pull request was made when the module was still firefox *bounded*... I did not have any plan to generalize it because it would have required to install many browsers and investigate the cache/profile directories which I did not want to do such a tedious job.

Just stumbled into a *funy* bash script a few weeks ago... which has a nice list of profile/cache-dir per browser... so I've took the *gold* that was of interest and left the *dirt* if I might say... because I had, since the *firefox-profile* days, {ba,z}sh variants which can be used as standalone script or sourced. (I happen, as many, to have bash as allback shell.) Thanks to this tedious investigation, generalizing was almost a piece of cake... actually easily said than done.

Cheers.

**PS:**
Squashing the commits wont be that easy because of the old history... still, this can be done in a new branch.


nasenatmer:
Isn't this rather the task for a dedicated program such as `profile-sync-daemon`¹ or `anything-sync-daemon`²? They were written in order to reduce wear on SSD disks by putting profile or other use-specified directories in `/tmp` and sync them back to disk from time to time.

If I understand correctly, your module adds the option to have empty profiles, which is nice, but deleting cookies every time and not creating a browsing history (or even using incognito modes) is almost getting the same effect, isn't it? But maybe I'm also missing something here…

¹: https://wiki.archlinux.org/index.php/Profile-sync-daemon
²: https://wiki.archlinux.org/index.php/Anything-sync-daemon

tokiclover:
Well, you can use *a dedicated program* if you please for such an easy task. My personal peference is to have something set up everything for you, and then have an efficient function to *sync* (de/compress) a profile/cache-directory when need be.

I realize that I have always many virtual terminal intances (at least [URxvt][1], [Terminology][2] (lightweight and eye candy), and sometimes bar [XTerm][3] for...), first; spawned in many virtual desktops, second. And I like to *sync* when I feel like it and desire to save a *session* for futur usages because many tabs are waiting to be completely read/dealt-with. So this script can keep a *session* forever with a tarball and a tarball backup without any risk to lose data/session consistently over time.

This is not... a standalone script nor a *fake* daemon that *sync* regularly almost by design to storage and hard modify the profile/cache-directory in such a way, that, if something bad happen by accident (broken script, corrupetd script...) no usable profile/cache-directory will be available.

Contrary to this, bhp does not touch the underlying profile/cache-directory. It does only use them (bind-)mounting upon them leaving everything in place, so that, a fallback or fake or empty profile/cache-directory is always available without needing to investigate/debug/revert previous change. Unmounting is the most to be done to retrieve the fallback/empty/default profile without further actions.

I was not aware of [profile-sync-daemon][4] untill a few weeks ago... I was busy with other stuff if I might say. This is where I am ironacally talking about mining and gold finding and took the opportunity to include other browsers... I did not want to waste my time to this front. Anybody could have done it--meaning that users of other browsers could have (and then could push their changes) if so desired.

Daemon in psd is only a *name* without any real substance, first. Second, the script is more than lengthy and tied entierly to bash... there is not actually much of a hard dependency here, it just seems the author decided to use bashisms--that's all.

First, after discovering this I decided to launch a POSIX shell compatible script named [browser-home-profile][5] because it seems many people are interested in it; and I did not found psd being optimal (again lengthy script) nor took a compatibilty issues seriously tying the script to bash without any purpose.

WARNING: I am talking very far way because I only took a look and then... got a slight grim on my face and decided to take off without further investigations/reviews.

Notice, *bhp* comes from a previous script called *fhp* (firefox-home-profile, [bhp.bash][6], [bhp.bash][7]) which has a very old history. Why bash and zsh variant? Because it can be surced to `~/.bashrc` or `~/.zshrc` and use only CPU cycles *only* to de/compress profile/cache-directory when need be without any extra waste. Second, a `bhp` call (previously `fhp`) can be placed to `~/.bash_logout` or `~/.zlogout` or even a `zsh-exit-hook` to save everything in the right spot. (This is what I am calling efficiency in the previous lines.)

I hope that I layed down enough infos, and not *too much*, to talk about a sourceable script rather than wasting times on discussing psd--for those who are interested in a bash script follow the links provided above, or [bhp.bash][7].

[1]: http://software.schmorp.de/pkg/rxvt-unicode.html
[2]: http://www.enlightenment.org/p.php?p=about/terminology
[3]: http://invisible-island.net/xterm/
[4]: https://github.com/graysky2/profile-sync-daemon/
[5]: https://github.com/tokiclover/browser-home-profile/
[6]: https://raw.githubusercontent.com/tokiclover/dotfiles/master/bin/bhp.bash
[7]: https://raw.githubusercontent.com/tokiclover/dotfiles/master/bin/bhp.zsh

zmwangx:
Having a look at the [current list of modules](https://github.com/sorin-ionescu/prezto/tree/master/modules), I imagine that this module would appeal to a far more limited audience (like <1% or even <‰ compared to the audience of, say, `python`?). Also, this is at best an external module.

> My personal peference is to have something set up everything for you.

Well, of course you can keep it in *your* fork, but I certainly don't want to see this in upstream. I'm already using or can see a personal use case for almost every module currently in Prezto's upstream, but not for this one. No.

> First, after discovering this I decided to launch a POSIX shell compatible script named [browser-home-profile](https://github.com/tokiclover/browser-home-profile/) because it seems many people are interested in it.

No offense, but I don't think the stats of your `browser-home-profile` repo checks out with the claim that people are interested in it. The other links provided also don't seem to speak to the popularity.

zmwangx:
By the way, several problems mentioned in https://github.com/sorin-ionescu/prezto/pull/683#issuecomment-58683290 are still there, as far as I can tell.

sorin-ionescu:
Thanks, @zmwangx; you said everything I was going to say.

PR #898, opened 2015-06-05
python module: pyenv-virtualenvwrapper support (or rather, fix)

zmwangx:
Currently the `python` module is fatal for folks using [`pyenv-virtualenvwrapper`](https://github.com/yyuu/pyenv-virtualenvwrapper), since `pyenv`'s `virtualenvwrapper.sh` is designed for internal use by `pyenv virtualenvwrapper` rather than being sourced. Basically, `pyenv`'s `virtualenvwrapper.sh` (in its `shims` directory) calls `exec` on the actual `virtualenvwrapper.sh`, which returns after making a few function defs and calling `virtualenvwrapper_initialize`, so affected users can't even start a shell session — it immediately returns.

In this commit I try to run `pyenv virtualenvwrapper` first, and check for and try to load `virtualenvwrapper.sh` only if the former fails.

Also, a user-defined `WORKON_HOME` shouldn't be overwritten by the `python` module, so I'm checking emptiness of `WORKON_HOME` before setting it to `$HOME/.virtualenvs`.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -37,12 +37,12 @@ fi
 # Load virtualenvwrapper into the shell session.
 if (( $+commands[virtualenvwrapper.sh] )); then
   # Set the directory where virtual environments are stored.
-  export WORKON_HOME="$HOME/.virtualenvs"
+  [[ -n $WORKON_HOME ]] || export WORKON_HOME="$HOME/.virtualenvs"
 
   # Disable the virtualenv prompt.
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
-  source "$commands[virtualenvwrapper.sh]"
+  pyenv virtualenvwrapper 2>/dev/null || source "$commands[virtualenvwrapper.sh]"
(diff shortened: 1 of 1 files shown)


amir20:
I also reported this on #918. Sorry I only search in the issues. Should I close #918?

archen:
Thanks for the fix! This has been driving me mad most of the night. Any chance that this is going to be merged into master any time soon?

zmwangx:
If this PR hasn't been merged for the last eight months, I bet it's not gonna be merged anytime soon. Just fork the repo.

archen:
Already forked last night, let's just call it wishful thinking on my part that +1'ing may help others.

belak:
There's been a fix for the overwriting of WORKON_HOME in another PR, so if someone wants to clean this up to only make the pyenv change, I'd be happy to merge it.

belak:
There are a number of other options for pyenv fixes, so I'm going to close this one. Thanks!

PR #904, opened 2015-06-15
Remove useless use of grep

zmwangx:
Shrink `grep 'Location:' | sed 's/Location: //'` to `sed -n 's/^Location: //p'` in `git-hub-shorten-url` (added `^` since `Location:` should appear at the beginning of a line in `curl`'s output).


belak:
Thanks for your contribution!

zmwangx:
Huh, back in business? 😮

belak:
Yep! :)

PR #909, opened 2015-06-26
Fix: Don't force ~/.virtualenvs as WORKON_HOME

jmagnusson:


robsonpeixoto:
:+1:

belak:
I plan on fixing this by merging #598 later today.

PR #910, opened 2015-06-26
Command to activate virtualenvs when changing directory

jmagnusson:
Enables automatic activation of a virtualenv when jumping into a directory. This is done by looking in the current directory for a file or directory named `.venv`. If it's not found in the current directory its parents will also be examined. `.venv` can be either:
1. A file containing the name of a virtualenv found in $WORKON_HOME
2. A directory containing bin/activate (meaning that the directory is assumed to be a virtualenv.)

If $WORKON_HOME is set it is assumed that virtualenvwrapper is installed and the `workon` command will be issued.

### DIFF
--- a/modules/python/functions/workon-cwd
+++ b/modules/python/functions/workon-cwd
@@ -1,0 +1,60 @@
+#
+# Enables automatic activation of a virtualenv when jumping into a directory
+# This is done by looking in the current directory for a file or directory
+# named `.venv`. If it's not found in the current directory its parents will
+# also be examined. `.venv` can be either:
+#
+#     1. A file containing the name of a virtualenv found in $WORKON_HOME
+#     2. A directory containing bin/activate (meaning that the directory is
+#        assumed to be a virtualenv.)
+#
+# If $WORKON_HOME is set it is assumed that virtualenvwrapper is installed and
+# the `workon` command will be issued.
+#
+
+function workon_cwd {
+  if [ ! $WORKON_CWD ]; then
+    WORKON_CWD=1
+    # Look for a proper PROJECT_ROOT path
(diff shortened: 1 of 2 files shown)


belak:
I was originally looking at something like this in https://github.com/zsh-users/prezto/pull/57 so I'll try to take a look at this later.

modules/python/functions/workon-cwd
```diff
@@ -0,0 +1,60 @@
...
+    unset PROJECT_ROOT
+    unset WORKON_CWD
+  fi
+}
+
+if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
+  # Append workon_cwd to the chpwd_functions array, so it will be called on cd
+  # http://zsh.sourceforge.net/Doc/Release/Functions.html
+  if ! (( $chpwd_functions[(I)workon_cwd] )); then
+    chpwd_functions+=(workon_cwd)
```
belak: We generally use the add-zsh-hook function over chpwd_functions.

modules/python/functions/workon-cwd
```diff
@@ -0,0 +1,60 @@
...
+      # We've just left the repo, deactivate the environment
+      # Note: this only happens if the virtualenv was activated automatically
+      deactivate && unset CD_VIRTUAL_ENV
+    fi
+    unset PROJECT_ROOT
+    unset WORKON_CWD
+  fi
+}
+
+if [[ ! $DISABLE_VENV_CD -eq 1 ]]; then
```
belak: This would be better as a zstyle option.

belak:
I'm going to close this in favor of the new PR because there's more activity there. Thanks for getting the ball started with this!

PR #915, opened 2015-07-06
Add prompt geertjohan

GeertJohan:
I would like to add my custom prompt as option in prezto.


sorin-ionescu:
If you do not intend to modify it further, it can be merged into Prezto provided it matches the coding style. If you intend to modify it frequently, keep it as an external module and from time to time, it will be updated in Prezto.

Themes require several votes from the community to be included.

GeertJohan:
Thanks @sorin-ionescu 
I do intend to modify it now and then, the repo is also open for issues/PR's.

This theme was originally created about 6 months ago, I think. Just polished the source a bit and uploaded this version now. I intend to make the theme a bit more configurable for anyone without zsh/shell knowledge. One thing I have planned to add is to modify the number of characters that are shown for each folder in the path. So when set to 3, the example would show `~/src/git/Gee/prezto_prompt_geertjohan`.

I hope you like the theme. I've chosen for three lines to be able to display more information while in narrow terminals (which I tend to have a lot). I like to use the first two lines to separate the information about the previous command from the current environment information.

sorin-ionescu:
If users want your theme in Prezto, they'll vote for it in these comments to this pull request.

I require votes to prevent the Wild, Wild, West nature of Oh-My-Zsh from infecting Prezto. I want code in Prezto, in this case a prompt theme, to be used by many, not just by the original author, and prompt themes are often personal.

I'll provide feedback on your theme after I play with it.

GeertJohan:
Thanks!

I really agree with you on preventing a huge amount of theme's flooding the project. Of course I believe my theme is distinct enough to be added, otherwise I wouldn't have opened a PR. If it was a lot like an existing theme then anyone could've used the existing one and modify to personal preference. I think my theme is a different enough and provides new features.

nasenatmer:
:+1:

stefanosc:
+1

PR #917, opened 2015-07-07
modified the check to support non-system provided openssh agents on OSX

### LINKED ISSUE #916: 'ssh' module does not work with brewed ssh-agent on OS X
I have created a patch which changes the check at the beginning of the 'ssh' module for prezto from checking to see if you're on OS X and returning to checking to see if you have an alternative ssh-agent installed if you are on OS X and returning if not.

The reasoning here is that historically it hasn't been important for prezto to load identities on OS X because the system provided ssh-agent has OS X keychain integration, as did the alternatives.  Starting with the most recent Homebrew update to the openssh package this is no longer true per https://github.com/Homebrew/homebrew-dupes/pull/482 which removed the ability to build openssh with the '--with-keychain-support' flag.

For those of us running brewed openssh (or other alternatives such as fink, macports, or manually installed) it's important that the expected behavior of the ssh module matches how it would behave on other POSIX systems, thus loading identities at start.

Tristor:
closes #916

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -136,4 +136,13 @@ function prompt_sorin_setup {
   SPROMPT='zsh: correct %F{1}%R%f to %F{2}%r%f [nyae]? '
 }
 
+function prompt_sorin_preview {
+  local +h PROMPT=''
+  local +h RPROMPT=''
+  local +h SPROMPT=''
+
+  editor-info 2>/dev/null
+  prompt_preview_theme 'sorin'
+}
+
 prompt_sorin_setup "$@"


Tristor:
Hold off on merging this.  It looks like the logic for checking for the existing ssh-agent and starting a new one doesn't work appropriately on OS X.  I may need to do another commit to ensure this patch works as expected.  ssh-add -l hangs because the ssh-agent isn't running when that part gets executed when loading the 'ssh' module.  Not sure why quite yet

Tristor:
Okay, this now tests successfully and behaves as expected, namely that you get prompted to enter the passphrase for your SSH key the first time you load a terminal and future terminals use the same SSH agent.

Tristor:
One additional thought.  I had a signal trap in my .zshrc when I was utilizing the previously patched version of ssh-agent from Homebrew rather than the prezto 'ssh' module to cleanup.  It doesn't appear that the ssh-agent is terminated when your last terminal session is closed.  I'm not sure if that's desirable behavior or not, so I haven't tried adding it in with this patch.  It is less convenient, but maybe more secure.

modules/ssh/init.zsh
```diff
@@ -6,10 +6,15 @@
 #
 
 # Return if requirements are not found.
-if [[ "$OSTYPE" == darwin* ]] || (( ! $+commands[ssh-agent] )); then
-  return 1
+if [[ "$OSTYPE" == darwin* ]]; then
+    if [[ "$(which ssh-agent)" == "/usr/bin/ssh-agent" ]]; then
```
sorin-ionescu: You do not need `which`. Zsh has the `$commands` array. Furthermore, you're making assumptions about where `ssh-agent` is installed, in this case `/usr/bin`. The logic here needs to change.

modules/ssh/init.zsh
```diff
@@ -28,6 +33,14 @@ if [[ ! -S "$SSH_AUTH_SOCK" ]]; then
   if ! ps -U "$LOGNAME" -o pid,ucomm | grep -q -- "${SSH_AGENT_PID:--1} ssh-agent"; then
     eval "$(ssh-agent | sed '/^echo /d' | tee "$_ssh_agent_env")"
   fi
+elif [[ "$SSH_AUTH_SOCK" == "$(launchctl getenv SSH_AUTH_SOCK)" ]]; then
```
sorin-ionescu: Why would `$SSH_AUTH_SOCK` in Zsh be different from `launchctl`?

modules/ssh/init.zsh
Tristor: The commands array lets me check the existence of a program, but it doesn't allow me to know where it's executable is located.  On OS X the system-provided ssh-agent is in /usr/bin/ssh-agent.  If you use Fink or MacPorts it's either in /opt or in /usr/local, if you use Homebrew it's in /usr/local.

Why it's important to differentiate is that the system-provided ssh-agent has a set of Apple custom patches to have non-standard behavior.  This behavior is a combination of support for integration with the OS X keychain (which used to be patchable to upstream via Homebrew but is no longer) and some additional flags it supports which are expected by how launchd interacts with it.

When using the system provides ssh-agent, rather than it running in the background, when a call is made to its socket, launchd starts it temporarily to respond to the request using socket redirection and then once the request is completed it shuts the agent down again.  The agent with Apple's patches is just a wrapper for the OS X keychain.

However, if you want something newer than the ancient version provided by OS X, you need to install it from Fink, MacPorts, Homebrew, or manually.  Once you've done so, it's behavior matches the expected behavior for all other POSIX systems.  In the original case, it was checking to see if the system was OS X or lacked ssh-agent, if either was true it merely returned 1 and then exited.  In my patch, it now checks to see if your OS is OS X, and if so, checks to see if you're using the system provided ssh-agent, if so it returns 1, otherwise it continues.  If you don't have ssh-agent installed, it returns 1, otherwise it continues.  This logic flow allows it to work properly on both modified and unmodified OS X systems.

modules/ssh/init.zsh
Tristor: SSH_AUTH_SOCK is an environment variable that gets set on starting your login session by launchd.  The system default behavior is to have this set to a launchd socket that resides in /tmp and captures the incoming connections to the system default ssh-agent and then starts it on demand and uses socket redirection to send these connections to it.

In the 'ssh' module prior to my patch, it was merely checking to see if $SSH_AUTH_SOCK was existed and was a socket.  Since this will always be true, it never causes the ssh-agent to get started on OS X.  After my modification, it checks to see if it's matching the value launchctl will set on login, and if so, continues setting up the connection.  During the ssh-agent startup process SSH_AUTH_SOCK gets changed.

tduzan:
Just for giggles, I updated one of my Macs to Yosemite.  I've now tested this patch under 10.10.4 and it works fine there as well.  Since the changes coming in 10.11 will make the previous patchset provided by Homebrew against OpenSSH unusable, it's pretty important that we accommodate Mac users which install OpenSSH via some other method rather than using the system provided version.  

Any further concerns with the patch or any suggested changes to make this PR acceptable?

tduzan:
Any updates on acceptability of this patch?  Thanks.

tduzan:
Any updates on this?  Thanks.

PR #924, opened 2015-07-10
sorin prompt improvements - rbenv/pyenv support

sgtsquiggs:
Adds support to `sorin` for ruby-info and python-info. Pull down information asynchronously and presents as part of `RPROMPT`.

squiggs screenshot


sorin-ionescu:
There is an awful lot of duplication in there.

sgtsquiggs:
Merged the unnecessarily broken out functions back into the trap function

sorin-ionescu:
It may be worth looking into [zsh-async](https://github.com/mafredri/zsh-async).

belak:
This seems quite useful. Is there any reason this can't be merged @sorin-ionescu ?

belak:
There have been a number of conflicting changes merged. Any chance this could be updated?

sorin-ionescu:
It's useful. I rejected it initially due to duplicate code. It's too dirty for my taste. I also wanted to migrate to zsh-async before adding more async functionality.

sorin-ionescu:
@belak Ask submitters to squash their commits into one before merging unless they are vastly complicated modules. It keeps the log clean.

belak:
If there are multiple commits I usually use the squash and merge feature GitHub has so I don't need to rely on other people doing that. :)

sorin-ionescu:
That tells you how much I use the Github interface.

sgtsquiggs:
Working on a rewrite of this PR using zsh-async!

sgtsquiggs:
Ready for review. Squashed all commits together.

belak:
Might take me a bit to look at this because it's a larger rewrite, but thanks for your work so far!

sorin-ionescu:
@sgtsquiggs Thanks for the rewrite.

@belak It's a big rewrite. I would have liked the async revisions independently from additional Ruby/Python info support so I can revert it if needed. Since this is the default theme, it needs to just work, as in, bug free.

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -32,36 +38,38 @@
...
-      RPROMPT+='${git_info:+${(e)git_info[status]}}'
-      unalias typeset
-    fi
-
-    # Reset PID.
-    _prompt_sorin_precmd_async_pid=0
-
-    # Redisplay prompt.
-    zle && zle reset-prompt
+function _prompt_sorin_async_job {
```
sorin-ionescu: Functions don't need to prefixed with an underscore.

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -32,36 +38,38 @@
...
+  prompt_sorin_style
```
sorin-ionescu: Why are all styles redefined on each job?

sgtsquiggs:
I can split the rewrite of the original from the Ruby/Python additions and force push. 

I'll fix the underscores too

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -71,30 +79,46 @@ function prompt_sorin_precmd {
...
   prompt_opts=(cr percent subst)
-  _prompt_sorin_precmd_async_pid=0
-  _prompt_sorin_precmd_async_data="${TMPPREFIX}-prompt_sorin_data"
 
   # Load required functions.
   autoload -Uz add-zsh-hook
+  autoload -Uz async && async
+
+  # Start async worker
+  # async_start_worker 'prompt_sorin' -n -u
```
sorin-ionescu: Why is this commented?

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -71,30 +79,46 @@ function prompt_sorin_precmd {
...
+
+  # Register callback
+  # async_register_callback 'prompt_sorin' _prompt_sorin_callback
```
sorin-ionescu: Why is this commented?

modules/prompt/functions/prompt_sorin_setup
sgtsquiggs: With zsh-async, the worker is spawned off ideally here, when the prompt initializes. This works fine for grabbing up-to-date git info.

This doesn't work with ruby-info python-info, because the environment changes made by rbenv/rvm/virtualenv are made only on the active terminal, not the worker. I guess those environment changes could be sent to the worker job as parameters but that seemed to me to couple too tightly with python-info and ruby-info, and would possibly break as python-info and ruby-info change in the future.

I commented it out when changing the worker to stop/start on every command and forgot to remove it.

modules/prompt/functions/prompt_sorin_setup
sgtsquiggs: Same as above

modules/prompt/functions/prompt_sorin_setup
sgtsquiggs: Not even necessary, removing this and reverting the style function back into `prompt_sorin_setup`

belak:
Does the current code mean that if multiple rprompt modules come in at different times, they could be in different orders each time? Could we just set variables depending on the job in the async callback and call `zle reset-prompt` or am I misunderstanding how that works?

sgtsquiggs:
So the way it works:

1.  Prompt script runs
2. Async job kicked off `prompt_sorin_async_job`
3. `prompt_sorin_async_job` gathers git, rbenv, pyenv all at once.
4. `prompt_sorin_async_job` creates what will be the RPROMPT display, feeds it into the callback `prompt_sorin_async_job`
5. callback sets RPROMPT and calls zle reset-prompt

Calling git/rbenv/pyenv individually resulted in the RPROMPT building itself live and was annoying in use. A lot of flashing as the individual jobs finished.

This could be improved to have each job run asynchronously as a group and then set the RPROMPT when all jobs have finished.

Thoughts?

jeeftor:
Whatever happened with this?  I'd love to see the python virtual env info in there

belak:
That's a good question... it looks like I forgot about this while I was doing my rewrite of the git handling, so there are conflicts now. However, there's now an async task in this prompt so I have a better idea of how it works and would be more likely to merge it...

I'd love to see an updated version of this branch, maybe with some sort of dirty indicator to pass back so it knows if it needs to modify variables (but will still have the caching I added in the git PR)

jeeftor:
I just made a new PR for the python only - the number of lines changed is very minimal #1421

PR #928, opened 2015-07-17
Add statusline theme

el1t:
If I may submit (yet another) custom prompt for consideration:

Screenshot

I made [statusline](https://github.com/el1t/statusline) originally as a clone of agnoster from oh-my-zsh, so it looks somewhat similar to the powerline prompt. However, I also added a host of features, such as task runtime, a single-line prompt mode, and dynamic directory abbreviation. Of course, I included the async loading of git info from sorin :wink:. In addition, I made it extremely easy to edit the segments displayed; the following two arrays can be changed on the fly:
```zsh
_prompt_statusline_left_segments=(status user git-branch directory)
_prompt_statusline_right_segments=(git-status clock history machine)
```
This theme uses a custom patched font, but I can add support for typical powerline-patched fonts if desired. Also, I will edit the coding style to match if this ends up getting voted on :sweat_smile:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "modules/prompt/functions/pure"]
 	path = modules/prompt/external/pure
 	url = https://github.com/sindresorhus/pure.git
+[submodule "modules/prompt/external/statusline"]
+	path = modules/prompt/external/statusline
+	url = https://github.com/el1t/statusline.git
--- a/modules/prompt/functions/prompt_statusline_setup
+++ b/modules/prompt/functions/prompt_statusline_setup
@@ -1,0 +1,1 @@
+../external/statusline/prezto/prompt_statusline_setup


rohanp:
+1

mjkaufer:
:+1:

jeffcox:
This is the best Powerline style theme I've seen for Prezto yet.  I'd like to see some more work done on single line mode (though I may have been seeing weirdness from iTerm2, not this theme), but it's definitely ready for inclusion in the main repo.

el1t:
Hey @jeffcox, I tried running this on iTerm2 and I believe I fixed the error. Problem is, iTerm2 can't load the async git data, but that seems to be an issue with the app. For now, statusline mimics sorin's prompt; git repos will be ignored by iTerm until a workaround is found—although if iTerm were to fix the bug, it should work right away.

sorin-ionescu:
Why would the terminal matter in loading the Git data?

el1t:
@sorin-ionescu Not quite sure, but without testing whether the async data variable exists before sourcing it, iTerm will error due to sourcing an empty variable. Currently, iTerm won't load git data in your sorin theme either. Same zsh binary, same Macbook, so I believe iTerm2 is the issue.
To reproduce, try your sorin theme in iTerm without [this check](https://github.com/sorin-ionescu/prezto/blob/f2a826e963f06a204dc0e09c05fc3e5419799f52/modules/prompt/functions/prompt_sorin_setup#L49).

PR #930, opened 2015-07-21
Load add-zsh-hook in GPG module

andschwa:
Fixes this error:

    ~/.zprezto/modules/gpg/init.zsh:43: command not found: add-zsh-hook

I use Prezto in conjunction with Liquidprompt, and so do not load the prompt module, which otherwise would have loaded `add-zsh-hook`. Regardless, it's required by the GPG module and so should be (re)loaded since it's needed.

### DIFF
--- a/modules/gpg/init.zsh
+++ b/modules/gpg/init.zsh
@@ -30,6 +30,9 @@ export GPG_TTY="$(tty)"
 
 # Integrate with the SSH module.
 if grep 'enable-ssh-support' "$_gpg_agent_conf" &> /dev/null; then
+  # Load required functions.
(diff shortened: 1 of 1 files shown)


modules/gpg/init.zsh
```diff
@@ -10,6 +10,9 @@ if (( ! $+commands[gpg-agent] )); then
   return 1
 fi
 
+# Load required functions.
+autoload -Uz add-zsh-hook
```
sorin-ionescu: Load it after line 35 inside the `if` statement.

andschwa:
Done.

PR #932, opened 2015-07-24
Update completion submodule to latest revision

modille:


sorin-ionescu:
All Git submodules are updated periodically.

sorin-ionescu:
Thanks but I update submodules periodically, and by the time I do, such pull requests are already out of date.

PR #936, opened 2015-07-30
Add symfony2 module ported from symfony2 plugin of oh-my-zsh

longkey1:
Symfony2 is one of the most popular framework in PHP.
I would like to add a module to help the Symfomy2-command.


yoyosan:
+1

posva:
I disagree with this. This a specific language framework. I think you should keep this module on a separated repo so anyone can add it to their fork

longkey1:
@posva Why Rails is no problem?

posva:
@longkey1 It actually is. I guess it was there at the creation of prezto, both roby and sorin were using it, it was on omz and sorin kept it.
However I do agree with you: it definitely is the same thing as rails.

IMHO rails should be removed and we should keep track of existing modules in the wiki or in the readme. Adding a function to add a module to our fork

modules/symfony2/README.md
```diff
@@ -0,0 +1,32 @@
...
+  - [Sorin Ionescu](https://github.com/sorin-ionescu)
+
+[1]: https://symfony.com
+
+sf='`_symfony_console`'
+sfcl='sf cache:clear'
+sfsr='sf server:run -vvv'
+sfcw='sf cache:warmup'
+sfroute='sf router:debug'
+sfcontainer='sf container:debug'
```
sorin-ionescu: Using full words is not much of a 3-4 letter alias. Try to keep the same sequence of letters and the same length.

elboletaire:
+1

modules/symfony2/README.md
longkey1: @sorin-ionescu
Changed, Please review.

sorin-ionescu: That's better.

longkey1:
Fix comment & readme.

laytoneverson:
Is there a framework or utility for installing modules in separate repositories? If it's not easy to install modules, Prezto will suffer users, as I am about to uninstall and switch back to Oh-My so I don't have to spend time figuring out how to implement this specific module / feature.

sorin-ionescu:
@laytoneverson You don't have to wait for me to merge something.

```
curl -sL https://patch-diff.githubusercontent.com/raw/sorin-ionescu/prezto/pull/936.patch | git am -
```

laytoneverson:
Thank you @sorin-ionescu, I always wondered if  you could do that.

skobkin:
+1

Composer module would be nice too.

belak:
Hi! Thanks for submitting this! I like the idea but we're also not really accepting new modules or prompts in the main repo. If you're still interested in getting this in, please feel free to submit it to the contrib repo (https://github.com/belak/prezto-contrib) and I'd be happy to merge it!

PR #937, opened 2015-07-30
modified the check to support non-system provided openssh agents on OSX

Tristor:
Sorry about the administrivia, I accidentally merged some local changes somehow when updating from upstream to keep this commit set current in my branch within my fork and had to generate a new PR because it required a force-push to fix, so that my commit history below would be clean.

This patch closes issue #916 and is a replacement for pull request #917, which I cannot get to reopen successfully in Github after it auto-closed on my force-push.

I addressed your concerns in the comments there about why I'm not using the commands array in ZSH and why I am explicitly testing further down to see if $SSH_AUTH_SOCK was set by launchctl.  This patch has now been live off my fork on ~20 people's systems, including my own, run on OS X Mavericks, Yosemite, FreeBSD, and Ubuntu 15.04 without issues.

Please let me know if you have any additional concerns.

### DIFF
--- a/modules/history-substring-search/README.md
+++ b/modules/history-substring-search/README.md
@@ -63,18 +63,21 @@ _`${ZDOTDIR:-$HOME}/.zpreztorc`_:
 
 ```sh
 zstyle ':prezto:module:history-substring-search' globbing-flags ''
+```
 
-To set the search matched against the start of each history entry, add the following line to
+### Case sensitive search
+
+To set the search case-sensitivity, add the following line to
 _`${ZDOTDIR:-$HOME}/.zpreztorc`_:
 
 ```sh
-zstyle ':prezto:module:history-substring-search' prefixed 'yes'
+zstyle ':prezto:module:history-substring-search' case-sensitive 'yes'
 ```
 
 ### Fuzzy search
 
-To enable fuzzy search add the following line to
-`${ZDOTDIR:-$HOME}/.zpreztorc`_:
+To enable search for fuzzy matches, add the following line to
+_`${ZDOTDIR:-$HOME}/.zpreztorc`_:
 
 ```sh
 zstyle ':prezto:module:history-substring-search' fuzzy 'yes'
@@ -82,11 +85,20 @@ zstyle ':prezto:module:history-substring-search' fuzzy 'yes'
 
 ### Unique results
 
-To retrieve only unique results (remove duplicates) add the following line to
-`${ZDOTDIR:-$HOME}/.zpreztorc`_:
+To enable unique results, add the following line to
+_`${ZDOTDIR:-$HOME}/.zpreztorc`_:
 
 ```sh
-zstyle ':prezto:module:history-substring-search' ensure-unique 'yes'
+zstyle ':prezto:module:history-substring-search' unique 'yes'
+```
+
+### Prefixed search
+
+To enable prefixed search matches, add the following line to
+_`${ZDOTDIR:-$HOME}/.zpreztorc`_:
+
+```sh
+zstyle ':prezto:module:history-substring-search' prefixed 'yes'
 ```
 
 ## Authors
--- a/modules/history-substring-search/init.zsh
+++ b/modules/history-substring-search/init.zsh
@@ -34,10 +34,6 @@ if zstyle -t ':prezto:module:history-substring-search' case-sensitive; then
   HISTORY_SUBSTRING_SEARCH_GLOBBING_FLAGS="${HISTORY_SUBSTRING_SEARCH_GLOBBING_FLAGS//i}"
 fi
 
-if zstyle -t ':prezto:module:history-substring-search' prefixed; then
-  HISTORY_SUBSTRING_SEARCH_PREFIXED='true'
-fi
-
 if ! zstyle -t ':prezto:module:history-substring-search' color; then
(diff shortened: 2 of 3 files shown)


sorin-ionescu:
I see no comments about `$commands` instead of `which`.

cedricpim:
Any news about this?

NeckBeardPrince:
Updates?

Tristor:
I apologize for not updating this previously.  The comments I'm referring to are in response to your questions on the original PR.

From PR #917:

sorin-ionescu added a note on Jul 8, 2015
You do not need which. Zsh has the $commands array. Furthermore, you're making assumptions about where ssh-agent is installed, in this case /usr/bin. The logic here needs to change.

```
The commands array lets me check the existence of a program, but it doesn't allow me to know where it's executable is located. On OS X the system-provided ssh-agent is in /usr/bin/ssh-agent. If you use Fink or MacPorts it's either in /opt or in /usr/local, if you use Homebrew it's in /usr/local.

Why it's important to differentiate is that the system-provided ssh-agent has a set of Apple custom patches to have non-standard behavior. This behavior is a combination of support for integration with the OS X keychain (which used to be patchable to upstream via Homebrew but is no longer) and some additional flags it supports which are expected by how launchd interacts with it.

When using the system provides ssh-agent, rather than it running in the background, when a call is made to its socket, launchd starts it temporarily to respond to the request using socket redirection and then once the request is completed it shuts the agent down again. The agent with Apple's patches is just a wrapper for the OS X keychain.

However, if you want something newer than the ancient version provided by OS X, you need to install it from Fink, MacPorts, Homebrew, or manually. Once you've done so, it's behavior matches the expected behavior for all other POSIX systems. In the original case, it was checking to see if the system was OS X or lacked ssh-agent, if either was true it merely returned 1 and then exited. In my patch, it now checks to see if your OS is OS X, and if so, checks to see if you're using the system provided ssh-agent, if so it returns 1, otherwise it continues. If you don't have ssh-agent installed, it returns 1, otherwise it continues. This logic flow allows it to work properly on both modified and unmodified OS X systems.
```

sorin-ionescu added a note on Jul 8, 2015
Why would $SSH_AUTH_SOCK in Zsh be different from launchctl?

```
SSH_AUTH_SOCK is an environment variable that gets set on starting your login session by launchd. The system default behavior is to have this set to a launchd socket that resides in /tmp and captures the incoming connections to the system default ssh-agent and then starts it on demand and uses socket redirection to send these connections to it.

In the 'ssh' module prior to my patch, it was merely checking to see if $SSH_AUTH_SOCK was existed and was a socket. Since this will always be true, it never causes the ssh-agent to get started on OS X. After my modification, it checks to see if it's matching the value launchctl will set on login, and if so, continues setting up the connection. During the ssh-agent startup process SSH_AUTH_SOCK gets changed.

I admit this piece could have potentially been cleaner (although less readable perhaps) by combining conditionals so that it doesn't require repetition of code. This was the best I could do at the moment and both of these changes are tested to be working as expected on my system running Mavericks 10.9.5.

Darwin tduzan-LT.local 13.4.0 Darwin Kernel Version 13.4.0: Wed Mar 18 16:20:14 PDT 2015; root:xnu-2422.115.14~1/RELEASE_X86_64 x86_64

Since the patch is merely providing an alternative condition under which to start the ssh-agent which is specific to OS X, it doesn't interfere with correct operation on Linux/BSD systems. I tested this last night at home as well on my personal system running FreeBSD and it seems to work as expected.
```

Tristor:
I probably should provide this patch again anyway in an updated PR before merging because I have actually fixed a weird edge case in this module in my fork that isn't present in the current PR.

belak:
Does anyone know how this interacts with linux? The launchctl stuff seems like it might cause unexpected behavior, but I haven't gotten a chance to test this yet.

johnpneumann:
@belak - Checked the diff of the `master` branch from @Tristor shows there's a difference [here](https://github.com/Tristor/prezto/blob/master/modules/ssh/init.zsh#L36) that checks the `OSTYPE` before attempting to run the command.

```bash
elif [[ "$OSTYPE" == darwin* && "$SSH_AUTH_SOCK" == "$(launchctl getenv SSH_AUTH_SOCK)" ]]; then
```

IIRC, there is a preference for using the `$(command ...)` syntax, but I may not be remembering correctly. 

I'll test this on Debian and CentOS and we can go from there.

Tristor:
Sorry for the delay in my reply, I'm currently traveling.  I have not updated this PR since the original author did not want to accept it, but have continued to maintain my fork separately with this and some other changes.  The change as present in the PR has a bug that is fixed in my fork.  The version of the openssh prezto plugin that's in my fork works fine on all OSes supported by prezto and is actively being used on OS X, FreeBSD, and Debian Linux by myself personally as well as being used by numerous of my past and current coworkers.

If someone would like to rewrite some of these changes to meet the original author's style guidelines so they can be accepted upstream that'd be great, but I currently don't have the time or inclination to do so.  I'm a bit surprised to see this getting attention 2 years later.  I started maintaining my own fork because I wanted certain things but it appeared nobody else cared to have them, so it made sense to just maintain it for myself.

tduzan:
I'm going to close this PR as it hasn't been merged in the 2 years since I've submitted it and I am maintaining my own fork at this time.  Thanks.

PR #943, opened 2015-08-04
Update submodules to latest commits

Eriner:


sorin-ionescu:
Submodules are updated periodically.

PR #948, opened 2015-08-09
Added option to skip git aliases definition

### LINKED ISSUE #947: Forced to have git aliases
Hello,

When using the git plugin to have git information in prompt there's no way to skip alias definition. Can we add an option that can be enabled to skip aliases ?

This option could also be used to skip alias definition in other plugins.

posva:
Should fix #947


belak:
I really like this since I'm not a huge fan of tons of aliases... so I'll look into using this (or something like this) fairly soon

posva:
Feel free to do however you want to 🙂 I enabled the _edit from maintainers_ option, which didn't exist back in the day 😆

belak:
This was merged in #1278 (along with some small tweaks). Thanks for the contribution!

PR #949, opened 2015-08-14
Feat/tmux session conventions

yock:
This adds a command to the tmux module to create and attach to sessions based on a naming convention, which can be overridden by parameter. `tmuxn` takes one optional argument, which is the name of the session to be created or attached to. Without the argument it looks to a shell variable `TMUX_CONVENTIONAL_SESSION_NAME` which is set via preexec hook to the basename of the current working directory.


yock:
Sorry, this was created by mistake

PR #952, opened 2015-08-19
Allow tmux session name to be passed as env var

alerque:
For normal usage this change should be a NOOP. However it enables the
use case where the user might want to predefine what session gets
connected to when the shell is launched. Running `zsh` will still net
you a connection to the last used tmux session or a new one named
"zprezto" if none exists. However `env tmux_session=myname zsh` will
connect you specifically to a session called 'myname' including creating
it if it doesn't exist.

This is particularly useful in window managers where a key binding might
open a terminal with a predefined session or even show a list of current
sessions and open a new terminal connected to that session.

### DIFF
--- a/modules/tmux/init.zsh
+++ b/modules/tmux/init.zsh
@@ -29,16 +29,16 @@ if [[ -z "$TMUX" && -z "$EMACS" && -z "$VIM" && -z "$INSIDE_EMACS" && -z "$VSCOD
 ); then
   tmux start-server
 
-  # Create a 'prezto' session if no session has been defined in tmux.conf.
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
I don't use tmux that often. I'll let users who do vote on this. There are a few other tmux-related pull requests you may want to verify.

alerque:
Thanks. There seems to be a little overlap with #607 and merging one or the other of these would require some minor rebasing to fix. Functionally however they cover different ground.

alerque:
I tried rebasing this since the conflicting PR was merged some time ago but the result didn't actually work as advertised. I'll re-open this when I figure out how to make it work again.

PR #953, opened 2015-08-20
completion: Cap max-errors at 7 to avoid hanging

### LINKED ISSUE #946: high max-errors causes prolonged max-cpu consumption
So, I don't know if the solution here is "so don't do that", but I figured it might be worth bringing it up;

https://github.com/sorin-ionescu/prezto/blob/master/modules/completion/init.zsh#L70 sets max-errors to the length of PREFIX+SUFFIX/3. Now I'm completely new to this, but on my system, this means that doing something like

```
ls aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa<tab>
```

when no file with a similar name exists in the current directory causes zsh to hang and peak on CPU% for a long time. The time needed to complete seems to grow close-to-exponentially. I know this example is pretty contrived but I think it demonstrates the issue well enough.

So, my query is: is there a better, perhaps more conservative way to set max-errors to avoid this hanging? It seems to me that deciding on a ceiling for max-errors, based on some clever heuristics might be worthwhile.

catharsis:
7 is pretty arbitrarily chosen, but seems like a reasonable tradeoff, at
least the completion no longer shows symptoms of exponential
time-growth when trying to complete something completely wrong.

This fixes #946.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -66,8 +66,9 @@ zstyle ':completion:*' completer _complete _match _approximate
 zstyle ':completion:*:match:*' original only
 zstyle ':completion:*:approximate:*' max-errors 1 numeric
 
-# Increase the number of errors based on the length of the typed word.
-zstyle -e ':completion:*:approximate:*' max-errors 'reply=($((($#PREFIX+$#SUFFIX)/3))numeric)'
+# Increase the number of errors based on the length of the typed word. But make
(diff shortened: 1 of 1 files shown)


belak:
I have no idea why this wasn't merged yet. Thanks for the contribution!

PR #958, opened 2015-08-29
python_info[version] will show python version number in the format "py-3.4.1"

mayuroks:
python_info[version] will show python version number in the format "py-3.4.1" (without the quotes.)

Also changed prompt_name_preexec to prompt_name_precmd. 
prompt_name_precmd is executed before the prompt is loaded.
This makes loading the prompt with python info sort of "proactive".

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -74,6 +74,8 @@ Functions
 
 Theming
 -------
+  - `python_info[virtualenv]` exposes virtualenv name
+  - `python_info[version]` exposes virtualenv python version (eg. py-3.4.1, py-2.7.6)
 
 To display the name of the current virtual enviroment in a prompt, define the
 following style in the `prompt_name_setup` function.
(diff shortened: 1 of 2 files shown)


sorin-ionescu:
Why not use a Python version manager? The *py-* prefix can confuse users who have installed other Python implementations than CPython.

mayuroks:
**Why not use a Python version manager?**
Python version will only show up when a virtualenv is activated. If a user is not using python module or if he hasn't activated virtualenv it wont show up.

**The py- prefix can confuse users who have installed other Python implementations than CPython.**
My pull request is only about python version and not about different interpreters. But I get your context. If you have a better format please recommend.

belak:
I've closed this in favor of the version included in master which gets the information from the `python` executable on the path in all instances.

However, I'd be open to improvements which only display the version when in a virtualenv or when using something like pyenv. Thanks for your contribution!

PR #959, opened 2015-08-30
Add io.js support to the node module so iojs/node can be output

robbiet480:
This adds `$node_info[fork]` which will contain either `node` or `iojs`, depending on what is currently active according to nvm.

### DIFF
--- a/modules/node/README.md
+++ b/modules/node/README.md
@@ -21,13 +21,15 @@ Functions
 Theming
 -------
 
-To display the version number of the current Node.js version, define the
+To display the version number or fork of the current Node.js or io.js version, define the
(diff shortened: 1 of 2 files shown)


sorin-ionescu:
Why have different styles for forks?

ezintz:
I think this can be closed as there is now node 4.1.0 out (iojs has been merged back).

robbiet480:
Yup, this can be closed now. Sorry I never responded to your question @sorin-ionescu, somehow missed the notification. The reason I added this is because I was working in projects with specific requirements about node v0.12 vs node v0.10 vs io.js.

PR #963, opened 2015-09-04
For VI folks: Remap ESC to something else

dvcrn:
Added the option to remap ESC to something else. Having `fd` on all my editors to enter command mode, this felt a little out of place :smile:

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -247,6 +247,13 @@ bindkey -M vicmd "v" edit-command-line
 bindkey -M vicmd "u" undo
 bindkey -M vicmd "$key_info[Control]R" redo
 
+
+# Allow remapping of command-mode key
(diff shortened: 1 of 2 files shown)


sorin-ionescu:
It makes more sense to just add `bindkey -M viins "foo" vi-cmd-mode` to *zshrc* or *editor/info.zsh* directly instead of adding another zstyle.

dvcrn:
The idea behind this was that not many people know zsh / zstyle scripting (me included) but still want the vim features they are used to. For example, until recently I didn't even know that I could rebind ESC to something else. 

Having a distinct zstyle setting there is telling the user directly when he activates vim-mode that he could also rebind the ESC key if he wants.

But yeah, I guess a documentation entry or a comment inside editor explaining how to do that should be ok too.

belak:
I think a better fix for this would be to add a `.zshrc.local` and use that to set any additional keybinds. There's some discussion in #1346

Issue #204, opened 2012-06-19
Don't alias grc to git rebase --continue


ghost:
I am not so sure if this effects many users, but there already is a programm called [grc](http://kassiopeia.juls.savba.sk/~garabik/software/grc.html), therefore I don't think that it is a good idea to [alias grc](https://github.com/sorin-ionescu/oh-my-zsh/blob/master/modules/git/alias.zsh#L105) to 'git rebase --continue' in the git module. But as already said I am not so sure if this effects many users...

PR #964, opened 2015-09-04
Fix grc alias to anything but grc so Generic Colourizer won't fail

leesavide:
The git module's aliases cause an issue when also using the Generic Colourizer tool, causing commands like make to instantly fail, since grc is called to colorize make's output. Other commands are also affected from this; the one I noticed first was make.

The grr alias was something I came up with just as a quickfix; If there's a better alias to give it, please feel free to change it. 

Unless you want git to growl :smiley:


sorin-ionescu:
This is a known issue, see #204. Also, there some issues with `grc`, see #413.

leesavide:
Ah, thank you, didn't realize it was already added as an issue. :+1:

PR #974, opened 2015-09-16
git: add alias that allows cd to the git-root

cgundogan:
This is an alias that I use quite often to change quickly to the top-level directory of a git repo.
I thought it might be a good idea to push this upstream, so that others can also benefit from it.


Eriner:
I ask to have `g..` be considered over `grt`, as it is more intuitive.

cgundogan:
@Eriner I adjusted the `grt` alias to be `g..` and amended the change immediately. You are right that it is more intuitive this way. Thanks for the feedback.

Eriner:
No problem; sorry for not providing attribution correctly in my original PR, I didn't find it when searching. Great alias :+1:

paulmelnikow:
Old thread, but thought I would share the version I use.

```sh
# Inspired by http://jeetworks.org/node/52

cdup=$(git rev-parse --show-cdup)
# git rev-parse returns non-zero when not in a git repository
if [ $? -eq 0 ]; then
    # and an empty string when you're already at the root
    target=${cdup}${1}
    if [ -n "$target" ]; then
        cd $target
    fi
fi
```

It supports the usage above, and also lets you `gcd modules/prompt/functions`.

I've got a completion for it too, which is super handy, but it seems to be broken at the moment.

PR #976, opened 2015-09-22
Make "o" to be able to open multiple files

xuhdev:
Current "o" is simply an alias of the default open command on a specific platform. This commit replaces this alias with a function, which accepts arbitrary number of arguments to open multiple files with default
applications at one command.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -111,15 +111,11 @@ if zstyle -t ':prezto:module:utility:grep' color; then
 fi
 
 # Mac OS X Everywhere
-if [[ "$OSTYPE" == darwin* ]]; then
-  alias o='open'
-elif [[ "$OSTYPE" == cygwin* ]]; then
+if [[ "$OSTYPE" == cygwin* ]]; then
   alias o='cygstart'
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
I never saw a need for this.

xuhdev:
You can open multiple files in one single shot, e.g. open two image files

sorin-ionescu:
I have never had that problem on the Mac. I can either use QuickLook to preview them or open them with the existing code since `open` on the Mac takes program arguments.

xuhdev:
Yes, I admit it is a problem if `open` takes arguments. OK, I can rename the function for open multiple files, or we can put this information in a wiki or something and forget about this PR. But I do believe this is useful, e.g. it would be a quick way in the case there are 10000 images you only want to check 2 of them.

sorin-ionescu:
I don't see a need for this when you can do `o *.jpg` right now, at least on a Mac. Also, Zsh file name generation can filter the files you want to open.

xuhdev:
@sorin-ionescu I don't see how zsh file name generation helps here. If you run `o a.png b.png`, it currently won't work without this PR.

PR #977, opened 2015-09-23
update submodules

antoineco:
Three submodules have been updated since the last commit:
* completion
* syntax-highlighting
* "pure" prompt


antoineco:
indirectly, this includes the changes proposed in #932

sorin-ionescu:
Thanks but I update them periodically, and by the time I do, such pull requests are already out of date.

PR #981, opened 2015-10-03
Persist dirstack across sessions

nomasprime:
Self-explanatory. Found it really useful so thought I'd submit a PR.

### DIFF
--- a/modules/directory/init.zsh
+++ b/modules/directory/init.zsh
@@ -28,3 +28,19 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
 
 alias d='dirs -v'
 for index ({1..9}) alias "$index"="cd +${index}"; unset index
+
+if (is-at-least 4.1) \
+    && zstyle -t ':prezto:module:directory' persist-dirstack
+then
+    if [[ -f ~/.zdirs ]] && [[ ${#dirstack[*]} -eq 0 ]]; then
+        echo "\n" >> ~/.zdirs
+        dirstack=( ${(uf)"$(< ~/.zdirs)"//"$HOME"[[:space:]]/} )
+    fi
+
+    function _persist_dirstack {
+        dirs -pl >! ~/.zdirs
+    }
+
+    autoload -Uz add-zsh-hook
+    add-zsh-hook chpwd _persist_dirstack
+fi
--- a/modules/directory/README.md
+++ b/modules/directory/README.md
@@ -24,6 +24,15 @@ Aliases
   - `d` prints the contents of the directory stack.
   - `1 ... 9` changes the directory to the **n** previous one.
 
+Settings
+--------
+
+### Persist Dirstack
+
+To persist the dirstack across sessions, add the following to *zpreztorc*:
+
+    zstyle ':prezto:module:directory' persist-dirstack 'yes'
+
 Authors
 -------
 
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -42,6 +42,13 @@ zstyle ':prezto:load' pmodule \
 # zstyle ':prezto:module:autosuggestions:color' found ''
 
 #
+# Directory
+#
+
+# Persist dirstack
+# zstyle ':prezto:module:directory' persist-dirstack 'yes'
+
+#
 # Editor
 #
 


sorin-ionescu:
You might not want to clobber `chpwd`. Perhaps use `add-zsh-hook`.

habitullence:
Hi Sorin,

Why would clobbering `chpwd` be bad?

I'll look into `add-zsh-hook`, thanks.

sorin-ionescu:
`chpwd` can be used, and, in fact, is used elsewhere.

```sh
function _directory_read_stack {
...
}
autoload -Uz add-zsh-hook
add-zsh-hook chpwd _directory_read_stack
```

habitullence:
Yeah, that's really cool. Hopefully I'll have a few minutes to get worked in the next couple of days.

habitullence:
Hi @sorin-ionescu, I've finally had the chance to update this.

It's based on the implementation you suggested - has been a good way to learn a bit of zsh.

I've also prevented duplication of the home directory, bug common to all other implementations I've seen. The regex is a bit a different from what I'm used to, could use a second look.

sorin-ionescu:
It's fine, but I do not personally need it, and no one else has voted for its inclusion. To be included, it must be documented in the README and enabled/disabled via a `zstyle`.

habitullence:
I know I'm not the only one that would find this useful. Hopefully as more features are added, zprezto's user base will grow so more will contribute and vote.

I'll update accordingly.

sorin-ionescu:
Prezto's user base grows without adding new features, especially features that are used by their authors and no one else. I, and I suspect users, favour stability over new features to prevent the Wild, Wild, West of Oh-My-Zsh.

habitullence:
I think you took my previous comment the wrong way, I'm trying to be constructive.

No one else has commented on this PR suggesting that we're the only people looking at it, rendering neutral consensus. I can understand you want to protect against bloat but there _are_ other users, potential at least, that would find this feature useful - search for "persist dirstack" if you want evidence of this. If you take this into account, general consensus is in favour. Even if you discard general consensus, you can toggle the feature.

Certainly, as a user that favours stability, I don't see how this feature would impact that: if we're talking about operational stability, you've already said that the PR looks "fine"; if we're talking about codebase stability, I don't see it but if that's the case, look at what's happening with vim and neovim.

I have many, many more arguments why this feature should be included but, ultimately, the decision is yours. I can't really spend any more time on this.

malikoth:
I'll weigh in with a +1 for this

habitullence:
Thanks @malikoth, made my day. It's really useful, switch shells and, boom, all you recent dirs only a cd +n away.

malikoth:
@belak What are your thoughts on this one?

belak:
@malikoth Same as on the other one, I've been a bit busier than I'd like lately because I'm in the middle of a move and starting a new job. If I don't get to this by the end of this weekend, feel free to ping me again.

Additionally, if anyone ever just wants a quick look or wants to chat, feel free to send me a message on Freenode.

modules/directory/init.zsh
```diff
@@ -28,3 +28,19 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
 
 alias d='dirs -v'
 for index ({1..9}) alias "$index"="cd +${index}"; unset index
+
+if (is-at-least 4.1) \
```
belak: From the prezto readme:

```
Prezto will work with any recent release of Zsh, but the minimum required version is 4.3.17.
```

belak: Essentially, I don't think the `is-at-least` check is needed in this instance.

modules/directory/init.zsh
```diff
@@ -28,3 +28,19 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
...
+    && zstyle -t ':prezto:module:directory' persist-dirstack
+then
+    if [[ -f ~/.zdirs ]] && [[ ${#dirstack[*]} -eq 0 ]]; then
+        echo "\n" >> ~/.zdirs
+        dirstack=( ${(uf)"$(< ~/.zdirs)"//"$HOME"[[:space:]]/} )
```
belak: There's some black magic here... also, the zsh docs don't make it easy to look this kind of stuff up. :/

belak: Ok, `(uf)`:

- f: Split the result of the expansion at newlines.
- u: Expand only the first occurrence of each unique word.

Then `//` appears to be find and replace all matches... Is this to make sure the home directory isn't in the dirstack? Should we really be removing the home dir if the user added it via pushd (or another method)?

Maybe I'm dense, but I'm still not sure how this line accomplishes the main goal of loading the list from `~/.zdirs` or why all these random flags are needed. Could anyone clarify?

modules/directory/init.zsh
```diff
@@ -28,3 +28,19 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
 
 alias d='dirs -v'
 for index ({1..9}) alias "$index"="cd +${index}"; unset index
+
+if (is-at-least 4.1) \
+    && zstyle -t ':prezto:module:directory' persist-dirstack
+then
+    if [[ -f ~/.zdirs ]] && [[ ${#dirstack[*]} -eq 0 ]]; then
+        echo "\n" >> ~/.zdirs
```
belak: This shouldn't go here. I'm fairly certain it will cause a newline to be appended each time a new shell is opened (without changing directories). Making sure there's a newline should probably go into the `_persist_dirstack` function.

nomasprime:
Hi @malikoth and @belak,

Since writing this I've discovered a function built into zsh called [cdr](http://zsh.sourceforge.net/Doc/Release/User-Contributions.html#Recent-Directories). Makes this PR redundant.

Thanks for taking a look though.

belak:
Thanks for following up and for mentioning an alternative. Would it be worthwhile to configure cdr anywhere or should that be left as an exercise for the user?

nomasprime:
No problem.

Might be worthwhile, I certainly use it a lot.

PR #1002, opened 2015-10-24
Add clipboard module

quicknir:


modules/clipboard/init.zsh
```diff
@@ -0,0 +1,62 @@
+#
+# Integrates the zsh clipboard with the system clipboard.
+#
+# Authors:
+#   Nir Friedman <quicknir@gmail.com>
+#
+
+# If running Mac OS, set this to 1
+local IS_MAC_OS=0
```
sorin-ionescu: You don't need that. Check for `[[ "$OSTYPE" == darwin* ]]`. See the [utility](https://github.com/sorin-ionescu/prezto/blob/master/modules/utility/init.zsh#L114) module for an example.

modules/clipboard/init.zsh
```diff
@@ -0,0 +1,62 @@
...
+
+if [[ IS_MAC_OS -eq 0 ]]; then
```
sorin-ionescu: Put the check inside of the function.

modules/clipboard/init.zsh
```diff
@@ -0,0 +1,62 @@
...
+    zle .$WIDGET
+  }
+fi
+
+zle_put_widgets=(
+  vi-put-after
+  vi-put-before
+)
+
+for widget in $zle_put_widgets
```
sorin-ionescu: Pleas use the Prezto coding style.

modules/clipboard/init.zsh
quicknir: This means the if check will be taken every single time the function is called though. I guess we don't care about this cost?

quicknir:
I'll make these changes. Should I add a fresh commit here or open a new pull request? Sorry, I'm a bit new to this (github, but not git).

modules/clipboard/init.zsh
sorin-ionescu: The cost is negligible.

leoj3n:
I've been using `set clipboard=unnamed` in `.vimrc` which integrates the system clipboard with vim. In iTerm if I highlight and copy text it uses the system clipboard. So this module will make vi mode on the command line also use the system clipboard?

quicknir:
@leoj3n Yes, although it's not particular to vi mode (if you use emacs bindings om the command line it still works). It's pretty nice, if e.g. a command isn't working, then up arrow, esc, yy, alt+tab to the browser, etc.

belak:
I realize this is a super old PR, but I've just gotten around to trying it and it doesn't look like you actually bind any of the emacs zle widgets, so it currently only works with vi bindings.

If it's updated to work with both, I'd be happy to merge this.

quicknir:
@belak I'm happy to do this if someone can tell me a quick way to list all the relevant emacs widgets. I also ticked off "allow edits from maintainers" so if you are willing to add those yourself that works too. I just never use emacs bindings myself (and I don't even know all of the emacs keybindings to test).

quicknir:
@belak Hey, so I'm updating and rebasing my prezto config. Was thinking it would be nice to get this merged. Can you maybe fill in the emacs zle widgets? It should be pretty straightforward, just easier for you to test.

belak:
There's a little bit of disagreement about what modules will be in core, so we've started work on a contrib repo. I've included this module there (also under the MIT license, with your author information for attribution).

quicknir:
@belak Ok, thanks for letting me know and attributing (I found this code via snippets on the web so I'm more than happy to have those people attributed instead or additionally). I can't say I'm a fan of such approaches, but this will be ok as long as the contributed modules repo is a sub-module of prezto. If it is not then it becomes the user's job to keep them in sync (if there is ever a change in prezto's module API). At which point it's probably less effort for most people (who are often going to only be using 1-2 modules from contrib) to just copy and paste into their own files... Not a win, IMHO. More generally it would be nice if it were possible to add modules to prezto without actually dropping them inside the prezto repo (maybe it already is and I'm just uninformed). Anything that saves users commits in their prezto forks is a win IMHO.

belak:
Yep, take a look at #1483. We're still ironing out the details, and we'd love any feedback you have.

The current plan isn't to use a submodule, but once we start making stable releases of prezto, that would make more sense.

belak:
I'm going to close this for now since it was merged into the contrib repo. Thanks for your contribution!

quicknir:
Thanks for merging!

Issue #692, opened 2014-10-08
Use the POSIX function syntax


sorin-ionescu:
Presently, Prezto uses `function foo {}`, which is a KornShell (`ksh`) compatibility syntax. Switch to the POSIX `foo () {}` syntax.

See http://wiki.bash-hackers.org/scripting/obsolete for more obsolete syntax supported by both BASH and Zsh.

Eriner:
Patch for fix here: 

https://github.com/Eriner/prezto/commit/4dfba0c4b06c64dd08cddd8c7b2c6daed5991112.patch

sorin-ionescu:
Thank you.

belak:
The PR was closed, but @Eriner left this useful snippet, which should make most of the needed changes:

`find . -path *external* -prune -o -type f -print0 | xargs -0 sed -r -i 's/(function )(.*)\s\{/\2\(\) \{/g'`

belak:
Is there any reason (aside from POSIX compliance) we don't want to keep `function` at the start? It seems easier to grep for and it's much clearer what's being defined. My honest opinion would be to go in the opposite direction and add `function` to everything... but I'm fine as long as we're consistent.

Eriner:
`function` declares the function to operate as a ksh-compatible function.

belak:
Thanks for following up. Unfortunately I can't find any information to back up those claims. The man pages seem to imply that either is fine to use:

```Shell functions are defined with the function reserved word or the special syntax `funcname ()'.```

Also:

```
function word ... [ () ] [ term ] { list }
word ... () [ term ] { list }
word ... () [ term ] command
```

I can't find any information in the man pages (for current versions - maybe it was something in an old version) about how they're treated differently (though I may be missing something - it's easy to miss stuff in the zsh man pages) and both sections on function declaration in zshmisc don't mention anything about ksh or a difference between declaration types.

Bash also does the same thing: all function declaration types are treated the same.

Eriner:
> If they do behave the same and it's not considered deprecated, is there any reason to avoid it?

Again, sorry for not having any documentation on hand, but I believe there is in fact a difference in the way it's handled by the shell, something about the function scope or alias interaction, but I don't remember. I'm sure I have a comment on GitHub somewhere about it, so I'll try to dig it up.

But the reason to avoid it is because it's non-POSIX, and because it's a legacy feature support from a shell no one uses anymore, and because of that it's non-standard. I

> It seems easier to grep for

[...]

PR #1008, opened 2015-11-02
convert from ksh to POSIX function syntax

Eriner:
fix for issue https://github.com/sorin-ionescu/prezto/issues/692

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -22,7 +22,7 @@ unset min_zsh_version
 #
 
 # Loads Prezto modules.
-function pmodload {
+pmodload() {
   local -a pmodules
   local pmodule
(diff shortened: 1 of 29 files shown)


Eriner:
It may be pertinent to note that there *are* functional differences between ksh style functions and POSIX style functions, but that has to do with alias expansion. That doesn't apply in this case (and I've been running with this patch in my fork for over 2 months with 0 issues)

belak:
Because there have been so many changes since this PR was opened, I'll most likely have to go through and do some of this manually. I do appreciate the head start though!

Eriner:
I'm closing this, but the PR can be regenerated with:
`find . -path *external* -prune -o -type f -print0 | xargs -0 sed -r -i 's/(function )(.*)\s\{/\2\(\) \{/g'`

PR #1009, opened 2015-11-02
add git-ignore-add function to git module

Eriner:
takes an arbitrary number of filenames as parameters and adds them to the .gitignore file in the project root.

### DIFF
--- a/modules/git/functions/git-ignore-add
+++ b/modules/git/functions/git-ignore-add
@@ -1,0 +1,17 @@
+#
+# Adds files passed as parameters to .gitignore in project root
+#
+# Authors:
+#   Matt Hamilton <m@tthamilton.com>
+#
+
+# make sure we have a git-root
+if ! git-root &> /dev/null; then
+  print 'not in a git repository' >&2
+  return 1
+fi
+
+# we are in a git repository. add parameters to .gitignore
+for file in "$@" do
+  print "$file" >>! $(git-root)/.gitignore
+done
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -231,6 +231,7 @@ Functions
   - `git-hub-shorten-url` shortens GitHub URLs.
   - `git-info` exposes repository information via the `$git_info` associative
     array.
+  - `git-ignore-add` adds any passed paramaters to the root's .gitignore.
   - `git-root` displays the path to the working tree root.
   - `git-stash-clear-interactive` asks for confirmation before clearing the stash.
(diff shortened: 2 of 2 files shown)


sorin-ionescu:
This seems useless. Just do `print abc[tab] >>! .gitignore`.

Eriner:
@sorin-ionescu That is a much better way to handle it as opposed to the if statement. In regards to it being 'useless', this was created upon request of someone I work with. After creating it and adding it to my own zshrc, I found myself using it frequently.

It's very nice to be able to look at `git status` and see that you have files, and with one command, add many of them to your `.gitignore`. An example of this is if you are debugging a program with gdb (as well as many other utilities), they may end up creating many files, so it's nice to be able to add multiple files to the `.gitignore` at one time.

Plus, typing `git-i<tab> abc<tab>` is faster than typing `print abc<tab> >>! $(git-root)/.gitignore`

modules/git/functions/git-ignore-add
```diff
@@ -0,0 +1,18 @@
+#
+# Adds files passed as parameters to .gitignore in project root
+#
+# Authors:
+#   Matt Hamilton <m@tthamilton.com>
+#
+
+# make sure we have a git-root
+if ! git-root &> /dev/null; then
+  echo 'not in a git repository'
```
sorin-ionescu: `print "$0: not in a git repository" >&2`

modules/git/functions/git-ignore-add
```diff
@@ -0,0 +1,18 @@
...
+  return 1
+fi
+
+# we are in a git repository. add parameters to .gitignore
+for file in "$@"
+do
```
sorin-ionescu: Put the `do` on the previous line.

modules/git/functions/git-ignore-add
```diff
@@ -0,0 +1,18 @@
...
+  print ${file} >>! $(git-root)/.gitignore
```
sorin-ionescu: `print "$file"`

sorin-ionescu:
You might want to make it smarter to support more [gitignore](https://git-scm.com/docs/gitignore) syntax.

Eriner:
are you referencing supporting things like wildcard patterns? This does that. It is, afaik, impossible to have a script un-glob input, as it is parsed by the shell first. If you want to insert a wildcard, do: `git-ignore-add '*.txt'`. Making it 'smarter' in this case seems silly, as it would have to parse the current directory, and guess at what the user wants. Assuming pattern matching is what you are referencing.

Issue #763, opened 2014-12-16
compinit insecure fails silently disabling all tab completion


ClashTheBunny:
On this [line](https://github.com/sorin-ionescu/prezto/blob/master/modules/completion/init.zsh#L18):

``` zsh
# Load and initialize the completion system ignoring insecure directories.
autoload -Uz compinit && compinit -i
```

compinit fails silently for cases where homebrew is owned by somebody other than the user.  This then causes all tab completion to fail silently.

I believe that there should be some sort of error shown or direction on how to fix it.  It's not very user friendly to all of a sudden lose tab completion because I've changed the group on /usr/local to be the one that my wife and I share on my computer so that we can both use homebrew.

The problem stems from a deeper problem of ZSH not allowing a list of safe groups, but prezto shouldn't fail silently like it does.  Debian tried to deal with this through a local patch of ZSH because of debian policy, so there are other large users of ZSH that need this functionality, but it seems to be a won't fix:
http://www.zsh.org/mla/users/2012/msg00062.html
or from 10 years before that:
http://www.zsh.org/mla/workers/2001/msg02404.html

A workaround for me is to run:
`rm ~/.zcompdump && compaudit && compinit` sometime after login.  I'm probably going to delete the `-i` from the above line in my prezto, so that it at least doesn't fail silently.

bensaufley:
My autocompletion stopped working out of nowhere this week. This solved my problem. Thanks!

timotheecour:
running `rm ~/.zcompdump && compaudit && compinit` didn't work for me but just filed https://github.com/zsh-users/zsh-completions/issues/480
it works after `brew switch zsh 5.2`
autocompletion had silently and suddenly stopped working after `brew upgrade`-ing all packages which had *upgraded* zsh to 5.3 (seems like a regression)

fikovnik:
This worked for me. From some reason the completion stopped working out of a blue. Also, when I started zsh, before getting a prompt, it listed some zsh functions.

```sh
...
url-quote-magic () {
	# undefined
	builtin autoload -XUz
}
vcs_info () {
	# undefined
	builtin autoload -XUz
}
wdiff () {
	# undefined
	builtin autoload -XUz
}
zmv () {
	# undefined
	builtin autoload -XU
}
 …/~ $
```

The problem was in the insecure directories. After fixing the permissions and re-running compinit all works fine.

christianbundy:
[...]

PR #1010, opened 2015-11-02
Remove nonsensical 'security' check

Eriner:
For some reason, zshcompsys thinks that it is up to it do be a sysadmin.

By default, `compinit` checks for 'insecure' directories and files.

From `man 1 zshcompsys`:
For security reasons compinit also checks if the completion system would
use files not owned by root or by the current user, or files in directories
that are world- or group-writable or that are not owned by root or by
The current user.

People suffering from the rammifications of this can be found here:
http://www.wezm.net/technical/2008/09/zsh-cygwin-and-insecure-directories/
http://stackoverflow.com/questions/13762280/zsh-compinit-insecure-directories
http://www.zsh.org/mla/users/2008/msg00566.html
https://github.com/Homebrew/homebrew/issues/7801

To sumarize, if you have files with permissions that are 'bad'
(ex: 777 on /usr/share/folder/), then zshcompsys enables you to ignore
those directories and files.

That begs the question, why is it zshcompsys' job to police the system?
If you have 777 permissions on a folder in /usr/share/something, that
makes the sysadmin DUMB. On the flip side, there are many situations
in which it is perfectly legitimate to allow a group to have write
permissions to a directory that contains binary files.

It shouldn't be up to zshcompsys to point out that you're an idiot.

Running compaudit (the 'security' checks) also takes up a significant
chunk of time, which can be exacerbated on low-end systems.

This commit disables the compaudit, while still enabling the .zcompdump
as provided by compinit, as it acts as a cache.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -14,8 +14,8 @@ fi
 # Add zsh-completions to $fpath.
 fpath=("${0:h}/external/src" $fpath)
 
-# Load and initialize the completion system ignoring insecure directories.
-autoload -Uz compinit && compinit -i
+# Load and initialize the completion system ignoring checks of insecure directories but still dumping to cache.
+autoload -Uz compinit && compinit -C -d ~/.zcompdump
 
 #
 # Options


sorin-ionescu:
Have you suffered from this? It's pretty rare for someone to encounter it.

Eriner:
Yes, I encountered this problem, as did someone else who I was helping debug prezto/zsh. This issue is what led to the creation of my debug module.

Eriner:
The amount of time it takes to perform these 'checks' is significant. Very significant. As in, almost half of the time it takes prezto to initialize significant.

Eriner:
This is related to issue https://github.com/sorin-ionescu/prezto/issues/763

Eriner:
"solved" by zcompdump caching change.

PR #1011, opened 2015-11-02
add archive function

Eriner:


modules/archive/functions/archive
```diff
@@ -0,0 +1,72 @@
...
+  _bzip2_bin='bzip2'
+fi
+
+case "${archive_name}" in
+  (*.tar.gz|*.tgz) tar -cvf "${archive_name}" --use-compress-program="${_gzip_bin}" "${dir_to_archive}" ;;
+  (*.tar.bz2|*.tbz|*.tbz2) tar -cvf "${archive_name}" --use-compress-program="${_bzip2_bin}" "${dir_to_archive}" ;; 
+  (*.tar.xz|*.txz) tar -cvJf "${archive_name}" "${dir_to_archive}" ;;
+  (*.tar.lzma|*.tlz) tar -cvf "${archive_name}" --lzma "${dir_to_archive}" ;;
+  (*.tar) tar -cvf "${archive_name}" "${dir_to_archive}" ;;
+  (*.zip) zip -r "${archive_name}" "${dir_to_archive}" ;;
```
zaro0508: commit  7242b4ed added list and unarchive support for .jar files so maybe this should be updated to for jar as well?

belak:
It looks like the original repo has been deleted so I'm going to merge this into a tmp branch, make a few changes and cherry pick it back to master. Thanks for mentioning it @zaro0508. I forgot about this PR.

modules/archive/functions/archive
belak: How would you imagine this working? If I remember correctly, there's additional information needed when you create a jar file (stuff like the main class, etc).

belak:
Merged in fb37539f

modules/archive/functions/archive
zaro0508: It would probably require support for 'jar' command with some simple defaults.  Something like this might work..

(*.jar) jar cvf "${archive_name}" "${dir_to_archive}" ;;

belak: For now I've just used zip. That way if you use unarchive to extract a jar, archive should still work. I haven't used java in a while though, so any improvements would be recommended.

PR #1012, opened 2015-11-02
don't use backticks, use $(COMMAND) syntax

Eriner:


belak:
Thanks for your contribution! Sorry for the delay.

PR #955, opened 2015-08-25
Add note in README about how to workaround installation with existing .zshrc

tombh:
This is a long-standing issue, first brought up in #231.

As many people have reported, the current installation instructions **silently** fail if the user has an existing `.zshrc`.

This PR simple makes the situation clearer in the README.

PR #1014, opened 2015-11-02
Add note in README about existing zshrc

Eriner:


Eriner:
dupe of https://github.com/sorin-ionescu/prezto/pull/955

Issue #701, opened 2014-10-16
Move prompt_sorin_pwd out of sorin theme


sorin-ionescu:
The function is used by other themes. Remove duplication by extracting it into its own function file.

Eriner:
Done.

https://github.com/Eriner/prezto/commit/8a3bedb0b4d91d464dc3be0d8bb62a9a0dfb01f4.patch

jeffwidman:
@belak is this closed by your modified version of #1016 that you merged into master?

belak:
Yep, good catch

PR #1016, opened 2015-11-02
move common prompt to external function

Eriner:


modules/prompt/functions/promptpwd
```diff
@@ -0,0 +1,15 @@
...
+#
+# prompt setup function common to many prompts
+# moved to external function to ruduce rudundant code
+#
+local prompt="$1"
+local pwd="${PWD/#$HOME/~}"
+ 
+if [[ "$pwd" == (#m)[/~] ]]; then
+	# because this function is sourced from other scripts, the vars must be exported
+  export _prompt_"${prompt}"_pwd="$MATCH"
```
sorin-ionescu: Why are you exporting it?

Eriner:
I believe the comment is self-explanatory. To go into more detail, if it isn't exported, the variable isn't available to the calling function ex: `prompt_damoekri_setup`.

An alternative way of handling it would be to create `local dir` and assign what is currently being exported to that, then at the end of the function, `echo ${dir}`. Inside the prompt setups, they could just assign the output of that to a variable and use that instead.

If you know of a better way to handle it, please let me know and I will commit a change.

Eriner:
This would close issue https://github.com/sorin-ionescu/prezto/issues/701

amerlyq:
Hm, I have slightly different approach to this problem as whole -- I have splitted all prompt into keyparts to make it modular. See at [prompt](https://github.com/amerlyq/airy/blob/master/cfg/shell/zsh/prompt) and [prompt.d](https://github.com/amerlyq/airy/tree/master/cfg/shell/zsh/prompt.d). Not ideal, of course, but allows much more control.

sorin-ionescu:
Your prompts will share styling. In this case, it's just one function, no styling.

modules/prompt/functions/promptpwd
```diff
@@ -0,0 +1,15 @@
+#
+# prompt setup function common to many prompts
+# moved to external function to ruduce code rudundancy
```
belak: Nit: should be "reduce"

belak:
I'm merging this into a temp branch so I can make a tweak and cherry-pick it to master. I guess weird things happen when a repo is deleted while there are still PRs open.

belak:
This has been merged into master along with some spelling fixes and a fix to ensure extendedglob is set for this function. Thanks for your contribution!

sorin-ionescu:
Be careful with those functions. They are not always completely identical. I was trying to think of a naming scheme since promptpwd is not descriptive enough.

belak:
Maybe shortpwd or something similar?

sorin-ionescu:
It's still doesn't say how short it is and what's included.

PR #1019, opened 2015-11-03
Add homebrew-installed completions to fpath

jedahan:
Used the node submodule as inspiration.

### DIFF
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -31,3 +31,7 @@ alias caski='brew cask install'
 alias caskl='brew cask list'
 alias casks='brew cask search'
 alias caskx='brew cask uninstall'
+
+if (( $+commands[brew] )); then
+  fpath=(`brew --prefix`/share/zsh/site-functions $fpath)
+fi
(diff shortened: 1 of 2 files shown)


modules/homebrew/init.zsh
```diff
@@ -31,3 +31,7 @@ alias caski='brew cask install'
 alias caskl='brew cask list'
 alias casks='brew cask search'
 alias caskx='brew cask uninstall'
+
+if (( $+commands[brew] )) && [[ -d "$(brew --prefix 2>/dev/null)" ]]; then
```
sorin-ionescu: Cache the result of `brew --prefix`; don't call it twice. Can you have `brew` without being installed?

jedahan:
Is this a bit better?

Would we want to change the node module to work the same? https://github.com/sorin-ionescu/prezto/blob/master/modules/node/init.zsh#L14-L15

sorin-ionescu:
Yes, `brew` is written for Ruby 1.8, which is slow. Calling it twice is not good. A separate pull request should be opened for that.

jedahan:
No longer using prezto, so can't test this anymore

PR #1021, opened 2015-11-04
Add alias-tips plugin

jedahan:
I really like [the alias-tips plugin](https://github.com/djui/alias-tips), thought other people might too

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "modules/prompt/functions/pure"]
 	path = modules/prompt/external/pure
 	url = https://github.com/sindresorhus/pure.git
+[submodule "modules/alias-tips/external"]
+	path = modules/alias-tips/external
+	url = https://github.com/jedahan/alias-tips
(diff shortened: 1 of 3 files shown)


Eriner:
I can't speak for anyone else, but this would drive me mad.

liujoey:
Pleeeeeeeeeeeeeeeeeeease add this! Prezto added so many great alias, but it really lack of a great alias reminder!

I couldn't remember how may times I have to go back to this repo just searching for alias defined in various modules.

diraol:
@liujoey you may just add this alias to a 'rc' file:
`alias sa='alias | grep -i'`
and then do `sa git` or `sa log`, for example.

belak:
Interestingly enough, we already include `alias sa='alias | grep -i'` in the `utility` module. 😄

liujoey:
@diraol @belak Thank you! Partially solved my headache.

But @belak comment is exactly the point, if alias-tips is included, then @diraol would find out that alias `sa` had been included already. 😄

Issue #1026, opened 2015-11-11
Module syntax-highlighting with prezto causes zsh to exit immediately


mjbshaw:
I'm on OS X 10.10.5 using zsh 5.0.5 (x86_64-apple-darwin14.0). This happens in both iTerm2 and Apple's Terminal app. If I add the syntax-highlighting module to my `.zpreztorc`, so the module loading looks like this:

```
zstyle ':prezto:load' pmodule \
  'environment' \
  'terminal' \
  'editor' \
  'history' \
  'directory' \
  'spectrum' \
  'utility' \
  'completion' \
  'syntax-highlighting' \
  'prompt'
```

Here's me logging in with a pristine prezto install (I just cloned and set it up, so it's prezto at revision 165879d4b30dbf90d436b02a73786a28519d9e93):
![works](https://cloud.githubusercontent.com/assets/1204698/11099782/35ff74de-8862-11e5-86d6-d0e4f905f52f.gif)

And here's me logging in after enabling syntax-highlighting (adding `'syntax-highlighting' \` is the only change I've made to my `.zpreztorc`):
![broken](https://cloud.githubusercontent.com/assets/1204698/11099803/4f789cce-8862-11e5-97ad-3b28bd915e8a.gif)

Note that zsh now immediately exits upon login.

mjbshaw:
Looks like it's an issue in zsh 5.0.5. Running `/usr/bin/zsh` (which is v5.0.5) results in a segfault:

```
$ /usr/bin/zsh
[1]    89512 segmentation fault  /usr/bin/zsh
```

However, if I run zsh 5.1 there is not segfault and the syntax-highlighting module works just fine. I don't know if this is something that can be worked around or if it's not worth the energy given v5.1 works.

sorin-ionescu:
Which highlights have you enabled?

mjbshaw:
I haven't explicitly enabled any highlights, which I assume means just the main highlight should be loaded by default.

This is with a fresh clone of prezto (with the only modification being the `  'syntax-highlighting' \` line in `.zpreztorc` to load the module; I didn't make any other modifications to change which highlights were used or anything like that), so I am just using whatever prezto with syntax-highlighting defaults to.

sorin-ionescu:
No script should segfault Zsh. Can you upgrade Zsh? You might want to open an issue on the zsh-syntax-highlighting page.

mjbshaw:
[...]

Issue #888, opened 2015-05-28
Sorin theme causes shell to hang.


breerly:
My shell gets into a state where I can't do anything, I have to close the window and open a new shell. If I just hold `^C`, this is my output:

```
~ ❯❯❯
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
~ ❯❯❯                                                                         ⏎
^C^C^C^C^C^C^C^C^C^C^C^C^C
```

At the end, you can see that the shell is busted, there is no way to get out. 

Any help debugging/fixing would be much appreciated.
- OSX 10.10.3
- github.com/breerly/dotfiles
- prezto - https://github.com/sorin-ionescu/prezto/tree/a275db55d6fdc8bf92a4722e95196bbfbf44cfae

breerly:
Some other notes:

* able to recreate by sending a bunch of `ENTER` keys, after like 4 seconds it hangs
* switching to `bash`, I'm unable to reproduce in any way
* disabling prezto by removing the initialization from zshrc, I'm unable to reproduce in any way

breerly:
Looks like it might be plugin related? If I narrow the plugins down to just these, I can still reproduce:

```
zstyle ':prezto:load' pmodule \
   'environment' \
   'terminal' \
   'editor' \
   'prompt'`
```

breerly:
[...]

PR #1028, opened 2015-11-12
Fix the sorin prompt theme

ziegs:
This fixes https://github.com/sorin-ionescu/prezto/issues/1026 by changing `zle reset-prompt` to `zle .reset-prompt`. This matches the behavior of the `pure` theme.


sorin-ionescu:
There are other themes that use `reset-prompt` without the period, some external, for which I do not have write access. There is also a `reset-prompt` in *modules/editor/init.zsh*. Also, I think #1026 is more related to #888 than ZLE.

aluttik:
Oh my god @ziegs, thank you. It took me hours to figure out why my terminal kept crashing at startup and this fixed everything. :clap:

PR #1029, opened 2015-11-12
Add fasd as an external module

### LINKED ISSUE #780: fasd module requires fasd to be installed?
Hi there, 

I checked the [README](https://github.com/sorin-ionescu/prezto/tree/master/modules/fasd) and have a question about the following:

> The Prezto Fasd configuration differs from the default. The default aliases have been disabled

Does this mean that I don't have to install the default fasd package from my distribution's repository?

I followed the steps and added fasd after the completion module. However, the shortcut `j` does not work.

Excerpt of my `~/.zpreztorc`:

```
zstyle ':prezto:load' pmodule \
  'environment' \
  'terminal' \
  'editor' \
  'history' \
  'directory' \
  'spectrum' \
  'utility' \
  'completion' \
  'fasd' \
  'git' \
  'prompt' 
```

kaluzki:
Will fix #780

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "modules/prompt/functions/pure"]
 	path = modules/prompt/external/pure
 	url = https://github.com/sindresorhus/pure.git
+[submodule "modules/fasd/external"]
+	path = modules/fasd/external
+	url = https://github.com/clvv/fasd
--- a/modules/fasd/init.zsh
+++ b/modules/fasd/init.zsh
@@ -9,10 +9,8 @@
 # Load dependencies.
 pmodload 'editor'
 
-# Return if requirements are not found.
-if (( ! $+commands[fasd] )); then
-  return 1
-fi
+# Source module file.
+source "${0:h}/external/fasd" || return 1
 
 #
 # Initialization


sorin-ionescu:
`fasd` comes with a variety of package managers; so, do many other directory jumping scripts. I'm inclined to reject this unless someone gives me a really good reason.

jeffwidman:
Agree with @sorin-ionescu - I'd prefer that prezto doesn't include fasd by default--rather install/manage it externally with `homebrew`, `yum`, etc

dryvenn:
If I may, I have a reason for this.
Sometimes I SSH onto a server where:
- my dotfiles are synced by NIS, so I expect to have my complete usual setup;
- fasd is not installed and I don't have root privileges to change that.

It is very frustrating indeed to use `j` repeatedly because of one's mnemonic memory and to get an error every time.

belak:
This seems like a reasonable change to me... avoiding an external dependency would make it easier to deploy this everywhere.

Alternatively, we could fall back to using the external module if the fasd command doesn't exist...

sorin-ionescu:
Actually, I was thinking of removing this in favour of z. It seems overkill. I only use it for directory changes.

malikoth:
lol  I have an `alias z='fasd_cd'` in my config because I don't need the advanced functionality much, either.

belak:
I'm going to merge this for now and update it to fall back to using the submodule.

Removing it entirely should be a separate issue.

Thanks for your contribution!

belak:
The additional change is in 04bfb51.

PR #1034, opened 2015-11-16
Use dynamic path for zsh binary

christianbundy:


sorin-ionescu:
`chsh` only works for shells listed in `/etc/shells`.

christianbundy:
All correctly installed shells should have an entry in `/etc/shells`.

``` console
$ which zsh
/usr/bin/zsh
$ grep zsh /etc/shells
/bin/zsh
/usr/bin/zsh
```

The current README isn't perfect when you have an old version of zsh installed at `/usr/bin` (like OS X does), but you're right that my suggested change isn't perfect when you have zsh in your path but not `/etc/shells`.

Maybe we could fix this and resolve #734 with an `install.sh` with something like this?

``` shell
#!/usr/bin/env zsh

# symlink runcoms
setopt EXTENDED_GLOB
for rcfile in "${ZDOTDIR:-$HOME}"/.zprezto/runcoms/^README.md(.N); do
  ln -s "$rcfile" "${ZDOTDIR:-$HOME}/.${rcfile:t}"
done

# find zsh path
for enabled in $(where zsh); do
  while read available; do
    if [[ "$enabled" == "$available" ]]; then
      zsh_path="$available" && break 2;
    fi;
  done < /etc/shells;
done;

# set default shell
chsh -s "$zsh_path";

# start prezto
exec "$zsh_path" --login --interactive;
```

A two-step install (`git clone` and `exec install.sh`) would also make life easier for people who install Prezto often.

sorin-ionescu:
By the way, you cause `$commands[zsh]` instead of `which`. #231 is the issue for the install script.

christianbundy:
Gotcha, thanks. How should I move forward on this PR?

sorin-ionescu:
If you wish to submit a pull request for an install script, do so.

PR #1036, opened 2015-11-22
Add CMake module

bchretien:
This module simplifies the creation of build directories with CMake. The install prefix, build prefix and build profiles can be specified in the prezto configuration file. Clang is supported, and CMake completion is enabled for the alias.

### DIFF
--- a/modules/cmake/README.md
+++ b/modules/cmake/README.md
@@ -1,0 +1,44 @@
+CMake
+=====
+
+Initializes build directories for [CMake][1].
+
+### Usage
+
+```sh
+# In the root directory of a CMake project:
+$ mb
+# Or with additional flags:
+$ mb -DSOME_EXTRA_FLAGS="foo"
+```
+
+### Options
+
+```sh
+# Set the install prefix.
+zstyle ':prezto:module:cmake' install-prefix '/usr'
+
+# Set the build prefix.
+zstyle ':prezto:module:cmake' build-prefix '_build'
+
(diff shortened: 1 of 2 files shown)


sorin-ionescu:
Thank you for the contribution. Would not an extensible make module that will handle `make`, `CMake`, `MSBuild`, et al. be better instead of a proliferation of modules for each one?

bchretien:
Any idea of what you would add in there for `make`? I usually have an alias that automatically adds a relevant `-jN` option to `make`, but that's about it. Same for `ninja`.

I'm not familiar with `MSBuild`, but if it's Microsoft's equivalent of CMake, then yes, that would make sense. How should we structure the module then?

As for CMake specifically, I'll add another option to specify a generator automatically (e.g. `Unix`, `Ninja`).

sorin-ionescu:
I don't use `make` that often; so, no.

I envision the following directory structure.

```
modules/make/init.zsh
modules/make/ant.zsh
modules/make/cmake.zsh
modules/make/make.zsh
modules/make/maven.zsh
modules/make/msbuild.zsh
modules/make/premake.zsh
modules/make/rake.zsh
modules/make/scons.zsh
modules/make/waf.zsh
modules/make/xcode.zsh
```

We source in *init.zsh* the various make systems when detected.

People will contribute.

belak:

I'm closing this for now because I think it would be a better fit for an external module but thank-you for taking the time to submit this.

PR #1040, opened 2015-11-25
Use less destructive git push --force-with-lease.

jfilip:
`alias gpf='git push --force'` is destructive. Using the `--force-with-lease` option is much safer and prevents accidentally destroying another's work in the same branch.

Also added a `gpF` alias to allow using `--force` if really desired.

See this post about the flag: http://movingfast.io/articles/git-force-pushing/

### DIFF
--- a/modules/git/alias.zsh
+++ b/modules/git/alias.zsh
@@ -117,7 +117,8 @@ alias gmt='git mergetool'
 
 # Push (p)
 alias gp='git push'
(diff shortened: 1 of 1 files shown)


akrawchyk:
@sorin-ionescu is it possible to get this merged?

belak:
Thanks for your contribution! This is a nice compromise between safety and usefulness and your including of `gpF` made this a no brainer.

PR #1043, opened 2015-11-29
Added support for python venv display in paradox prompt

kireledan:
Just a simple tweak to the paradox prompt to show the name of the python virtualenv being worked on at the end of the prompt. 

2015-11-29--10 55 26--0x57_scrot

### DIFF
--- a/modules/prompt/functions/prompt_paradox_setup
+++ b/modules/prompt/functions/prompt_paradox_setup
@@ -52,6 +52,10 @@ function prompt_paradox_build_prompt {
     prompt_paradox_start_segment green black '${(e)git_info[ref]}${(e)git_info[status]}'
   fi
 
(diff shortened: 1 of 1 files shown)


belak:
Thanks for the contribution! Nice to get some of these small tweaks in.

wojtekgalaj:
It would be nice if this was not obligatory. As is I'm seeing the following error in my console:

prompt_paradox_preexec:2: command not found: python-info

I am not a python dev and would rather not have to worry about figuring out why my rather stock macos Sierra installation is missing said command.

belak:
Sorry about that @wojtekgalaj. I've pushed a fix here: https://github.com/sorin-ionescu/prezto/commit/b963484fe7a929db309946d777988c39f082ebf6

PR #1054, opened 2015-12-24
Add new functions (wherefrom and settings) to OS X module

josh-h:
These functions have been handy in my environment, so I thought I'd share with a larger audience.
- New function 'wherefrom' sets the OSX metadata in file(s) for Finder to
  show where the file was downloaded from
- New function 'settings' updates the Terminal window with a different settings
  configuration, e.g. to quickly change the window theme.
- Fix a missing local variable in the 'tab' function

### DIFF
--- a/modules/osx/README.md
+++ b/modules/osx/README.md
@@ -23,6 +23,8 @@ Functions
   - `osx-rm-dir-metadata` deletes .DS\_Store, \_\_MACOSX cruft.
   - `osx-ls-download-history` displays the Mac OS X download history.
   - `osx-rm-download-history` deletes the Mac OS X download history.
+  - `wherefrom` sets the Mac OS X download location for a file.
+  - `settings` changes the terminal settings for the current window.
 
 Authors
 -------
--- a/modules/osx/functions/_settings
+++ b/modules/osx/functions/_settings
@@ -1,0 +1,10 @@
+#compdef settings
+
+_known_settings=$(osascript 2>/dev/null <<EOF
+    tell application "Terminal" to get name of every settings set
+EOF
+)
+
+_arguments \
+  "1:Terminal Settings:( \${(s:, :)_known_settings} )"
+
--- a/modules/osx/functions/_wherefrom
+++ b/modules/osx/functions/_wherefrom
@@ -1,0 +1,5 @@
+#compdef wherefrom
+
+_arguments \
+  "1:Download URL" \
+  "*:Downloaded File:_files"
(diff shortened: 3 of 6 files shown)


sorin-ionescu:
`wherefrom` data is found in an SQLite database; see `osx-ls-download-history`.
`settings` may be too generic? Any conflicts? I have prefixed other osx-specific functions.

josh-h:
Thanks for responding to the PR.

The proposed wherefrom function let's me tell Finder where a file came from. For example, My Firefox.dmg that I just downloaded is from https://download-installer.cdn.mozilla.net/pub/firefox/releases/43.0.4/mac/en-US/Firefox%2043.0.4.dmg, ls-download-history and Finder both report that address. However, OSX doesn't let me change that value to something sane, like http://firefox.com. Worse, when I open the disk image and install the application I've lost the information on where that file came from. So, wherefrom can be used to tell Finder where /Applications/Firefox.app was originally downloaded from.

Now it's pretty obvious where firefox was downloaded from, but imagining other scenarios this quickly becomes a handy tool. Eg. where did I get this font from, where did this script originally come from, where did this desktop photo come from, or where did on our corporate wiki did this document from, etc.

Second, regarding settings as a function name. The name is derived from the applescript function name used for changing terminal options. The GUI uses 'profile' instead, but that clashes with bash profile concepts. I am not familiar with any tools that use settings as a name. While I fear 'osx-settings' implies a very different purpose. I'd entertain changing the name to 'terminal-settings' and adding an alias named 'ts'.

belak:
"settings" appears to be a terminal launcher, so maybe just `osx-terminal` would be more accurate. I'm also not sure about wherefrom... it just seems a bit too specialized.

ezintz:
Not sure about the "downloaded from". Why do you care? I mean for software, especially that ones which are modifying operating system resources, it might be worth... but Mac OS (not sure since which versions) is trying to verify downloaded Mac OS applications on its own.

Mac OS settings

However, I think that the terminal settings are valuable.

josh-h:
Having some distance I agree that 'settings' is too general of a name. I have no objection with 'osx-terminal'.

My actual use case for the downloaded from tool is to add metadata to images, and documents that I've downloaded. Sometimes the actual download link shown in Finder is a lengthy amazon S3 link, whereas I will change it to the site that I actually obtained it from. In other cases the download is a zip and the document itself looses the download location.

In the year plus that I've created this it hasn't had any feedback, so if it isn't generally useful to this audience then I'm fine with the pull request being declined.

For what it's worth I use settings daily as a tool to quickly show that I'm ssh into a different host by using a different window style. While I rarely use wherefrom these days as most docs that I download are now cloud hosted (e.g. google docs) so my personal need for that has declined.

PR #1056, opened 2015-12-30
paradox issue with powerline font cleared

LoganArnett:
Found this issue when updating my local copy of the prezto project. None of my powerline fonts were working to provide some of the more specific icons in the paradox theme and found that they were missing in the main prompt setup. Put them back in and everything is working again. Just thought I would open this up for you guys if you want to use it.

I really love the whole project and having a much nice visual in my terminal, still jump from prompt to prompt but paradox is typically what I default to.


sorin-ionescu:
What looks good for you now will break it for someone else.

LoganArnett:
So is there not a common ground that will work for everyone?

sorin-ionescu:
There is not. That is the nature of these fonts.

PR #1057, opened 2016-01-05
Ignore .DS_Store

yasu100:
#### Ignore ".DS_Store".
".DS_Store" is OS X 's meta data file.


sorin-ionescu:
I have that in **~/.gitignore**. Put it there.

Issue #532, opened 2014-01-11
Using zsh-autosuggestions with prezto?


leoj3n:
Fish-like autosuggestions for zsh: https://github.com/tarruda/zsh-autosuggestions

Is it possible to use this widget in tandem with prezto and, if so, How?

Sorry for the shallow question, but I'm not sure what to try. Thank you.

breerly:
update here? would love love love fish-like autosuggestions in prezto

sorin-ionescu:
I have not played with it. Try it and let me know.

breerly:
This ticket seems to indicate that it works fine:

https://github.com/tarruda/zsh-autosuggestions/issues/37

sorin-ionescu:
I will try the following and report.

```zsh
zstyle ':prezto:load' pmodule \
    "history" \
    "completion" \
    "history-substring-search" \
    "syntax-highlighting" \
    "autosuggestions"
```

breerly:
:+1:

breerly:
@sorin-ionescu how did it fare?

breerly:
@sorin-ionescu I think you need history after syntax highlighting like so:

```bash
zstyle ':prezto:load' pmodule \
    "history" \
    "completion" \
    "syntax-highlighting" \
    "history-substring-search" \
    "autosuggestions"
```

sorin-ionescu:
No, syntax-highlighting should be after history-substring-search since it modifies a function supplied by history-substring-search.

breerly:
If there's anything I can do to help get movement here let me know - I'm kind of a shell noob but I'll do my best.

sorin-ionescu:
#972 is a good alternative for this as well as zsh-history-substring-search for those who prefer more sophisticated history browsing.

breerly:
@sorin-ionescu did you manage to get autocompletions working?

sorin-ionescu:
It kind of works. There are coloring issues when combined with syntax-highlighting.

breerly:
I can fiddle with it - what's the best way to install it?

sorin-ionescu:
Just wait for me to push into master and update Prezto after this issue is closed.

breerly:
Got it - I'll play around with it then and see how well I can get it working.

sorin-ionescu:
[...]

PR #1060, opened 2016-01-07
Add autosuggestions module

ohcibi:
The modules uses zsh-autosuggestions as an external submodule and defines
useful keybindings for vi-mode.


sorin-ionescu:
It's in the works. See #532. There have been issues with it.

PR #1064, opened 2016-01-17
git: Add revert to special action contexts

zmwangx:
`revert` is very similar to `cherry-pick` and has a `sequence` variant, so detection code is directly adapted from that of `cherry-pick`.

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -299,6 +299,8 @@ The following contexts must be enabled with the following zstyle:
 | rebase               |    value    | Rebasing
 | rebase-interactive   |    value    | Rebasing interactively
 | rebase-merge         |    value    | Rebasing merge
+| revert               |    value    | Reverting
+| revert-sequence      |    value    | Reverting sequence
 
 First, format the repository state attributes. For example, to format the branch
 and remote names, define the following styles.
--- a/modules/git/functions/git-info
+++ b/modules/git/functions/git-info
@@ -5,7 +5,7 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
-# Gets the Git special action (am, bisect, cherry, merge, rebase).
+# Gets the Git special action (am, bisect, cherry, merge, rebase, revert).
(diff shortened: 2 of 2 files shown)


sorin-ionescu:
I've used `git revert` in the past, but it has never failed for me; so, I have never witnessed the special action.

zmwangx:
`git-revert` basically applies an inverted diff, and like all actions that apply diffs, it's possible to fail. A very simple example:

```diff
From 270c130efc8cab8c0998fb44fe4f5176baa86921 Mon Sep 17 00:00:00 2001
From: Zhiming Wang <zmwangx@gmail.com>
Date: Tue, 1 Mar 2016 19:29:45 -0800
Subject: [PATCH 1/2] Add file `foo' with two lines

---
 foo | 2 ++
 1 file changed, 2 insertions(+)
 create mode 100644 foo

diff --git a/foo b/foo
new file mode 100644
index 0000000..1191247
--- /dev/null
+++ b/foo
@@ -0,0 +1,2 @@
+1
+2
-- 
2.7.2


From 0d8e4fff45d89354f301f36b78ff9c03a361b10d Mon Sep 17 00:00:00 2001
From: Zhiming Wang <zmwangx@gmail.com>
Date: Tue, 1 Mar 2016 19:30:16 -0800
Subject: [PATCH 2/2] Remove second line of foo

---
 foo | 1 -
 1 file changed, 1 deletion(-)

diff --git a/foo b/foo
index 1191247..d00491f 100644
--- a/foo
+++ b/foo
@@ -1,2 +1 @@
 1
-2
-- 
2.7.2

```

Apply these two patches, then

```
> git revert HEAD~1
error: could not revert 270c130... Add file `foo' with two lines
hint: after resolving the conflicts, mark the corrected paths
hint: with 'git add <paths>' or 'git rm <paths>'
hint: and commit the result with 'git commit'
```

belak:
I haven't seen this before, but it seems to line up with the documentation and it's easily reproducible. Thanks for your contribution!

Issue #1071, opened 2016-01-27
Tmux in iTerm and $TERM_PROGRAM


bax-:
Tmux integration works on local machine and on SSH connection.
However if I do `su - root`, the integration doesn't work. Only if I do `tmux -CC`myself.
Is there a way to disable checking for the $TERM_PROGRAM or recognize it in these situations?

bax-:
The variable `TERM_PROGRAM` isn't being taken to the root when you change to superuser via `su - root`, even if it is set on the client:
```
SendEnv TERM_*
```
And on the server:
```
AcceptEnv TERM_*
```
As this only works for the user the session is initiated with and not for root, I suggest to remove the `$TERM_PROGRAM` condition (as oh-my-zsh does).

sorin-ionescu:
I don't use `tmux`. So, others who use it will have to comment.

bax-:
the most universal and simple fix is the commit I referenced

mkatsevVR:
kinda related question - is TERM_PROGRAM supposed to be automatically passed to the remote server? because it doesn't for me (env | grep TERM_PROGRAM is empty) and therefore the integration doesn't actually work

cebrusfs:
I just notice this issues is long stand such that the iterm integration is broken for many years.... not body use this functionality on server?

re 
> is TERM_PROGRAM supposed to be automatically passed to the remote server? because it doesn't for me (env | grep TERM_PROGRAM is empty) and therefore the integration doesn't actually work

No, the modern SSH doesn't have this by default. People have to setup `SendEnv` on client and `AcceptEnv` on server.

I think this `TERM_PROGRAM` is quite a hack workaround here. It only applies for the users that only enable tmux integration on local and share the config with non-iterm terminal. I personally support #1073 fix as ohmyzsh also implement in that way.

PR #1073, opened 2016-01-28
remove TERM_PROGRAM condition

bax-:
The variable `TERM_PROGRAM` isn't being taken to the root when you change to superuser via `su - root`, even if it is set on the client:
```
SendEnv TERM_*
```
And on the server:
```
AcceptEnv TERM_*
```
As this only works for the user the session is initiated with and not for root, I suggest to remove the `$TERM_PROGRAM` condition (as oh-my-zsh does).

see
https://github.com/sorin-ionescu/prezto/issues/1071

### DIFF
--- a/modules/tmux/init.zsh
+++ b/modules/tmux/init.zsh
@@ -17,9 +17,7 @@ fi
 # Auto Start
 #
 
-if ([[ "$TERM_PROGRAM" = 'iTerm.app' ]] && \
-  zstyle -t ':prezto:module:tmux:iterm' integrate \
-); then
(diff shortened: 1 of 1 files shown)


belak:
Correct me if I'm wrong, but won't this break over ssh connections when using iTerm? Maybe a better fix would be changing it to use some sort of os detection?

bax-:
it does break when you're on a virtual console and open tmux automatically when doing `su - root`.
I haven't found a middle ground yet.
I guess os also wouldn't be taken over when I'm SSHing and then doing `su - root`?

cebrusfs:
+1 for this change

PR #1075, opened 2016-01-31
Fix unsets

d125q:
This pull request fixes two `unset` lines.  One was unsetting a nonexistent variable, the other was not present at all.


belak:
Thanks for your contribution!

PR #1087, opened 2016-02-22
Add alias to update Prezto

fabioperez:
The command for updating Prezto is somewhat big to type and remember:

    git pull && git submodule update --init --recursive

This PR introduces a alias for updating Prezto:

    prezto-update


nasenatmer:
`git submodule update --init --recursive` [is defined as gSI](https://github.com/sorin-ionescu/prezto/blob/master/modules/git/alias.zsh#L164) already

PR #1088, opened 2016-02-23
Detach from tmux if attached, before attach

zeroasterisk:
If you are in a tmux session, and auto-start is enabled
Then you `sudo su` and then `su <username>` it will re-attach, in a loop.
This resolves that problem, by forcing a detach before a (re)attach.


belak:
This is a breaking change, but it makes sense. Sorry it's taken so long to get to this. Thanks for your contribution!

PR #1094, opened 2016-03-08
Git tag aliases

nomasprime:
There aren't any git tag aliases. If you use git, you probably use tags. If you don't use aliases, you're inefficient.

This PR will make you more efficient.

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -193,6 +193,11 @@ Aliases
   - `gSu` fetches and merges the latest changes for all submodule.
   - `gSx` removes a submodule.
 
+### Tag
+
+  - `gt` lists tags or creates tag.
+  - `gtl` lists tags matching pattern.
+
 ### Working directory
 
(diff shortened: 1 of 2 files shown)


modules/git/README.md
```diff
@@ -193,6 +193,11 @@ Aliases
   - `gSu` fetches and merges the latest changes for all submodule.
   - `gSx` removes a submodule.
 
+### Tag
+
+  - `gt` lists tags or creates tag.
+  - `gtg` lists tags matching pattern.
```
sorin-ionescu: Why not `gtl`?

habitullence: I was toying between the two, maybe gtl is better - it's not a grep.

belak:
If you can rebase this against master, these seem like a worthwhile addition.

belak:
Merged in 4325f74f9dfc83d01409c78fd126a5d29f0a3a55. Thanks for your contribution!

PR #1095, opened 2016-03-08
Update ssh module for multi session hosts

mchataigner:
create ssh_agent_env with current user id in file name to avoid collisions with other users
create ssh_agent_sock with current user id in file name to avoid collisions with other users

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -14,10 +14,10 @@ fi
 _ssh_dir="$HOME/.ssh"
 
 # Set the path to the environment file if not set by another module.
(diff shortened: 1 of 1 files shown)


sorin-ionescu:
`$TMPDIR` should be set to the user's temporary directory, not `/tmp`, the fail-safe; so, there should not be a collision.

You can use `$UID` instead of `id -u`.

mchataigner:
Ok, I changed to UID instead of id -u.
So I agree with you that one should set TMPDIR, but if it is not set it will collide.
With this change it will never collide.
Thanks.

sorin-ionescu:
Presto should set it for you.

belak:
Thanks for your contribution!

PR #1097, opened 2016-03-10
Add alias for db reset

Fipo:
Suggestion for alias for `bundle exec rake db:reset`
It was useful in developer mode when dev need to change working branches.
ref: https://github.com/rails/rails/blob/v4.0.5/activerecord/lib/active_record/railties/databases.rake#L72

What do you think guys ?


belak:
This seems like a disaster waiting to happen. If you really want this, feel free to add it to your own zshrc but I would not be comfortable merging this in here.

PR #1099, opened 2016-03-12
add "--human-readable" in ls alias

YurixAnd:
Improves output of ls

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -62,7 +62,7 @@ alias type='type -a'
 # ls
 if is-callable 'dircolors'; then
   # GNU Core Utilities
-  alias ls='ls --group-directories-first'
+  alias ls='ls --group-directories-first --human-readable'
 
   if zstyle -t ':prezto:module:utility:ls' color; then
     if [[ -s "$HOME/.dir_colors" ]]; then


dlucian:
would not work for me:

```shell
~/Homestead ❯❯❯ ls --human-readable                                                                                                 ⏎ master ◼
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

belak:
I don't want to merge this in if it breaks things on some platforms. It may be possible to use `-h` or something else though... or improve the check for GNU coreutils... since dircolors is available on OSX as well which seems to use the BSD versions of most commands by default.

belak:
@dlucian What platform are you on where that didn't work for you?

sorin-ionescu:
@belak It's fine. It's only set when GNU Coreutils `ls` detected.

belak:
I'd rather not merge this, as it could get us into a situation where the default alias doesn't allow all options to be overridden as there's not a `--no-human-readable` flag I'm aware of. There was recently another PR where we removed a flag for diff like this because it made certain operations impossible without using `\diff` and I'd like to avoid a similar situation here.

Thank-you for the idea, but I think this may be a better fit if you want to put this in your own zshrc or look into adding it to some of the ls aliases we have defined.

PR #1104, opened 2016-03-16
Update copyright

rhencke:


belak:
Thanks for your contribution, but I think we'll end up doing this when we move the README to a separate file.

PR #1105, opened 2016-03-18
Add utility alias for pydf

maethor:
Hi,

This PR simply adds a utility alias for pydf (if installed), comparable to the alias for htop.

Thank you for prezto, it's awesome!


belak:
Thanks for your contribution!

PR #1110, opened 2016-03-26
Show rev info when in detached HEAD

hristoast:
If you're in a detached HEAD state, it's not a symbolic ref, and so
`$branch` was left empty.


sorin-ionescu:
It's not a branch then. It should fall back on commit or position. See the Git module readme.

PR #1120, opened 2016-04-09
Correct vim-style autosuggest-accept binding.

braham-snyder:
Update the vim-style command line binding for Ctrl-E such that it
properly accepts autosuggestions (instead of doing nothing).


sorin-ionescu:
It may be better to add `vi-add-eol` to https://github.com/zsh-users/zsh-autosuggestions/blob/master/src/config.zsh#L30. What say you, @ericfreese?

PR #1124, opened 2016-04-11
Add notice about potential shadowing of the gb command

astropanic:
When you're using the git module, it will override your `gb` command from the [Go build tool](https://getgb.io/about/).
I added just a note about it.


belak:
Thanks for your contribution!

PR #1130, opened 2016-04-19
Add perl-info function & plenv subcommand aliases

delphinus:
## perl-info function

I implemented `perl-info` with referring to `ruby-info` in the `ruby` module. The usage is almost the same with `ruby-info`. This function load info from `perlbrew` or `plenv` or pure `perl` executable.

``` sh
ruby-info
RPROMPT+=' %F{5}${ruby_info[version]}%f'
perl-info
RPROMPT+=' %F{6}${perl_info[version]}%f'

zstyle ':prezto:module:ruby:info:version' format 'version: %v'
zstyle ':prezto:module:perl:info:version' format 'version: %v'
```

## plenv subcommand aliases

Like Pelbrew, with [plenv](https://github.com/tokuhirom/plenv) you can manage plural binary installations of Perl. plenv has almost the same interfaces with rbenv.


leoj3n:
Looks good! This would be faster for `perlbrew`:

```
if (( $+commands[perlbrew] )); then
  version="${PERLBREW_PERL##*perl-}"
fi
```

leoj3n:
Could also source Perlbrew here similar to the `node` module sources NVM:

```
# Load Perlbrew into the shell session.
if [[ -s "${HOME}/perl5/perlbrew/etc/bashrc" ]]; then
  source "${HOME}/perl5/perlbrew/etc/bashrc" 

  # Load Perlbrew completion.
  if [[ -s "${HOME}/perl5/perlbrew/etc/perlbrew-completion.bash" ]]; then
    source "${HOME}/perl5/perlbrew/etc/perlbrew-completion.bash"
  fi
fi
```

delphinus:
I will move this to https://github.com/zsh-users/prezto

PR #1141, opened 2016-05-08
Change VIRTUALENVWRAPPER_VIRTUALENV_ARGS example

jeffwidman:
`--no-site-packages` option is deprecated, so we should use something else
as an example so we don't mislead folks: 
https://virtualenv.pypa.io/en/latest/reference.html#cmdoption--no-site-packages

I wasn't sure what might be best as the new example, this looked reasonable.

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -55,11 +55,10 @@ is used. Replace *Developer* with your projects directory.
     export PROJECT_HOME="$HOME/Developer"
 
 The variable `$VIRTUALENVWRAPPER_VIRTUALENV_ARGS` tells virtualenvwrapper what
-arguments to pass to `virtualenv`. For example, set the value to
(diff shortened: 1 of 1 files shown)


belak:
Thanks for your contribution! I think the best example of a command might be the opposite of the one in there now, `--system-site-packages` as it still references the same option but in a different way.

belak:
I went ahead and merged a change with `--system-site-packages`. Thanks for the idea!

jeffwidman:
Thanks! I saw your earlier comment and didn't have time to fix yet. 

I'm not exactly sure I like the `--system-site-packages` example--there's a reason the default was flipped. But it's not exactly clear to a n00b what the value does, so using it in the example might lead someone astray by implying it's a bad default that should be overridden. Versus something like the `--download` or another flag that is quite clear that typically you don't want it.

PR #1143, opened 2016-05-12
Paradox Python Virtualenv fix

egpbos:
Building on @kireledan's pull request, I made Python virtualenv integration in the Paradox prompt theme work a bit nicer. The python-info function should be in precmd instead of preexec (which is an error in the python module README, by the way).

Also, I put in a snake :) 🐍


egpbos:
To clarify: the fact that it was in preexec caused a bug: the prompt would not immediately update after switching to a virtualenv. This is because the python-info function looks for a environment variable that is set by virtualenv (containing the venv's name). However, when python-info is called in preexec (i.e. before the virtualenv switching command), the env var isn't there yet, and so the prompt would not be updated. Only after the next command was executed would the prompt actually show the virtualenv name. In fact, even pressing enter wouldn't help (which makes sense, since that's not an execution of anything, but it's still confusing from a user point of view). Anyway, by putting python-info in precmd, everything works fine.

kireledan:
awesome! I was wondering why it the prompt wouldn't update until another command was run. Thanks for solving that mystery.

belak:
I think when I merged the other PR conflicts may have shown up in this one. Would you be willing to rebase this?

egpbos:
Conflict solved, will this work?

belak:
Sorry about the delay! I'm not sure what happened with this...

I pushed c38c2dd9 (You are credited as the author there) which fixes the issue with preexec vs precmd, but I didn't merge the python emoji change because there's some weirdness with how different terminals handle double-wide unicode characters. Essentially there should only be one space between the python and the information, but iTerm displays it differently than other terminals and so it needs an additional space because it doesn't calculate the width of the emoji correctly.

If you find a way around this or find an emoji that isn't a double-wide character, I'd be happy to merge that change as well.

I'm going to close this for now as the main issue is fixed, but please feel free to open a new PR if you find a good way of handling the python emoji.

PR #1144, opened 2016-05-12
Added conda virtualenv support to python module.

egpbos:
Conda is both a package manager like pip (but without the tedious compilation, because binary) and an environment manager like virtualenv (inside conda environments one can still use pip, if necessary). I like it a lot better than pip+venv. This PR slightly modifies the python module to be compatible with Conda as well as virtualenv. Virtualenv support works as before, the only real changes are:
- when using conda, virtualenv things are not loaded, since the two cannot be used together (maybe they can, but I can't think of a reason why one would want this)
- when inside a conda venv, python_info[virtualenv] is set to the right conda environment name.

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -3,6 +3,19 @@ Python
 
 Enables local Python and local Python package installation.
 
+Settings
+--------
+
+This module supports virtual environments from conda and virtualenvwrapper. By default, only virtualenvwrapper is enabled. To disable virtualenvwrapper, add the following to *zpreztorc*.
+
+    zstyle ':prezto:module:python' skip-virtualenvwrapper-init 'on'
+
+Conda support is enabled by adding the following to *zpreztorc*.
+
+    zstyle ':prezto:module:python' conda-init 'on'
+
+Caution: using conda and virtualenvwrapper at the same time may cause conflicts.
+
 Local Python Installation
 -------------------------
 
--- a/modules/python/functions/python-info
+++ b/modules/python/functions/python-info
@@ -4,6 +4,7 @@
 #
 # Authors:
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
+#   Patrick Bos <egpbos@gmail.com>
 #
 
 local virtualenv_format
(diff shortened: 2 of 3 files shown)


belak:
I believe there were some changes to allow skipping of virtualenvwrapper init if an option is set, so that should probably be used, rather than the conda hacks here.

Could you provide a link to where CONDA_DEFAULT_ENV comes from? Based on that variable name, it sounds like it's a fallback and might not catch all cases, but you say it works... so maybe it's just not named very well?

Once that's done, I can take another look at merging this.

egpbos:
I'll use zstyle for activating conda as well and rebase the whole branch to current master to include skipping of virtualenvwrapper.

I can't find any official documentation for CONDA_DEFAULT_ENV online. However, the command `conda info --all` lists it. Also, in the activate script that comes with any conda environment (try `less $(which activate)` if you have conda installed) you'll find the following:

```sh
    # CONDA_DEFAULT_ENV is the shortest representation of how conda recognizes your env.
    #    It can be an env name, or a full path.
    #    Last date of change: 2016-06-21
    # If the string contains / it's a path
    if [[ "$@" == */* ]]; then
        export CONDA_DEFAULT_ENV=$(get_abs_filename "$args")
    else
        export CONDA_DEFAULT_ENV="$args"
    fi
    # CONDA_PREFIX is always the full path to the activated environment.  It is not set
    #    when no environment is active.
    # Legacy support: CONDA_DEFAULT_ENV is either env name or full path if given as path.
    #    CONDA_PREFIX is always the full path, for consistency.
    #    Last date of change: 2016-06-21
```

egpbos:
Ah, and, crucially, it also contains this:

```sh
    # if CONDA_DEFAULT_ENV not in PS1, prepend it with parentheses
    if [ $("$_CONDA_DIR/conda" ..changeps1) = "1" ]; then
        if ! $(grep -q CONDA_DEFAULT_ENV <<<$PS1); then
            if ! $(grep -q "POWERLINE" <<<$PS1); then
                export PS1="(${CONDA_DEFAULT_ENV}) $PS1"
            fi
        fi
    fi
```

That is: conda itself uses this variable to modify the prompt. That's actually how I think I found it :)

egpbos:
To clarify, the conda part in modules/python/init.zsh doesn't actually do anything except check whether conda settings are correct. If not, it tells you how to set it so that it works (see the conda code in the comment above, conda checks for the changeps1 setting, so we use that to deactivate conda's own prompt changing). The real work is done in the python-info function, just like with virtualenvwrapper.

egpbos:
Ok, done.

I also threw in some documentation for good measure :)

kepbod:
Great PR! 👍

belak:
Sorry it's taken me this long to get to this... so, this doesn't actually load conda? It just checks settings? Would it be worthwhile changing this so it *does* load conda?

kepbod:
I tested it, and it works.

egpbos:
Loading conda would be difficult to do automatically, since its installation location is often non-default (at least in my case, my preference changes regularly), sometimes it's installed system-wide, sometimes only for the user, etc. Ideally, conda itself would provide support for zsh. For bash it already adds the conda bin dir to $PATH in .bash_profile or .bashrc or something. Once conda is loaded in PATH, everything else is set, except the handling of the prompt, which is what I fixed in this PR.

kepbod:
I really want this feature.

belak:
Alright, once the conflicts are fixed, I think this will be ready

belak:
Thanks for submitting this. It's been merged in #1505.

PR #1147, opened 2016-05-16
Feat (docker): add docker aliases

akarzim:
Add some concise aliases for docker, docker-machine and docker-compose.

AFAIK, theses aliases will not shadow other system commands or presto’s module aliases.


malikoth:
Can I suggest a few more?

```
alias dki='docker images'
alias dkin='docker inspect'
alias dkpsa='docker ps -a'
alias dkR='docker run -it --rm'
alias dkRe='docker run -it --rm --entrypoint /bin/bash'

# Clean up exited containers
alias dkcc='docker rm $(docker ps -a | grep Exited | awk '"'"'{ print $1 }'"'"')'

# Clean up dangling images
alias dkci='docker rmi $(docker images -f dangling=true -q)'
```

malikoth:
This PR is absolutely essential, @sorin-ionescu !!!

jeffwidman:
See also #863 and https://github.com/felixr/docker-zsh-completion (which has instructions for installing into prezto)

saoula:
May I also suggest

```
alias dkce='docker-compose exec'
```

akarzim:
🙏  Thanks @malikoth ! I added your proposals.

I've renamed the proposed cleanup aliases to avoid conflicts with docker-compose (`dkc`) aliases.

akarzim:
🙏 Thanks @saoula ! I added your proposal.

malikoth:
What about `dkcR=docker-compose run --rm`?

akarzim:
@malikoth it's already there in my standalone repository. Prezto seems in standby and @sorin-ionescu is not reactive, so I switched to zplug. You can check my up to date docker aliases here: https://github.com/akarzim/zsh-docker-aliases

malikoth:
Thanks for the information that you've switched to zplug.  After looking at their README, it looks like I'm switching to zplug too!  (May take me awhile though.  /sigh)

facastagnini:
Hey guys this repo is not being maintained, see #1239

The community fork is here https://github.com/zsh-users/prezto, please move this issue to zsh-users and we will be able to help you!

belak:
It looks like something happened with this PR and it may have been rebased against zsh-users/prezto... so there's quite a few unrelated information in here. If you'd like this to be merged in, it would first have to be written against this repo rather than the fork.

akarzim:
@belack Done.

belak:
I think you may have misunderstood... somehow there are commits from the zsh-users/prezto repo showing up in this pull request... if you look at the commits, it says there are over 100 (along with a ton of unrelated changes and conflicts). This shouldn't be based off the zsh-users repo, just this one.

akarzim:
My bad. Something went wrong. It seems to be OK right now.

modules/docker/init.zsh
```diff
@@ -0,0 +1,218 @@
+#
+# Defines Docker aliases.
+#
+# Author:
+#   François Vantomme <akarzim@gmail.com>
+#
+
+# Return if requirements are not found.
+if (( ! $+commands[docker] )); then
+    return 1
```
belak: Please use two-space tabs to match with the rest of the project.

modules/docker/init.zsh
```diff
@@ -0,0 +1,218 @@
...
+    if [[ ! -d $1 ]]; then
+        echo "Docker machine '$1' does not exists. Abort."
+        popd
+        return 1
+    fi
+
+    if [[ -L default ]]; then
+        eval $(rm -f default)
+    elif [[ -d default ]]; then
+        echo "A default manchine already exists. Abort."
```
belak: Typo. Should be machine.

modules/docker/init.zsh
```diff
@@ -0,0 +1,218 @@
...
+        popd
+        return 1
+    fi
+
+    eval $(ln -s $1 default)
+    popd
+}
+
+#
+# Aliases
```
belak: There are a ton of aliases here. Could you do something similar to what the git module does? Aliases go in alias.zsh and init.zsh calls `source "${0:h}/alias.zsh"` to load it.

belak:
Also (and I realize this is a much larger task than the small ones given in the review) would it be worthwhile reorganizing the aliases and perhaps changing them to be similar in style to the git aliases.

In the git module, it focuses around a single letter (g) followed by an action/object (as an example c is commit) followed by what's actually being done. So `gp` is git push and `gpF` is git push --force (it uses a capital letter because it's potentially destructive).

Git is also aliased to just g. Is there any reason Docker couldn't be aliased to just d rather than dk? Are there conflicts with existing commands? If not, I feel like at this point docker may be important enough to warrant a single letter.

When a module gets this many aliases, it can be worthwhile to think carefully about what's worth including and what isn't, so we can avoid some of the alias hell that has been discussed in many oh-my-zsh issues.

I'd personally like to see similar aliases for docker. As an example, `dck` could be kill container (meant as docker container kill) rather than `dkk` for docker kill. This is, of course, just an example. There may be better ways to do this... and I'd probably be alright with merging this in... I'd just like to do everything possible to avoid confusing aliases if at all possible.

Does anyone else have opinions on this? I realize this may be a bit more than most people wanted to read, but I'd be happy to hear what everyone else thinks as well.

modules/docker/init.zsh
akarzim: You're right. Done.

modules/docker/init.zsh
akarzim: Fixed.

belak:
It looks like the submodules may have accidentally been updated. If that needs to be done again, I'd appreciate that going in a separate PR.

akarzim:
@belak in fact, I tried to match git aliases pattern: `dk` for docker, `dkc` and `dkm` for docker-compose and docker-machine, adding more letters for more precise aliases. A capital letter is used for alternative commands with option, like `dke` vs `dkE` ; or subsets like like `dkC`, `dkV`, `dkI`.

I choosed `dk` because `d` is already an alias from the directory module, and some sub-commands based on `d` would overwrite system commands (ie. `dc` is an arbitrary precision calculator). So `dk` appears as the safer alias for docker.

belak:
Alright, I'm going to merge this to a temporary branch, squash the commits, and undo the submodule changes before merging it into master, but it looks pretty much ready.

akarzim:
Great ! Thank you Kaleb.

belak:
Merged in 8a4333103dc990400a5d800a49f48418d6a53449. Thanks for all your work on this!

PR #1148, opened 2016-05-22
fixed vivisual key binding

Nargoth:
v is used as default for character wise visual mode since zsh 5.0.8; using emacs style C-x C-e to edit current command in full editor in editor module instead, which allows to use v as meant to

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -240,8 +240,8 @@ fi
 # Vi Key Bindings
 #
 
-# Edit command in an external editor.
-bindkey -M vicmd "v" edit-command-line
+# Edit command in an external editor emacs style (v is used for visual mode)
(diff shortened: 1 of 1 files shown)


belak:
Ah, interesting... maybe there would be a better bind than C-x C-e though? Especially if this is for vimode... this seems like a very non-vi style key binding.

malikoth:
I think `C-x C-e` was picked because that's the default command for doing this in `bash`, which a lot of people got used to before we all found `zsh` and `prezto` and `vimode`.  When I saw this PR and realized that visual mode works perfectly fine on the command line as long as you have a key for it, I put this in my personal config as well.  And I couldn't think of any better keybinding for it than `C-x C-e` either.

belak:
Thanks for submitting this! Sorry it took so long to merge.

PR #1150, opened 2016-05-25
Add "sa" alias to make it easy to search aliases

malikoth:
I've had this alias in my own `.zshrc` for awhile to remind me of my own aliases when I can't remember.  There are many new aliases as a new convert to Prezto, and having this alias has really sped up my ability to learn all the things I get with Prezto.

Don't remember how to do a `git reset --hard HEAD`?  Just type `sa reset` and you get:

``` bash
gcR='git reset "HEAD^"'
giR='git reset --patch'
gir='git reset'
gwR='git reset --hard'
gwr='git reset --soft'
```

Ah, I wanted to type `gwR HEAD`.  Easy!

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -75,6 +75,7 @@ Aliases
   - `pu` changes the directory and pushes the old directory onto the stack
     (`pushd`).
   - `rm` removes files and directories interactively.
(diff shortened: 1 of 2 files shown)


belak:
This is pretty clever

PR #1151, opened 2016-05-27
add deer support - a ranger like tool

rphillips:
[deer](https://github.com/Vifon/deer) is a ranger-like file navigator for zsh.

This PR adds the `deer` module with `alt-k` completion functionality.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "modules/autosuggestions/external"]
 	path = modules/autosuggestions/external
 	url = https://github.com/tarruda/zsh-autosuggestions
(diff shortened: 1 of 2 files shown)


rjcoelho:
alt+k doesn't work for me (probably iterm2), I suggest configure it

```
#in .zpreztorc
zstyle ':prezto:module:deer' key-bindings '\ek'

#in deer/init.zsh
zstyle -s ':prezto:module:deer' key-bindings 'key_bindings'
bindkey '$key_bindings' deer
unset key_bindings
```

facastagnini:
duplicates #1220

belak:

PR #1173, opened 2016-07-29
Add pyenv version to python-info

leoj3n:


jeeftor:
Doesn't this functionality already exist?

leoj3n:
@jeeftor Does what already exist? What do you mean by "this"?

jeeftor:
I meant wasn't there already `pyenv` support in the model.  What you want is to basically get the current version regardless of whether you are in a virtual environment? 

As long as its not a performance hit i don't see an issue with the patch (except you need to resolve your conflicts) :)

leoj3n:
No, this was added to make pyenv version info available to the prompt.

```
❯ pyenv version
system (set by /Users/leoj/.pyenv/version)

❯ python --version
Python 2.7.13
```

belak:
Would it make more sense to just get the version info from the python executable in the path? Or maybe when we're in a virtualenv as well?

I guess the main question for me would be "when is this information important to show in a prompt?"

belak:
I've closed this in favor of the version included in master which gets the information from the `python` executable on the path.

If you want to update it to include support for pyenv, an updated PR would be welcome. Thanks for the idea!

belak:
I actually just pushed a new commit to add support for pyenv. Thanks for your contribution!

PR #1176, opened 2016-08-09
Remove manb from completion filename

leoj3n:
Not sure if `manb` is a function that was removed, but it's not being completed by the completion script so should be removed from the filename.


belak:
I may be missing something, but can you even get mand and manp to complete properly? I haven't been able to at all, so I'm sure there's something a bit off somewhere.

belak:
Finally got it working. This seems fine. Thanks for your contribution!

PR #1177, opened 2016-08-18
Export $TMPPREFIX in non-login interactive shells

leoj3n:
The current implementation causes arguably confusing behavior; at least it tripped me up:

[asciicast](https://asciinema.org/a/dfigewhqtq3238i4l1uaasdou)

Unless there is a reason `$TMPPREFIX` should differ when saying `zsh` vs `zsh -l`, lack of `export` for this profile variable was probably an oversight.


belak:
I'm closing this now because TMPPREFIX isn't used any more.

PR #1183, opened 2016-08-23
Better TTY logic for zlogin and zlogout

leoj3n:
Handles more cases more gracefully.

### DIFF
--- a/runcoms/zlogin
+++ b/runcoms/zlogin
@@ -14,10 +14,13 @@
   fi
 } &!
 
-# Print a random, hopefully interesting, adage.
-if (( $+commands[fortune] )); then
-  if [[ -t 0 || -t 1 ]]; then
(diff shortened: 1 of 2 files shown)


facastagnini:
Hi @leoj3n , your PR got merged in the community maintained fork of this repo.

belak:
Thanks for your contribution! What's the advantage of using -o INTERACTIVE over -t 0? It seems like they both accomplish similar goals in different ways.

jeffwidman:
Nudge @leoj3n

leoj3n:
@belak `-t 0` could be a non-interactive terminal.

belak:
This makes sense. Sorry it's taken me so long to get around to merging it!

Is there any reason to check -t 2 as well as -o INTERACTIVE? It seems like checking if the terminal is interactive would be enough.

PR #1185, opened 2016-08-26
Lazy load nvm

rtfpessoa:
nvm takes on average half of a second to load, which is more than whole prezto takes to load.

This can be noticed when you open a new shell.

To avoid this, we are creating placeholder function for nvm, node, and all the node packages previously installed in the system to only load nvm when it is needed.

This code is based on the scripts:
* https://www.reddit.com/r/node/comments/4tg5jg/lazy_load_nvm_for_faster_shell_start/d5ib9fs
* http://broken-by.me/lazy-load-nvm/
* https://github.com/creationix/nvm/issues/781#issuecomment-236350067

### DIFF
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -128,6 +128,7 @@ zstyle ':prezto:module:prompt' theme 'sorin'
 #   'main' \
 #   'brackets' \
 #   'pattern' \
+#   'line' \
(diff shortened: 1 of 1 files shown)


rtfpessoa:
Moved to a separated branch so I can merge other stuff to my master.

PR #1178, opened 2016-08-18
Add support for nodenv

chauncey-garrett:

PR #1186, opened 2016-08-26
Lazy load nvm

rtfpessoa:
nvm takes on average half of a second to load, which is more than whole prezto takes to load.

This can be noticed when you open a new shell.

To avoid this, we are creating placeholder function for nvm, node, and all the node packages previously installed in the system to only load nvm when it is needed.

This code is based on the scripts:
- https://www.reddit.com/r/node/comments/4tg5jg/lazy_load_nvm_for_faster_shell_start/d5ib9fs
- http://broken-by.me/lazy-load-nvm/
- https://github.com/creationix/nvm/issues/781#issuecomment-236350067

### DIFF
--- a/modules/node/init.zsh
+++ b/modules/node/init.zsh
@@ -6,13 +6,43 @@
 #   Zeh Rizzatti <zehrizzatti@gmail.com>
 #
 
+lazy_load_nvm() {
+  NVM_DIR=$1
+  # Skip adding binaries if there is no node version installed yet
+  if [ -d $NVM_DIR/versions/node ]; then
+    NODE_GLOBALS=(`find $NVM_DIR/versions/node -maxdepth 3 \( -type l -o -type f \) -wholename '*/bin/*' | xargs -n1 basename | sort | uniq`)
+  fi
+  NODE_GLOBALS+=("nvm")
+
+  load_nvm () {
+    # Unset placeholder functions
+    for cmd in "${NODE_GLOBALS[@]}"; do unset -f ${cmd} &>/dev/null; done
+
+    # Load NVM
+    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+
+    # (Optional) Set the version of node to use from ~/.nvmrc if available
+    nvm use 2> /dev/null 1>&2 || true
+
+    # Do not reload nvm again
+    export NVM_LOADED=1
+  }
+
+  for cmd in "${NODE_GLOBALS[@]}"; do
+    # Skip defining the function if the binary is already in the PATH
+    if ! which ${cmd} &>/dev/null; then
+      eval "${cmd}() { unset -f ${cmd} &>/dev/null; [ -z \${NVM_LOADED+x} ] && load_nvm; ${cmd} \$@; }"
+    fi
+  done
+}
+
 # Load manually installed NVM into the shell session.
 if [[ -s "$HOME/.nvm/nvm.sh" ]]; then
-  source "$HOME/.nvm/nvm.sh"
+  lazy_load_nvm "$HOME/.nvm"
 
 # Load package manager installed NVM into the shell session.
 elif (( $+commands[brew] )) && [[ -d "$(brew --prefix nvm 2>/dev/null)" ]]; then
-  source $(brew --prefix nvm)/nvm.sh
+  lazy_load_nvm "$(brew --prefix nvm)"
 
 # Return if requirements are not found.
 elif (( ! $+commands[node] )); then


modules/node/init.zsh
```diff
@@ -6,13 +6,43 @@
...
+    # (Optional) Set the version of node to use from ~/.nvmrc if available
+    nvm use 2> /dev/null 1>&2 || true
+
+    # Do not reload nvm again
+    export NVM_LOADED=1
+  }
+
+  for cmd in "${NODE_GLOBALS[@]}"; do
+    # Skip defining the function if the binary is already in the PATH
+    if ! which ${cmd} &>/dev/null; then
```
leoj3n: Nice, I was wondering how you were going to handle package binaries like https://www.npmjs.com/package/which

rtfpessoa: I am also trying #1178. Nodenv seems much faster on init and uses a different method for managing different node versions.

Since I already use rbenv which is much better than rvm and nodenv is a fork. I decided to give it a try.

Seems good.

chauncey-garrett: I've had great success with `nodenv` and started using it because I was also familiar with `rbenv`. I too also tried `nvm` and also found it noticeably slowed down my shell.

joaosa:
@sorin-ionescu Any chance this could get merged? :)

before:

```
/usr/bin/time zsh -i -c exit
2.75 real         1.20 user         1.00 sys
```

after:

```
/usr/bin/time zsh -i -c exit
0.63 real         0.25 user         0.38 sys
```

sectioneight:
I also run a custom fork solely so I can include this patch.

facastagnini:
This PR got merged in the community maintained fork of this repo.

modules/node/init.zsh
belak: This could probably be improved by using `$+commands[$cmd]`.

rtfpessoa: Let me know if you want any change I can update

modules/node/init.zsh
```diff
@@ -6,13 +6,43 @@
 #   Zeh Rizzatti <zehrizzatti@gmail.com>
 #
 
+lazy_load_nvm() {
+  NVM_DIR=$1
+  # Skip adding binaries if there is no node version installed yet
+  if [ -d $NVM_DIR/versions/node ]; then
+    NODE_GLOBALS=(`find $NVM_DIR/versions/node -maxdepth 3 \( -type l -o -type f \) -wholename '*/bin/*' | xargs -n1 basename | sort | uniq`)
```
belak: Does this work on OSX as well? I seem to remember the find command being fairly limited.

modules/node/init.zsh
```diff
@@ -6,13 +6,43 @@
...
+    nvm use 2> /dev/null 1>&2 || true
+
+    # Do not reload nvm again
+    export NVM_LOADED=1
+  }
+
+  for cmd in "${NODE_GLOBALS[@]}"; do
+    # Skip defining the function if the binary is already in the PATH
+    if ! which ${cmd} &>/dev/null; then
+      eval "${cmd}() { unset -f ${cmd} &>/dev/null; [ -z \${NVM_LOADED+x} ] && load_nvm; ${cmd} \$@; }"
```
belak: I understand why this is here, but stuff like this worries me...

modules/node/init.zsh
```diff
@@ -6,13 +6,43 @@
...
+    for cmd in "${NODE_GLOBALS[@]}"; do unset -f ${cmd} &>/dev/null; done
+
+    # Load NVM
+    [ -s "$NVM_DIR/nvm.sh" ] && . "$NVM_DIR/nvm.sh"
+
+    # (Optional) Set the version of node to use from ~/.nvmrc if available
+    nvm use 2> /dev/null 1>&2 || true
+
+    # Do not reload nvm again
+    export NVM_LOADED=1
```
belak: Is NVM_LOADED needed since all the lazy-loaded functions have been unset already?

modules/node/init.zsh
belak: I'd be willing to bet that this doesn't work in either osx or linux. There should be a way to use extended glob syntax to do something like this but only use zsh internals.

modules/node/init.zsh
belak: Unsetting this specific command and using NVM_LOADED shouldn't be needed because we unset all the functions in load_nvm anyway, so we don't need to track it.

belak: Additionally, because this is using eval, I'd feel a lot better if there was a way to know that the command names were sanitized.

modules/node/init.zsh
belak: Depending on what you're trying to do, `is-callable` from the helper module or `$+commands` would be preferable.

ericbn:
Replaced `nvm` by `nodenv` and my startup time when from 2s to 0.2s. Why bother about `nvm`?

belak:
What's the status on this? It doesn't look like there's been an update by the author more than a month after changes were requested. Does anyone else want to take this up?

rtfpessoa:
I completely missed this @belak.

It will probably take me some days to update this. If you want to do it I added permissions for the owners to edit this pr.

belak:
No worries. Thanks for the response! I don't know if I'll get to it soon either, so no rush. Just trying to follow up on a few loose ends. :)

modules/node/init.zsh
rtfpessoa: What would be the part not working in OSX or Linux?

modules/node/init.zsh
rtfpessoa: Any suggestions?

modules/node/init.zsh
belak: After looking a bit closer, it looks fine. I've just run into strange issues with find on linux and osx being incompatible.

jeffwidman:
I'm hesitant about this, and I'm hoping @creationix would be willing to take a look before we merge this.

I used to follow the `nvm` repo for a while, and there were often users complaining about speed, but every time they suggested a "fix" @creationix pointed out problems with their solution.

That said, I don't work on any node projects right now, so haven't stayed up to date with `nvm`, so perhaps this implementation works fine, I just don't know and given the context I'd rather be cautious.

creationix:
There has been quite a bit of feature and code bloat since I handed nvm over to a new maintainer.  If you look at the contribution graph my initial implementation is a fraction of the overall code.

If I were to write nvm again knowing what I now know, it would look a lot like nvs. https://github.com/jasongin/nvs

There is no need to have all the heavy logic in shell.  The only thing that has to be shell script is changing the current `PATH` value.  Everything else can be done in a sub-process on-demand.

creationix:
I'm afraid I don't know enough about the current nvm implementation to give good feedback on your lazy load approach.  Good luck!

jeffwidman:
Thanks for weighing in @creationix! I actually meant to tag the new maintainer, as he is the one whose comments I've seen and didn't realize he wasn't you... so tag @ljharb 😄

creationix:
@ljharb 

> "code bloat" is pretty uncharitable

Sorry, sometimes I forget the negative connotations with that phrase.  You've done great work!

jeffwidman:
Thanks for chiming in @ljharb, much appreciated

PR #1191, opened 2016-09-04
Provide a simple installation script

ezintz:
Hello,

from time to time, I am updating my prezto. Today was one of those days and somehow I found on the issues and pull requests a lot of installers that did not made it into master. This was kind of a motivation for me to try it as well, since I have a similar installer for my `dotfiles`.

#### _Which steps is the installer performing?_
- Asking the user if he wants to proceed with the installation
- Clone the repository to `PREZTO_DIRECTORY="${ZDOTDIR:-$HOME}/.zprezto"`
- Sym linking the runcoms and creating backups of the existing files 

#### _Why did you change the headings in the README.md?_

Somehow I wanted to have the headings in a consistent way to be identified when updating the file.

#### _Why does the installer does not change the shell for the user?_

Users that are switching from any other shell may have it already customized and might want to migrate later. Besides that there are many Terminals which provide the user to run a customized command which means: chsh -s may have no effect at all.

I guess that somehow the TRAPINIT needs somehow to do a cleanup, but for a first version it should be okay. If you see any issues with this pull request please let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,37 +1,40 @@
-Prezto — Instantly Awesome Zsh
-==============================
+# Prezto — Instantly Awesome Zsh
+
 
 Prezto is the configuration framework for [Zsh][1]; it enriches the command line
 interface environment with sane defaults, aliases, functions, auto completion,
 and prompt themes.
 
-Installation
-------------
+## Prerequisites
 
-Prezto will work with any recent release of Zsh, but the minimum required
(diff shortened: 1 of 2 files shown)


ezintz:
I just discovered that you already have been working on this and your branch already is a bit more advanced than this one, I am going to close this.

PR #1193, opened 2016-09-04
Use python3 if available

### LINKED ISSUE #1080: http-serve does not work in Arch Linux
In Arch Linux, `/usr/bin/python` points to Python 3.x and not Python 2.x. So using `http-serve` results in the following error:

```
/usr/bin/python: No module named SimpleHTTPServer
```

I think there should be a way to select which Python interpreter to use, so Arch Linux users could set it to `/usr/bin/python2`. Another user case: some distros may not install Python by default, and the user may install Python in their home directory or `/opt`.

ezintz:
A simple change to fix #1080, I hope it works for everyone.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -158,7 +158,11 @@ fi
 # Miscellaneous
 
 # Serves a directory via HTTP.
-alias http-serve='python -m SimpleHTTPServer'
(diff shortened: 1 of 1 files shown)


belak:
Thanks for the contribution!

PR #1198, opened 2016-09-21
Add files via upload

jonny4475:
just added virtual env name when active to prompt


jonny4475:
sorry if this did something niggling. i thought i was making a PR for my local fork so i could close the erroneous upload.

PR #1201, opened 2016-10-01
fix url for node-docs

ahmedelgabri:
If you do `node-doc fs` with the current implementation it won't work because `all.html#fs` doesn't exist, there is a page already called `fs.html` so this fixes this issue.

### DIFF
--- a/modules/node/functions/node-doc
+++ b/modules/node/functions/node-doc
@@ -11,4 +11,4 @@ if [[ -z "$BROWSER" ]]; then
 fi
 
 # TODO: Make the sections easier to use.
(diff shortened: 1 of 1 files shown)


facastagnini:
Hi @ahmedelgabri , your PR got merged in the community maintained fork of this repo.

ahmedelgabri:
Thanks @facastagnini, so from this I can assume that this repo won't be maintained anymore?

facastagnini:
Ahmed that is correct, this repo is not maintained. Please see #1239 to learn how we end up with a community fork.

PR #1203, opened 2016-10-03
[Homebrew] remove --all flag

pruett:
Remove flag as per homebrew's warning:

```
Warning: We decided to not change the behaviour of `brew upgrade` so
`brew upgrade --all` is equivalent to `brew upgrade` without any other
arguments (so the `--all` is a no-op and can be removed).
```

```
❯ brew --version
Homebrew 1.0.6
```

PR #1205, opened 2016-10-06
Resolves: Warning: We decided to not change the behaviour of `brew up…

lancespeelmon:
Resolves: Warning: We decided to not change the behaviour of `brew upgrade` so
`brew upgrade --all` is equivalent to `brew upgrade` without any other
arguments (so the `--all` is a no-op and can be removed).

as of:
Homebrew 1.0.6
Homebrew/homebrew-core (git revision 5b51; last commit 2016-10-06)


indiesquidge:
@lancespeelmon, this is a duplicate of #1203, which was created two days prior.
I would suggest closing it :+1:

PR #1210, opened 2016-10-14
Speeds up completion startup

samjonester:
Before this change, completion startup would rebuild ~/.zcompdump each
time a new shell was created.

After this change, completion startup will only rebuild ~/.zcompdump once per
day.

This dramatically speeds up the startup time for Prezto for all but the
first startup of the day.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -15,7 +15,15 @@ fi
 fpath=("${0:h}/external/src" $fpath)
 
 # Load and initialize the completion system ignoring insecure directories.
-autoload -Uz compinit && compinit -i
+# Will only rebuild ~/.zcompdump once per day
+autoload -Uz compinit
+if [ $(date +'%j') != $(stat -f '%Sm' -t '%j' ~/.zcompdump) ]; then
+  compinit
+else
+  compinit -C
+fi
+
+
 
 #
 # Options


sbadia:
Hi @samjonester, thanks for the PR!

On a Linux host, I've an issue with `stat`

```
/home/foobar/.zprezto/modules/completion/init.zsh:stat:20: bad file descriptor
/home/foobar/.zprezto/modules/completion/init.zsh:20: parse error: condition expected: 292
```

And running the command manually…

```
❯ stat -f '%Sm' -t '%j' ~/.zcompdump                 ⏎
stat: bad file descriptor
```

samjonester:
Ah! Thanks :) I'm on a mac, so I assume the flags are different. I've got a
couple options. Which would you prefer?

  

  1. Wrap this in a conditional to test the OS
  2. Find a better way to test whether to rebuild ~/.zcompdump
  3. Scrap the PR, as this complexity is not something you'd like to support.

  

  
On Oct 18 2016, at 10:47 am, Sebastien Badia <notifications@github.com> wrote:  

> Hi [@samjonester](https://github.com/samjonester&r=cmVwbHkrMDAwOWRhMjhjOTk1Y
WZkYTM3NDQ2OGMzNGNjN2E5NGRkZjhkZGU0OGNjNDZmMjcxOTJjZjAwMDAwMDAxMTQxZGY3ZTI5MmE
xNjljZTBhZTkyOWRhQHJlcGx5LmdpdGh1Yi5jb20=), thanks for the PR!

>

> On a Linux host, I've an issue with `stat`

>  
>  
>     /home/foobar/.zprezto/modules/completion/init.zsh:stat:20: bad file
descriptor

>     /home/foobar/.zprezto/modules/completion/init.zsh:20: parse error:
condition expected: 292

>  
>

> And running the command manually…

>  
>  
>     ❯ stat -f '%Sm' -t '%j' ~/.zcompdump                 ⏎

>     stat: bad file descriptor

>  
>

> —  
You are receiving this because you were mentioned.  
Reply to this email directly, [view it on GitHub](https://github.com/sorin-ion
escu/prezto/pull/1210#issuecomment-254530911&r=cmVwbHkrMDAwOWRhMjhjOTk1YWZkYTM
3NDQ2OGMzNGNjN2E5NGRkZjhkZGU0OGNjNDZmMjcxOTJjZjAwMDAwMDAxMTQxZGY3ZTI5MmExNjljZ
TBhZTkyOWRhQHJlcGx5LmdpdGh1Yi5jb20=), or [mute the
thread](https://github.com/notifications/unsubscribe-auth
/AAnaKKswDzlqow0ccYWiNrA-QkRtR4JOks5q1NvigaJpZM4KXBU2&r=cmVwbHkrMDAwOWRhMjhjOT
k1YWZkYTM3NDQ2OGMzNGNjN2E5NGRkZjhkZGU0OGNjNDZmMjcxOTJjZjAwMDAwMDAxMTQxZGY3ZTI5
MmExNjljZTBhZTkyOWRhQHJlcGx5LmdpdGh1Yi5jb20=).

belak:
I'm a big fan of the concept behind this PR, though it would have to work on multiple platforms before we could merge it. If anyone wants to take the time to clean this up and make it work on Linux as well as OSX, I'd be happy to merge it in.

modules/completion/init.zsh
```diff
@@ -15,7 +15,15 @@ fi
 fpath=("${0:h}/external/src" $fpath)
 
 # Load and initialize the completion system ignoring insecure directories.
-autoload -Uz compinit && compinit -i
+# Will only rebuild ~/.zcompdump once per day
+autoload -Uz compinit
+if [ $(date +'%j') != $(stat -f '%Sm' -t '%j' ~/.zcompdump) ]; then
```
malikoth: So, here are my thoughts:

First, `date +'%j'` should be replaced with `date -I`.  Not only could relying on the `%j` formatting string from [`strftime`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man3/strftime.3.html#//apple_ref/doc/man/3/strftime) produce a false negative if you came back to your terminal exactly a year after the last time you used it, but this also makes it easier to compare the output to the GNU coreutils version of stat.  See below.

I think for Mac, the second half of that comparison command should be modified to `stat -f '%Sm' -t '%F' ~/.zcompdump` in accordance with the above change.

For Linux, at least the latest Ubuntu image in Docker, the version of stat being used by default is `stat (GNU coreutils) 8.25`.  For this version of stat, the appropriate command to get the data is `stat -c %y ~/.zcompdump | cut -d ' ' -f1`

The biggest question I have is how we should tell between them.  You might first be tempted to just check whether you're on Mac or Linux with `uname -s` or something, but I have the GNU coreutils installed on my Mac via Homebrew that supercedes the BSD version of [`stat`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/stat.1.html) that comes installed by default.  If you were to run the "Mac" version of the `stat` command on my machine, it would fail due to invalid arguments.  So that's out.

The GNU coreutils version of [`stat`](http://www.gnu.org/software/coreutils/manual/html_node/stat-invocation.html#stat-invocation) comes with a `--version` flag you can pass to get the version and copyright information.  The BSD version doesn't have any kind of version flag that I can find.  So we could do a `stat --version` and if the command completes successfully, just assume it's the GNU version, and issue that version of the command.  Depending on what other versions of `stat` exist out there somewhere, we might also want / need to `grep` for the "GNU coreutils" string inside the version info to verify the version match.  But if `stat --version` returns with a non-zero error code, then we could assume it's BSD `stat`, and issue that version of the command.  Regardless of which version ends up running, the results of the command would need to be collected into a variable prior to running the comparison against the output of `date -I`.

Last of all, my `.zcompdump` file is located at `~/.config/zsh/.zcompdump` instead of `~/.zcompdump`.  So instead of hard coding that file location, it would be good to define the folder that file is in as the parent of the Prezto installation folder, wherever that happens to be.  (I know, I know, I have a really weird setup)

jeffcox: I also like the spirit of this pull and was running it my personal fork at some point.  I run all BSDs so didn't notice, but if there's a reliable command that errors out on BSD and works on GNU couldn't one simply `stat --version 2>/dev/null && foo=yes` or `stats --version 2>/dev/null; if $? = 0...`

belak:
I'm going to close this in favor of #1314. Thanks for the idea!

PR #1214, opened 2016-10-24
Integrate zsh-notify

rjcoelho:
zsh-notify is a plugin for the Z shell that posts desktop notifications when a command terminates with a non-zero exit status or when it took more than 30 seconds to complete, if the terminal application is in the background (or the terminal tab is inactive).


belak:

sorin-ionescu:
Don't prefix modules with zsh. It's redundant.

sorin-ionescu:
This is useful. I used to jerry rig my own years ago even before notifications were available.

PR #1217, opened 2016-10-27
add first pass at adding NVM to modules

jimiray:
Add an NVM module to manage Node versions and autoswitching. I have several projects using NVM and thought others could benefit from this.


belak:
It seems like this might be a better fit for an option in the node module. Would you be willing to rewrite this to fit in there?

PR #1218, opened 2016-10-27
copy bundler from oh-my-zsh

jimiray:
Add bundler support from oh-my-zsh.  I really missed the auto bundle exec addition to common gems.

### DIFF
--- a/modules/bundler/README.md
+++ b/modules/bundler/README.md
@@ -1,0 +1,50 @@
+# Bundler
+
+- adds completion for basic bundler commands
+- adds short aliases for common bundler commands
+  - `be` aliased to `bundle exec`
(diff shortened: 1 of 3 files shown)


facastagnini:
Hi @jimiray , your PR got merged in the community maintained fork of this repo.

paulmelnikow:
Here's the commit, for reference: https://github.com/zsh-users/prezto/commit/de09c9f5766964e6d967a17f85f24e13dceeaa67

beloso:
I don't think this should be part of the ruby bundle.

I for one use `bundle` but I don't use `ruby` (not directly though).

I need to execute some gems daily, but it's nothing related to ruby development.

PR #1220, opened 2016-10-28
Add deer (ranger clone for zsh) external module support

rjcoelho:


belak:

PR #1222, opened 2016-10-28
Integrate autoenv module

rjcoelho:
Autoenv automatically sources (known/whitelisted) .env and .out files.

https://github.com/horosgrisa/autoenv


facastagnini:
Hi @rjcoelho , your PR got merged in the community maintained fork of this repo.

belak:

PR #1223, opened 2016-10-28
Add alias to brew cask upgrade

rjcoelho:
See https://github.com/caskroom/homebrew-cask/issues/309


rjcoelho:
Using https://github.com/buo/homebrew-cask-upgrade instead

PR #1224, opened 2016-10-28
Add git alias to diff tags from remote and prune local tags no longer…

rjcoelho:
See http://stackoverflow.com/questions/1841341/remove-local-tags-that-are-no-longer-on-the-remote-repository

### DIFF
--- a/modules/git/alias.zsh
+++ b/modules/git/alias.zsh
@@ -168,6 +168,11 @@ alias gSs='git submodule sync'
 alias gSu='git submodule foreach git pull origin master'
 alias gSx='git-submodule-remove'
 
+# Tags (t)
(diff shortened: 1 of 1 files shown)


paulmelnikow:
The Stack Overflow license at the time that was posted was CC-BY-SA, which [it's not clear is compatible with the MIT license](http://meta.stackexchange.com/questions/271080/the-mit-license-clarity-on-using-code-on-stack-overflow-and-stack-exchange). They've since switched, which is good.

belak:
Good catch Paul.

These look useful, however it most likely cannot be merged as-is because of the license compatibility issues with gtd.

For this to be merged, the new aliases should be added to the readme and a different snippet would need to be used.

Because of this (and the conflicts) I'm going to close this for now, but please feel free to open up a new pull request if those issues get resolved. :)

PR #1226, opened 2016-10-28
Include iterm2 integration (if applicable)

rjcoelho:
See https://iterm2.com/documentation-shell-integration.html

### DIFF
--- a/modules/terminal/init.zsh
+++ b/modules/terminal/init.zsh
@@ -119,3 +119,7 @@ then
   # Sets the tab and window titles before command execution.
   add-zsh-hook preexec _terminal-set-titles-with-command
 fi
+
+# Iterm2 integration
+test -e "${HOME}/.iterm2_shell_integration.zsh" && \
+  source "${HOME}/.iterm2_shell_integration.zsh"


belak:
It would be nice if this could be moved to the osx module and only loaded if the user is in iTerm2. It would be a little weird to try to load this if the user was using plain Terminal.app.

sorin-ionescu:
This is terminal-related; so, it should be in the terminal module, but right now, it conflicts. Both the terminal module and prompt themes will have to be rewritten to support it.

belak:
If we want to merge this, I'd like to see an option to disable it as well (or default to off and have a way to turn it on)

jeffwidman:
I would like to see this merged as I run it myself and think it's helpful.

> This is terminal-related; so, it should be in the terminal module, but right now, it conflicts. Both the terminal module and prompt themes will have to be rewritten to support it.

Not sure what you mean by "it conflicts"? I run this myself with no issues using the default `sorin` prompt.

I am split on whether this should be added to the terminal module or OS X module, either one makes sense to me. 

However, regardless of which module, the readme needs to be updated to document this functionality. Note that this actually won't work unless the iterm integration script is also installed.

PR #1227, opened 2016-10-29
Add git alias for `gcM` (GPG-sign the commit)

mager:
Add the `-S` option to `gcm`

See: https://git-scm.com/docs/git-commit


facastagnini:
Hey @mager that is a git specific configuration, it is better to save those preferences in .gitconfig

mager:
👍

PR #1228, opened 2016-11-04
#780 Updated fasd README.md

Solisol:
Suggested changes from #780


belak:
Do you think you could fix the typo in here? `isntalled` should be installed.

belak:
This was merged in 95036c908c414569c59bc9b8a97108fa922ea656. Thanks for the contribution!

Issue #1055, opened 2015-12-30
coalesce:1: scalar parameter arg created globally in function coalesce`


kud:
Does it mean anything to you?

I've often got this. Mostly with git apparently.

sorin-ionescu:
Yes, I think there is a conflict, between what I assume is the Git completion, and the [`coalesce`](https://github.com/sorin-ionescu/prezto/blob/master/modules/helper/init.zsh#L25) function, which, thus far, has been used in prompt themes. How do I reproduce it?

kud:
I've got it (I think) when:

- git pull
- the pull isn't finished yet, I do git log

Or maybe 

- git pull
- the pull isn't finished yet, git checkout <tab>

sorin-ionescu:
Are you running those commands in parallel?

kud:
No. It happens when I'm too fast for the cli in fact. I already typed the next command when the previous one wasn't finished.

sorin-ionescu:
What Zsh version are you running? It has never happened to me. I cannot reproduce it.

sorin-ionescu:
It happened to me too. It seems to be a race condition.

kud:
An example.

```
~/P/_/play-tv ❯❯❯ git add .                                                                                           feat.tv-bottom ✱
gi%                                                                                                                                    ~/P/_/play-tv ❯❯❯ git commit -m "WIP"                                                                                 feat.tv-bottom ✚
git re[feat.tv-bottom a14bf5d] WIP
 3 files changed, 5 insertions(+), 1 deletion(-)
~/P/_/play-tv ❯❯❯ git recoalesce:1: scalar parameter arg created globally in function coalesce
```

kud:
`zsh 5.2 (x86_64-apple-darwin15.0.0)`

sorin-ionescu:
`coalesce` is called [here](https://github.com/sorin-ionescu/prezto/blob/master/modules/prompt/functions/prompt_sorin_setup#L131) and implemented [here](https://github.com/sorin-ionescu/prezto/blob/master/modules/helper/init.zsh#L25-L31).

sorin-ionescu:
Do you have `setopt WARN_CREATE_GLOBAL`? Execute `setopt | grep warn` to find out.

kud:
When `setopt | grep warn`, nothing returns.

sorin-ionescu:
[...]

PR #1241, opened 2016-12-03
resolve race condition between git completion and coalesce function

wonbyte:
Replace existing coalesce function with
`
function coalesce { print "${${(s: :)@}[1]}" }
`
to prevent race condition between git completion and coalesce.

Fix for issue #1055

### DIFF
--- a/modules/helper/init.zsh
+++ b/modules/helper/init.zsh
@@ -23,9 +23,7 @@ function is-true {
 
 # Prints the first non-empty string in the arguments array.
 function coalesce {
-  for arg in $argv; do
-    print "$arg"
-    return 0
-  done
-  return 1
+  print "${${(s: :)@}[1]}"
 }
+
+


facastagnini:
Hi @wonbyte , your PR got merged in the community maintained fork of this repo.

belak:
As mentioned in the above PR, it looks like this was reverted on the community fork because of this: https://github.com/zsh-users/prezto/issues/23

I'd like to see a fix for that before merging this.

belak:
I'm going to close this for now because we're not sure what a better fix would be and this has been inactive for over a week. There's more information in https://github.com/prezto-inactive-community-fork/prezto/issues/23 about a few specific cases this breaks.

sorin-ionescu:
@belak Could you label issues? Then I jump into the more important ones first.

belak:
Sure, I'll do that in the future. Still getting used to this.

sorin-ionescu:
Would switching to zsh-async fix this? It's far more sophisticated than the caveman async I'm using.

PR #1242, opened 2016-12-21
Auto load identities under macOS.

gnap:
Since macOS sierra, the Keychain won't load saved identities during startup, which causes manually input passphrase for every login.

The solution is to enable the ssh module on macOS previously disalbed，and use `ssh-add -A` instead of `ssh-add` command。


paulmelnikow:
Note: In the new fork, this is superseded by zsh-users/prezto#5 and zsh-users/prezto#7.

Issue #1239, opened 2016-11-28
READ ME BEFORE OPENING AN ISSUE: Is this an abandoned project.


facastagnini:
Hi guys,

It seems that @sorin-ionescu is not active anymore. There is a pile of useful PRs and no one with merge rights. Just take a look at the [code-frequency](https://github.com/sorin-ionescu/prezto/graphs/code-frequency)... 

When a cool project like prezto is abandoned by the repo owner and no one else has commit rights the community should find the way to orderly move forward. Is there a well maintained fork that we can move to or someone willing to do it?

CCing some contributors that I have seen active around here... @rtfpessoa @ezintz @ericbn @joaosa @sectioneight @leoj3n @toogley

PD: @sorin-ionescu are you there? Please drop a line here so we can figure out what to do next!

rtfpessoa:
I only did a fork and started merging some prs but only for private use.

To be fair, I know very few bits of the project so I do not think I would be a good maintainer alternative.

Though, I would not mind helping.

ericbn:
Also CCing everyone from https://github.com/sorin-ionescu/prezto/issues/1163: @kontrafiktion, @kress95, @ezintz, @fuadsaud, @stefanosc.

toogley:
I would also be interested in helping. (Although i have little software development experience)

Am 28. November 2016 16:03:31 MEZ, schrieb Federico Ariel Castagnini <notifications@github.com>:

toogley:
(Just for reference: I have currently no interest in this project - I Initially, answered with the opposite, because I thought i was tagged for another project i'm interested in. Sorry.)

ezintz:
Yes, sure let's see how we can make progress. As most of you probably did, I also merged some of the pull requests which made sense to me. Let me know how I can contribute and support you.

However I will need to keep a separate project as I've integrated it to my dotfiles to take my configuration always with me.

stefanosc:
Thank you for this conversation. Unfortunately, I have not been active at all in development for almost 18 months. I am still using prezto and the sorin theme despite the lack of commits / merged pull requests.
I would have already moved to zim if it wasn't for this:  https://github.com/Eriner/zim/issues/82
I don't know how helpful I could be, however if others are interested I could try. To me it makes more sense to move to zim and integrate the missing pieces.

stefanosc:
@rjcoelho this sums it up (first paragraph from readme):

[...]

PR #1257, opened 2017-02-25
Remove extra newline under numbered list

rootulp:
This is to remove the spacing above and below the `3`


Neo2SHYAlien:
You should make pull request here - https://github.com/zsh-users/prezto. For more information checkout this issue https://github.com/sorin-ionescu/prezto/issues/1239

PR #1258, opened 2017-03-03
Updated Command to change the default shell in README.md

ishanjain28:
This change only updates the command to change the default shell after installing prezto.

Currently it is 

`chsh -s /bin/zsh`

Issues with this are: 
1. On Ubuntu 14.04 and 16.04, This command will not change the default shell, Instead it'll give an error as shown in image.
2. The path where zsh is installed isn't always `/bin/zsh`, So i also changed that and made it a little bit better. 

What does this Improves: 

The suggested edit correctly changes the shell with very little input from user and also allow the user to change the default shell for a user account they are not logged in at that moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ version is 4.3.17.
 
   4. Set Zsh as your default shell:
 
-        chsh -s /bin/zsh
+        chsh -s `which zsh` [username here]
 
   5. Open a new Zsh terminal window or tab.
 


paulmelnikow:
Since development has stopped on this repo (see #1239) a community fork [zsh-users/prezto](https://github.com/zsh-users/prezto) was created in December 2016. zsh-users is an organization dedicated to host Zsh community projects and therefore became the rational home for a community maintained version of prezto.

You may want to migrate over. There are some [upgrade instructions](https://github.com/zsh-users/prezto#upgrading-from-sorin-ionescuprezto). You may also want to reopen this PR over there.

ishanjain28:
@paulmelnikow Oh! Thank you for telling me about the active fork of this project. I was looking at these PRs and wondering why they haven't been merged. 

Looks like my PR can be helpful in [zsh-users/prezto](https://github.com/zsh-users/prezto) So, I'll send this PR over there. 

Thanks again.

PR #1264, opened 2017-03-13
Make it easier to integrate with prezto by not forcing a directory name

belak:


belak:
This will let antigen remove a number of hacks they're using to load prezto and make it possible to place this in any directory and source it. The existing location will still work.

PR #1272, opened 2017-04-02
Adds additional support for PS prompt context expansion

### LINKED ISSUE #1246: syntax-highlighting module breaks '%_' in PS2 (continuation prompt)
When adding the syntax-highlighting module, the `%_` expansion (current constructs) doesn't work in PS2/PROMPT2.

Steps to reproduce:

1. Clone prezto repo, init submodules and setup symlinks
2. Add `syntax-highlighting` to modules, second to last like docs say
3. Open a new terminal
4. Try starting a for statement and continuing on next line
5. PS2 is `>` instead of `for>`

This screenshot illustrates the issue. The first command is done without syntax-highlighting and the second with it:
prezto_issue_ps2_highlighting

This was tested on all vanilla prezto, with no other module added and no change to .z* files. I use zsh 5.3.1 on Arch Linux.

johnpneumann:
Resolves #1246. 

I'm not sure that I'm on-board with changing the default functionality, however, I think this provides a better scenario given that it has to be enabled versus disabled.

### DIFF
--- a/modules/editor/README.md
+++ b/modules/editor/README.md
@@ -20,6 +20,13 @@ To enable the auto conversion of .... to ../.., add the following to
 
     zstyle ':prezto:module:editor' dot-expansion 'yes'
 
+### PS Context
+
+To enable the prompt context to be set, add the following to your
+*zpreztorc*.
+
+    zstyle ':prezto:module:editor' ps-context 'yes'
+
 Theming
 -------
 
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -103,11 +103,26 @@ function editor-info {
   fi
 
   unset REPLY
-
-  zle reset-prompt
-  zle -R
+  zle zle-reset-prompt
 }
 zle -N editor-info
+
+# Reset the prompt based on the current context and
+# the ps-context option.
+function zle-reset-prompt {
+  if zstyle -t ':prezto:module:editor' ps-context; then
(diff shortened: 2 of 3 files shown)


belak:
This seems like a strangely specific hack... do you know how other frameworks like OMZ handle this?

johnpneumann:
@belak - OMZ doesn't handle this so far as I know. This is specific to being able to have the visual notification that you're in a different mode, specifically centered around the alternate keymaps specified within the prompts, which I do not believe is something that's supported by OMZ. 

It's also how the syntax-highlighting module determines whether or not it should be highlighting anything within the PS3 prompt ([relevant line](https://github.com/zsh-users/zsh-syntax-highlighting/blob/15d4587514a3beaa13972093e335bf685b6726a9/highlighters/main/main-highlighter.zsh#L224)]. 

If there's a better way of not consuming the prompts, I'm very open to doing something different as I'm not a huge fan of the implementation, however, without it, those that do want to see the context that they're in are unable to. Example of why it may be relevant:

with_context

without_context

belak:
This looks fine to me... if anyone's curious about the CONTEXT variable, it can be found here: http://zsh.sourceforge.net/Doc/Release/Zsh-Line-Editor.html

johnpneumann:
Awesome. Thanks @belak. I found the docs in the `man zshzle` and the info on the zstyle in the `man zshmodules`.

PR #1273, opened 2017-04-04
Bump license years and tidy up readme

paulmelnikow:
Makes the readme a bit friendlier.

Re-created from zsh-users/prezto#34.


belak:
I'm not a copywrite expert, but I think it would be best to keep it identical to how it was seen in the README.

paulmelnikow:
Makes sense, I'll update it.

belak:
Thanks!

PR #1278, opened 2017-04-05
Added option to skip git aliases definition

### LINKED ISSUE #947: Forced to have git aliases

belak:
Fixes #947 

I fought with git for a while and eventually decided to just cherry pick the commit and make some fixes. This replaces #948.


belak:
This one may be a little more controversial than some of the other PRs I've merged, but I still feel like this is worthwhile, at least with the git module where there are already a ton of aliases, some of which are known to shadow existing commands. This is a way to allow people to continue to add aliases while letting people frustrated by the fact that there can be hundreds of additional aliases disable them.

PR #1283, opened 2017-04-06
Add rust module

georgemarshall:


belak:
I've love to have a full featured rust module at some point. I'd like to see a bit more before we consider adding it as a module, though. It seems a bit of a waste to add a module for just a path change. Any other ideas and additions would be welcome!

belak:
Hi! Thanks for submitting this! I recently started using rust and like the idea but it still seems like a fairly small module. We're also not really accepting new modules or prompts in the main repo, if you're still interested in getting this in, please feel free to submit it to the contrib repo (https://github.com/belak/prezto-contrib) and I'll merge it as a starting point for a full-fledged rust module.

PR #1288, opened 2017-04-10
Improve node module performance

tough-griff:
`brew --prefix` can be an expensive operation, and if `nodenv` is installed it should live in the user's $PATH anyway. These changes improve performance and parity with the `ruby` module.


belak:
I understand the `brew --prefix` change, but is there a reason `--no-rehash` was added as well?

tough-griff:
It improves start-up time since nodenv doesn't rehash on every shell init. It mirrors behavior from https://github.com/sorin-ionescu/prezto/blob/master/modules/ruby/init.zsh, as does the inclusion of the path modification.

belak:
Makes sense. Thanks for the explanation.

belak:
Merged! Thanks for your contribution!

PR #1289, opened 2017-04-10
Add support for iTerm2 v3 in tab function

tough-griff:
iTerm 2 version 3.0 and up is registered as `iTerm2` rather than `iTerm`, and uses a different AppleScript API.

This provides support for using the `osx` module's `tab` command with new versions of iTerm 2.


belak:
AppleScript is so weird

PR #1290, opened 2017-04-10
Fix path in node module

tough-griff:
The last PR got merged before I got the chance to fix this line! 😩


belak:
Thanks for catching this. Sorry I missed it.

PR #1292, opened 2017-04-11
ADDED: tab-title function in osx module

nstCactus:
Added a tab-title function in the osx module that changes the current tab title (works in iTerm and Terminal).


belak:
Thanks for the contribution, but I'm fairly certain set-tab-title (in the terminal module) already does this. It uses a slightly different escape code, but it should still do what you want.

PR #1295, opened 2017-04-12
Do not call `editor-info` in `zle-line-finish`

maximbaz:
Porting the bug fix from zsh-users/prezto.
The details of the original discussion can be found here: https://github.com/zsh-users/prezto/pull/17

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -153,8 +153,8 @@ function zle-line-finish {
     echoti rmkx
   fi
 
-  # Update editor information.
-  zle editor-info
+  # Editor info is not updated as it causes unnecessary refresh in previous prompt.
+  # See discussion here: https://github.com/zsh-users/prezto/pull/17
 }
 zle -N zle-line-finish
 


belak:
It would be nice to have a bit more explanation in the comment itself because I believe that fork will eventually go away, but I think this should be fine as-is... I'll do a bit more research and hopefully merge it soon

sorin-ionescu:
If you don't update editor info, you won't get override/vi command information in the prompt theme. It's not a bug.

sorin-ionescu:
Emacs mode has a text override mode, which can be indicated. vi has viins and vicmd.

belak:
I'm fairly certain this is a bug. There's no reason for the prompt text to update *after* a command is run... This is responsible for changing the time on a previous prompt after a command is done which isn't expected behavior.

sorin-ionescu:
@belak Wait for me to comment when you encounter code that makes no sense. I should have commented, but I thought it was self-explanatory.

sorin-ionescu:
I don't know how to update it otherwise.

belak:
Fair enough. If there's not a way to work around this maybe we should document it and move on.

sorin-ionescu:
Yes, document it. Do a `bindkey -v` to see how it works.

belak:
And usually when I ping you I plan to wait for a response unless I do my research and figure out why something exists. There aren't very many things that are so urgent that they need to be merged in right away. :)

sorin-ionescu:
I dislike the refresh as well, but I don't know how to fix this. I know that there are vi prompts out there that build a status bar, perhaps those get rid of the refresh. I haven't studied them since they looked too complicated to just get around a minor issue.

PR #1300, opened 2017-04-16
Python virtualenv auto workon cwd hook

sgtsquiggs:
This is a continuation of @jmagnusson's #910 

----

Enables automatic activation of a virtualenv when jumping into a directory. This is done by looking in the current directory for a file or directory named .venv. If it's not found in the current directory its parents will also be examined. .venv can be either:

A file containing the name of a virtualenv found in $WORKON_HOME
A directory containing bin/activate (meaning that the directory is assumed to be a virtualenv.)
If $WORKON_HOME is set it is assumed that virtualenvwrapper is installed and the workon command will be issued.


malikoth:
Oh, I like this!  Does it work when navigating with fasd?

belak:
This looks like a port of the oh-my-zsh plugin (which is fine) but I believe there are a few known issues with it (See: https://github.com/robbyrussell/oh-my-zsh/pull/5817 for one of the fixes) and I'd like to avoid them if possiblem

@malikoth I don't see any reason it wouldn't... any time the pwd changes, the hook should call so things like autojump, z and fasd should have no trouble (I used the OMZ version with z for a while).

modules/python/init.zsh
```diff
@@ -34,6 +34,51 @@ if (( ! $+commands[python] && ! $+commands[pyenv] )); then
   return 1
 fi
 
+function _python-workon-cwd {
+    if [[ -z "$WORKON_CWD" ]]; then
```
belak: What is this check used for?

sgtsquiggs: I believe it was originally there to prevent it running twice before finishing. I don't see how it is useful as a cwd hook call. Going to remove it.

sgtsquiggs: https://github.com/jayusa123/oh-my-zsh/commit/69116fa806f5a7405d33cc1f7c66d4c26ad4253b
> Avoid infinite `cd` loops under certain conditions.

Not sure if this is still relevant

modules/python/init.zsh
```diff
@@ -65,6 +110,12 @@ if (( $+commands[pip] )); then
   source "$cache_file"
 
   unset cache_file
+
+  # Load auto workon cwd hook
+  if zstyle -t ':prezto:module:python' enable-virtualenv-auto-workon-cwd 'yes'; then
```
belak: Opinions on the name? Something shorter would be nice. We've used stuff like `:prezto:module:tmux:auto-start` in other places... how would `:prezto:module:python:auto-cwd-virtualenv` be? Suggestions would be welcome because I'm terrible at naming things.

modules/python/init.zsh
belak: Not that I'm aware of... the previous version worked because it wasn't a local variable and because there were other blocks other than this one.

I'd say remove it, and we can add it back later if needed... because this isn't enabled by default and it's a new feature, I'm fine with fixing this later (if it comes up at all)

modules/python/init.zsh
belak: Or cwd-auto-virtualenv?

malikoth: I've been using [autoenv](https://github.com/kennethreitz/autoenv) to accomplish roughly the same thing in the past, and that repo has roughly half the stars that Prezto itself has, so it's not a tiny or unknown project.  So barring other reservations, I'd go for ever shorter like, just `:prezto:module:python autoenv`.  My only problem with that one is that almost sounds like it's activating Reitz' `autoenv` tool.  Is that a valid concern?

Personally, I at least don't feel like the `cwd` really needs to be part of the name.  If you didn't want `autoenv`, my next vote would be for `auto-virtualenv`.  iunno, just my $0.02.

belak: I'd prefer `auto-virtualenv` to avoid confusion. I think that makes sense.

diraol: or `auto-venv` would also work, doesn't it?

belak: Hm, what about `zstyle ':prezto:module:python:virtualenv' auto-switch 'yes'`?

It lines up with the ruby module and conveys the same meaning.

Once we decide on a final setting name, it should be added to the sample zpreztorc. Other than that, I think this looks ready.

malikoth: `auto-switch 'yes'` sounds fine to me.

belak:
Awesome! Let's get that setting in the zpreztorc and I'll get this merged in! Thanks for your work on this @sgtsquiggs !

modules/python/init.zsh
```diff
@@ -34,6 +34,48 @@ if (( ! $+commands[python] && ! $+commands[pyenv] )); then
...
+  # Check for virtualenv name override
+  local ENV_NAME=""
+  if [[ -f "$PROJECT_ROOT/.venv" ]]; then
+    ENV_NAME="$(cat "$PROJECT_ROOT/.venv")"
+  elif [[ -f "$PROJECT_ROOT/.venv/bin/activate" ]];then
+    ENV_NAME="$PROJECT_ROOT/.venv"
+  elif [[ "$PROJECT_ROOT" != "." ]]; then
+    ENV_NAME="${PROJECT_ROOT:t}"
+  fi
+  if [[ -n $CD_VIRTUAL_ENV && -n $VIRTUAL_ENV ]]; then
```
sgtsquiggs: This block here fires every `cwd` within a project, causing you to `deactivate`/`activate` on every `cwd`. Not the desired outcome.

sgtsquiggs: Perhaps this would be better as
```
if [[ -n $CD_VIRTUAL_ENV && "$ENV_NAME" != "$CD_VIRTUAL_ENV" ]]; then
```

belak: That would make sense... good catch.

It's worth noting that we might not need to call "deactivate" then "workon" if they don't match but your fix should still work.

belak:
Sorry I didn't get around to this until now. Thanks for all the fixes! Looks good to me!

belak:
I added a little bit of info to the python readme as well. Thanks again!

PR #1149, opened 2016-05-25
Add some new git aliases, modify a few more.

malikoth:
# Additions
- Add `gbV` command to show more verbose git branch info.
- Add `gcam` to make it possible to execute `gca; gcm '<your message
  here>'` more simply.
- Add `gii` command to temporarily untrack (ignore) a file.
- Add `giI` command to uningore a file.
# Modifications
- Change alias of `gbl` to `gbv`. Personally, I think aliases that
  include a switch in the command should include the switch in the alias
  if possible.  This makes them easier to remember.
- Change alias of `gbL` to `gba`. I think the `-a` switch is more
  salient to what this alias does than the `-v`.  Furthermore, with this
  PR there are already `gbv` and `gbV` aliases, so those are out.
- Change implementation of `gCl` alias to use built-in capabilities of
  git, rather than sed.

Please, let me know what you like, and what I should have left well-enough alone.  :D  I'm not married to any of the changes here.  What I started this PR for was to include the `gii` and `giI` commands.  I made a few other changes I'd like to see, that I think make git aliases more consistent, but I would happily back them out to get my two commands in.

PR #1301, opened 2017-04-17
Add some new git aliases, modify a few more.

malikoth:
# Additions
* Add `gbV` command to show more verbose git branch info.
* Add `gcam` to make it possible to execute `gca; gcm '<your message
  here>'` more simply.
* Add `gii` command to temporarily untrack (ignore) a file.
* Add `giI` command to uningore a file.

# Modifications
* Change alias of `gbl` to `gbv`. Personally, I think aliases that
  include a switch in the command should include the switch in the alias
  if possible.  This makes them easier to remember.
* Change alias of `gbL` to `gba`. I think the `-a` switch is more
  salient to what this alias does than the `-v`.  Furthermore, with this
  PR there are already `gbv` and `gbV` aliases, so those are out.
* Rename `gbx` and `gbX` commands to `gbd` and `gbD` respectively
* Change implementation of `gCl` alias to use built-in capabilities of
  git, rather than sed.

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -40,17 +40,17 @@ Aliases are enabled by default. You can disable them with:
 ### Branch
 
   - `gb` lists, creates, renames, and deletes branches.
-  - `gba` lists all local and remote branches and their commits.
   - `gbc` creates a new branch.
-  - `gbd` deletes a branch.
-  - `gbD` deletes a branch irrespective of its merged status.
+  - `gbl` lists branches and their commits. (also `gbv`)
+  - `gbL` lists all local and remote branches and their commits.
   - `gbm` renames a branch.
   - `gbM` renames a branch even if the new branch name already exists.
   - `gbs` lists branches and their commits with ancestry graphs.
   - `gbS` lists local and remote branches and their commits with ancestry
     graphs.
-  - `gbv` lists branches with verbose information about their commits.
   - `gbV` lists branches with more verbose information about their commits.
+  - `gbx` deletes a branch. (also `gbd`)
+  - `gbX` deletes a branch irrespective of its merged status. (also `gbD`)
 
 
 ### Commit
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -44,13 +44,13 @@ Aliases are enabled by default. You can disable them with:
   - `gbc` creates a new branch.
   - `gbd` deletes a branch.
   - `gbD` deletes a branch irrespective of its merged status.
+  - `gbm` renames a branch.
+  - `gbM` renames a branch even if the new branch name already exists.
   - `gbs` lists branches and their commits with ancestry graphs.
   - `gbS` lists local and remote branches and their commits with ancestry
     graphs.
   - `gbv` lists branches with verbose information about their commits.
   - `gbV` lists branches with more verbose information about their commits.
-  - `gbm` renames a branch.
-  - `gbM` renames a branch even if the new branch name already exists.
 
 
 ### Commit
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -40,14 +40,15 @@ Aliases are enabled by default. You can disable them with:
 ### Branch
 
   - `gb` lists, creates, renames, and deletes branches.
+  - `gba` lists all local and remote branches and their commits.
   - `gbc` creates a new branch.
-  - `gbl` lists branches and their commits.
-  - `gbL` lists local and remote branches and their commits.
+  - `gbd` deletes a branch.
+  - `gbD` deletes a branch irrespective of its merged status.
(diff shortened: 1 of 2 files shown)


malikoth:
This PR supercedes #1149 which I am unable to rebase because I deleted my fork from which that PR originated.

belak:
I'm not sure what to do about some of these, but here are my thoughts. Please feel free to refute them or offer different opinions:

1. gbx and gbX appear to have been named that way to match all the other aliases. The x seems to be used for delete. As such, I'm not sure it makes sense to rename them to gbd and gbD because they wouldn't match the rest of the aliases.

2. gbl and gbL were meant to stand for branch list. Changing that to a v obscures the meaning a bit (even if it does match the command line switches). I do get where you're coming from though... I'm not sure what the final decision on this will be especially because there isn't a logical alias to put git branch -vv in if we stay with the "list" mnemonic.

I also dislike changing aliases after the fact, purely because there are most likely people using the existing aliases... I am in favor of changing them when they don't make a lot of sense, but I'm not sure that's quite the case here. There was at least some form of logic behind them.

Because I'm not sure what the final decisions will be, it might take me a little while to nail down what would be best. Having other people chime could help move the discussion along and make the decisions easier, especially if there's agreement.

malikoth:
I wanted to propose these changes that I thought were logical and see what people thought of them.  Now that some time has passed to think about it more, I think I agree with you actually, that I shouldn't try to move gbx to gbd.  Both because it breaks uniformity (arbitrarily) and because people are already using it.  No doubt, if you merged this as is, THEN you would get people showing up to comment and ask why we removed the shortcuts they were using.

The changes to gbl were inspired by the fact that I just wasn't sure what alias to assign to `git branch -vv`.  I don't agree with you about changing the `l` to `v` obscuring the meaning.  After all, it's doing the same thing as `gb`, which is just displaying branches.  They just show different levels of information.  One thing I saw in Akarzim's docker aliases was that one command, `docker ps` had multiple aliases, `dkps` and `dkls` to appeal to whichever mnemonic is more memorable to you.  I wonder if it might be similarly OK to keep my aliases here for `gba`, `gbv`, and `gbV`, while restoring `gbl` and `gbL` to their original values.  What do you think?

malikoth:
Also, I should change all the single-dash single-character arguments to the double-dash full-word arguments used in the rest of the module.

diraol:
+1 for `gba`!

why not, instead of `gbV`, `gbvv`?

Another one I would like to see added here is `gst` (for `git status`). Despite the fact that `gs` is being used for `stash`.
I'm a `oh-my-zsh` user, and over there `gst` states for `git status`, and it seems reasonable to me. =)

Ow, and maybe giap (git add -p) would go fine too.

About the "multiple aliases" for one command, I do not see any problem with that.

malikoth:
@diraol I came from OMZ as well, and I made the same transition from `gst` to `gws`.  With so many git aliases, I actually really like the way Prezto has them cordoned off into logical groups, each of which then gets an additional letter tacked on to identify its group.  This has made it much easier for me to learn and remember far more aliases than I did when using OMZ, because their aliases were kind of just arbitrary choices.  It's a bit of a transition, but I really felt much more solid with my use of aliases after making the jump myself.

For what it's worth, use the new `sa` alias (for "search aliases") from #1150 to help you learn all the new Prezto aliases.  =)

> why not, instead of gbV, gbvv?

I dunno...  I like it better.  It kind of goes with existing conventions of having slightly different variations of commands just use the capitol of the last letter in the alias.  Any strong opinions from elsewhere?

diraol:
@malikoth Thanks for the answers =)
They helped me understand a little better the logic behind the chosen aliases.

Wow! `sa` is just wonderful! Thanks! Life changing!

malikoth:
> Wow! `sa` is just wonderful! Thanks! Life changing!

RIGHT?!?  That's what I thought too!

malikoth:
@belak Thoughts on the current state of this one?  I'll remove whatever you think I should to get this merged.

belak:
Sorry, I've been in the middle of a move so I'm a bit behind in reviews. If I don't get to it this weekend, feel free to ping me again.

belak:
I'm pretty sure this will be my last comment before this gets merged. I think this is pretty much ready, but for `gbm` and `gbM` it would be nice to also have `gbr` and `gbR` (or replace it with `gbr`) because the action is a rename, even if the command line flag is `--move` ("rename" is even used in the readme). Once one of those two is done, I'll merge this.

Also, sorry about the long delay. It's been a busy few weeks for me.

malikoth:
Understandable.  :)  I really appreciate all you're doing for Prezto.  Hope life returns to a sustainable pace for you.  

`gbr` and `gbR` are in, go ahead and merge.

belak:
Merged! Thanks for doing this!

PR #1303, opened 2017-04-18
Add perl-info function & plenv subcommand aliases

delphinus:
This PR is copied from https://github.com/prezto-inactive-community-fork/prezto/pull/10 that has the original in closed #1130. I rebased and PR-ed again.

----

## perl-info function

I implemented `perl-info` with referring to `ruby-info` in the `ruby` module. The usage is almost the same with `ruby-info`. This function load info from `perlbrew` or `plenv` or pure `perl` executable.

``` sh
ruby-info
RPROMPT+=' %F{5}${ruby_info[version]}%f'
perl-info
RPROMPT+=' %F{6}${perl_info[version]}%f'

zstyle ':prezto:module:ruby:info:version' format 'version: %v'
zstyle ':prezto:module:perl:info:version' format 'version: %v'
```

## plenv subcommand aliases

Like Pelbrew, with [plenv](https://github.com/tokuhirom/plenv) you can manage plural binary installations of Perl. plenv has almost the same interfaces with rbenv.

## Load Perlbrew or plenv (suggestions from comments in https://github.com/sorin-ionescu/prezto/pull/1130)

Like the `ruby` module, Perlbrew or plenv will be loaded if that exists.


belak:
What do you think would be the best way to handle it if both perlbrew and plenv are installed? How should it be determined which should be loaded... also does it make sense to only set up 1 set of aliases if both exist?

delphinus:
I think Perlbew & plenv should be used exclusively and almost people do so. But  if there are any people who uses both, theay should switch them with some steps... such as, setting ENV before zsh launching.

I can add such a feature with the prezto way (zstyle), but I think it is too much. For example, the Ruby module has rvm & rbenv features, but also it can be used exclusively (first, searches rvm, then rbenv).

modules/perl/init.zsh
```diff
@@ -11,6 +11,29 @@ if (( ! $+commands[perl] )); then
 fi
 
 #
+# Load Perlbrew or plenv
+#
+
+# Load Perlbrew into the shell session.
+if [[ -s "${HOME}/perl5/perlbrew/etc/bashrc" ]]; then
```
belak: Could this use PERLBREW_ROOT and default to "${HOME}/perl5"?

delphinus: Fixed! and used `$HOME/perl5/perlbrew` as default value. `/perlbrew` is needed.

belak:
Merged! Thanks for your contribution. :)

Issue #1267, opened 2017-03-21
Injection Exploit in git-info


mattmolo:
`git-info` seems to not do proper escaping when retrieving the branch name. For example, setting the branch to `./an_executable` will run the local executable. An example repo can be found [here](https://github.com/njhartwell/pw3nage).

Edit by belak: There are still a few prompts which have problems. The below listed prompts have known problems:

- [x] sorin (fixed in #1361 but it would be better to switch to async so that tempfile isn't needed)
- [ ] agnoster (fix available in https://github.com/agnoster/agnoster-zsh-theme/pull/33 but there hasn't been any activity for over 4 months, so it may be worth pulling in agnoster or switching to a fork)

belak:
This is a prompt specific issue. Which one are you seeing it with? It is a known issue with the agnoster and pure prompt themes. Pull requests have already been opened against both themes (I think the one against pure has already been merged) and when that's finished the submodules will need to be updated in this repo (or you can do it manually for now).

ericbn:
Fixed in [Zim's git-info](https://github.com/Eriner/zim/pull/159) as a non-prompt specific issue.

belak:
@ericbn Even if it's fixed there, there are still prompts which are vulnerable. As an example, using `print -P` will perform prompt expansion and if you're not careful, you can end up dumping in the values there to trigger this. Similarly, the prompt_subst option can cause similar issues.

See the pure and agnoster prompts for examples: https://github.com/agnoster/agnoster-zsh-theme/pull/33 , https://github.com/sindresorhus/pure/pull/305

belak:
Hm, when I was investigating this for powerlevel9k, I also ran across it in the sorin prompt... so maybe this isn't completely an upstream issue.

belak:
I've updated the main comment with the prompts which are known to be vulnerable. Please let me know if any others are found.

belak:
Does anyone have opinions about what to do with the agnoster prompt? I realize it's relatively popular so I don't want to just remove it, but I also don't like leaving something in here with a known security issue. Should we just vendor it? It doesn't look like there are many updates... and it would make it easier to make our own changes. Or maybe just maintain a fork?

samcv:
Just made a PR to fix this for sorin prompt https://github.com/sorin-ionescu/prezto/pull/1361

belak:
The sorin prompt has been changed to using `zsh-async` and the agnoster theme has been updated to include my security fixes, so I'm closing this.

PR #1306, opened 2017-04-19
Fix sorin prompt double expansion

belak:
Refs #1267

This is fairly easy to reproduce with the https://github.com/njhartwell/pw3nage repo.

That being said, I think how this prompt actually goes about handling the git info is probably ripe for other injection issues. A more complete fix would probably be good at some point.

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -117,7 +117,7 @@ function prompt_sorin_setup {
   zstyle ':prezto:module:git:info:unmerged' format ' %%B%F{3}═%f%%b'
   zstyle ':prezto:module:git:info:untracked' format ' %%B%F{7}◼%f%%b'
   zstyle ':prezto:module:git:info:keys' format \
-    'status' '$(coalesce "%b" "%p" "%c")%s%A%B%S%a%d%m%r%U%u'
+    'status' "\$(coalesce '%b' '%p' '%c')%s%A%B%S%a%d%m%r%U%u"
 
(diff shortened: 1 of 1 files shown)


samcv:
Just noticed this vulnerability today in the tickets. Tested the the patch and it works. Please merge this ASAP. Further changes can be done later but this is quite important as it's the default prompt.

belak:
The main problem with this patch (and why I didn't just blindly merge it) is that as far as I can tell it just shifts the issue somewhere else. It would still be fairly easy to craft a branch name with single quotes which worked around this "fix". Really, the way this cache file is done isn't very good and that's a part of the problem.

samcv:
Maybe I'm not understanding in which function/file it's actually getting expanded? If I make a weird branch name and just set:
```
  zstyle ':prezto:module:git:info:keys' 'format' 'status' '%b %p %c %s%A%B%S%a%d%m%r%U%u'
```
Then it shows the error is in `prompt_sorin_git_info:14: command not found: echo_hi` (for branch `$(echo_hi)`.

This line shows: [`zle && zle reset-prompt`](https://github.com/sorin-ionescu/prezto/blob/60c433d4ce680ddb2ca899b460e3714728cb221c/modules/prompt/functions/prompt_sorin_setup#L49).

belak:
It's in a cache file. The `source "$_prompt_sorin_precmd_async_data"` line loads it by sourcing it.

samcv:
Oh no… :-(

belak:
Closed in favor of #1361

PR #1307, opened 2017-04-19
Add support for 'pip', 'pip2', 'pip3' completion

indrajitr:
Since `pip completion --zsh` might be slow, we cache it beforehand.
The implementation is along the lines of 'npm' completion in 'node' module.

## Proposed Changes
  - `python` module doesn't support completion for `pip` yet
  - this adds support for `pip` (`pip2`, `pip3`) completion
  - the output is cached for speed

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
   source "$commands[virtualenvwrapper.sh]"
 fi
 
+# Load PIP completion.
+if (( $+commands[pip] )); then
+  cache_file="${0:h}/cache.zsh"
+
+  if [[ "$commands[pip]" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
+    pip completion --zsh | sed -e "s|compctl -K [-_[:alnum:]]* pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
+  fi
+
+  source "$cache_file"
+
+  unset cache_file
+fi
+
 #
 # Aliases
 #


modules/python/init.zsh
```diff
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
   source "$commands[virtualenvwrapper.sh]"
 fi
 
+# Load PIP completion.
+if (( $+commands[pip] )); then
+  cache_file="${0:h}/cache.zsh"
```
belak: This should probably use $TMPDIR (though that might be set in a different module... I can't remember, so using mktemp might be best) and the name should probably be more specific.

modules/python/init.zsh
```diff
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
...
+
+  if [[ "$commands[pip]" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
+    pip completion --zsh | sed -e "s|compctl -K _pip_completion pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
```
belak: I assume this may cause oddities for pip2 vs pip3 if the user has both installed... because the pip command can only point to one of them... I'm not sure what the correct method of handling this would be though.

modules/python/init.zsh
indrajitr: Good point, updating to use `$(mktemp -d)`. We'll have to take care of this in other places as well.

modules/python/init.zsh
indrajitr: 1. `pip completion --zsh` just generated the stub function `function _pip_completion {}`. And it is identical regardless of `pip` version. (Try comparing `pip2 completion --zsh` and `pip3 completion --zsh`).

2. While both `pip2` and `pip3` might be available/installed, almost certainly one of them would be soft-linked to `pip`. And that's all that we need to generate and source in.

3. All that we are doing here is:
    - Sourcing in `_pip_completion pip()`
    - Tweaking `compctl -K _pip_completion pip` to work for `pip2` and `pip3` as well.

modules/python/init.zsh
```diff
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
   source "$commands[virtualenvwrapper.sh]"
 fi
 
+# Load PIP completion.
+if (( $+commands[pip] )); then
+  cache_file="$(mktemp -d)/pip-cache.zsh"
+
+  if [[ "$commands[pip]" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
+    pip completion --zsh | sed -e "s|compctl -K _pip_completion pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
```
belak: I'd personally prefer to not do a sed replacement here and just use compctl -K _pip_completion pip2 pip3 after sourcing the cache file... I assume that would still work.

indrajitr: Hmm, doing that would _leak_ private function `_pip_completion` out in the call-site. Would you consider that reasonable?

indrajitr: Updated the `sed` replacement to honor any function in case it helps to avoid leaking the function name `_pip_completion` completely.

modules/python/init.zsh
```diff
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
   source "$commands[virtualenvwrapper.sh]"
 fi
 
+# Load PIP completion.
+if (( $+commands[pip] )); then
+  cache_file="$(mktemp -d)/pip-cache.zsh"
+
+  if [[ "$commands[pip]" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
+    pip completion --zsh | sed -e "s|compctl -K [-_[:alnum:]]* pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
```
indrajitr: Updated the sed replacement to honor any function in case it helps to avoid leaking the function name _pip_completion completely.

indrajitr: Updated the sed replacement to honor any function in case it helps to avoid leaking the function name _pip_completion completely.

modules/python/init.zsh
```diff
@@ -46,6 +46,20 @@ if (( $? && $+commands[virtualenvwrapper.sh] )); then
   source "$commands[virtualenvwrapper.sh]"
 fi
 
+# Load PIP completion.
+if (( $+commands[pip] )); then
+  cache_file="${0:h}/cache.zsh"
+
+  if [[ "$commands[pip]" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
+    pip completion --zsh | sed -e "s|compctl -K [-_[:alnum:]]* pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
```
indrajitr: Updated the sed replacement to honor any function in case it helps to avoid leaking the function name `_pip_completion` completely.

belak:
Sorry about the delay. I'm in the middle of a move and starting a new job so I might not have the quickest response time.

This looks good. I'll try to test it soon and get this merged.

modules/python/init.zsh
belak: Fair enough. Thanks for the explanation. It makes quite a bit of sense.

indrajitr:
Thanks @sorin-ionescu and @belak for the review. Went ahead and did rebase/merge.

PR #1174, opened 2016-07-29
add sorin config variable to allow full PWD to be printed

vanrijn:
I have multiple source code paths for work that I need to be able to see at a glance, otherwise I have to keep typing "pwd" manually to see what directory I'm in. For example, I'll have 6 iTerm tabs open at once with these directories open:

/Volumes/SourceCode/src/git/project1-2014/foo/bar/baz/wibble/
/Volumes/SourceCode/src/git/project1-2014-rel/foo/bar/baz/wibble/
/Volumes/SourceCode/src/git/project1-2015/foo/bar/baz/wibble/
/Volumes/SourceCode/src/git/project1-2015-rel/foo/bar/baz/wibble/

And if all I can see is "/V/S/s/g/p/f/b/b/wibble" in my pwd for sorin, I can't tell what branch I'm in without typing pwd manually.

I've added a config variable to allow the full pwd to be printed instead of the shortened version. Totally new to prezto, so definitely open to comments/suggestions/changes.

Thanks!

PR #1310, opened 2017-04-26
Add zstyle option to show pwd in long format

malikoth:
Alternative implementation to #1174.

### DIFF
--- a/modules/prompt/functions/promptpwd
+++ b/modules/prompt/functions/promptpwd
@@ -9,6 +9,8 @@ local ret_directory
 if [[ "$current_pwd" == (#m)[/~] ]]; then
   ret_directory="$MATCH"
   unset MATCH
+elif zstyle -t ':prezto:module:prompt' long-pwd; then
+  ret_directory=${current_pwd}
 else
   ret_directory="${${${${(@j:/:M)${(@s:/:)current_pwd}##.#?}:h}%/}//\%/%%}/${${current_pwd:t}//\%/%%}"
 fi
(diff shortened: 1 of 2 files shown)


belak:
Anyway, sorry about the bikeshedding but I'd like to avoid changing setting names in the future if possible.

`long-pwd` (current), `full-pwd`, and other options may be preferable.

What are everyone's opinions on those?

malikoth:
Certainly, getting a good variable name to avoid arbitrary changes to it later can only be a good thing, so I support spending some time seeing what people like.

Personally, I don't care what it's called.  I just want the option to be there!

It's worth noting that I could also use `zstyle -T` instead of `zstyle -t` to test for a variable named something like `short-pwd` as the condition for the line that shortens the pwd.  This would preserve the short form as the default if no zstyle is set, so that the default behavior does not change.

indrajitr:
> Anyway, sorry about the bikeshedding but I'd like to avoid changing setting names in the future if possible.

Agreed totally!

That said, `full-pwd` might imply expanding `$HOME` to `/home/username` instead of just `~`. In that sense, `long-pwd` sounds marginally better in this particular case :)

belak:
Alright, this seems good then. Thanks for your contribution!

malikoth:
So @belak, @indrajitr brings up a good point...  Sorry to do this right after merge, but ...

Should `long-pwd` be replaced with a variable like `pwd-length` that can contain one of `short`, `long`, or `full`?  It might be nice to make it possible to expand the `$HOME` part of the pwd...

belak:
That would be useful and would be easier to expand in the future.

Issue #436, opened 2013-05-22
Broken link in README.md for rsync module


chauncey-garrett:
[Bombich's rsync](http://www.bombich.com/rsync.html)

FIXME: broken link...as of 2013-05-22 @ 5:00 PM CST

wadkar:
FWIW, here's the current link to the [diff file showing Bombich's modification to rsync-3.0.6](https://bombich.com/software/opensource/rsync_3.0.6-bombich_20150427.diff) . Does anyone know if those patches were accepted upstream?
Also, you can search for rsync in their help desk search pages and reach at following url which contains link to rsync credits and patches:
https://bombich.com/kb/ccc4/credits#rsync

indrajitr:
@wadkar, they haven't been yet. If you install `rsync` via homebrew, you'll have the patched binary though.

PR #1311, opened 2017-04-26
rsync: Update reference to Bombich's rsync page

indrajitr:
Seems like Bombich's rsync page moved with the previous link doing 404.
Adjusted to fix this.

FWIW, a similar change was done in the past (#436). But the link moved to their new KB.


indrajitr:
Since this is a trivial one, applying directly.

PR #1312, opened 2017-04-27
Convert long-pwd to pwd-length

malikoth:
Extend #1310 to specify a length for `pwd` output, rather than just have a boolean for whether than output should be short or long form.

### DIFF
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -97,8 +97,8 @@ zstyle ':prezto:module:editor' key-bindings 'emacs'
 # Auto set to 'off' on dumb terminals.
 zstyle ':prezto:module:prompt' theme 'sorin'
 
-# Display pwd in long format for themes that use promptpwd function
-# zstyle ':prezto:module:prompt' long-pwd 'yes'
+# Set how themes that use promptpwd function display the pwd, can be 'short', 'long', or 'full'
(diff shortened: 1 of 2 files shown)


belak:
Thanks for doing this! Looks good.

runcoms/zpreztorc
```diff
@@ -97,8 +97,8 @@ zstyle ':prezto:module:editor' key-bindings 'emacs'
 # Auto set to 'off' on dumb terminals.
 zstyle ':prezto:module:prompt' theme 'sorin'
 
-# Display pwd in long format for themes that use promptpwd function
-# zstyle ':prezto:module:prompt' long-pwd 'yes'
+# Set how themes that use promptpwd function display the pwd, can be 'short', 'long', or 'full'
+# zstyle ':prezto:module:prompt' pwd-length 'short'
```
indrajitr: Hmm, `pwd-length 'short'` isn't handled in the code at all it seems :)

belak: I think this was meant as the default.

malikoth: Exactly so, belak.  Much as I prefer `long` over `short`, my highest priority with this change was to make it not change things for people who didn't want them changed.  While putting the code for the `short` version in the `else` statement as I did doesn't really explicitly refer to it as the `short` case, that's what it is, along with being the default case.

PR #1314, opened 2017-05-02
Cache the completion for a quicker startup

belak:
The original idea was by @samjonester in #1210. This version aims to avoid
relying on the stat or date programs because they have different flags depending
on the OS and if it uses BSD or GNU coreutils.

This relies on extendedglob which allows us to list all files matching the "glob" .zcompdump which have been modified in the last 20 hours. If there are any, we load the cached completion, rather than simply initializing completion.

While this doesn't fix the issue with insecure directories being ignored (I specifically left that check in for now), it should ensure caching is used which will lessen the impact of that check.

CC @malikoth, @jeffcox, @Eriner

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -14,9 +14,6 @@ fi
 # Add zsh-completions to $fpath.
 fpath=("${0:h}/external/src" $fpath)
 
-# Load and initialize the completion system ignoring insecure directories.
-autoload -Uz compinit && compinit -i
-
 #
 # Options
 #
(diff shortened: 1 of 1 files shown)


malikoth:
Neat.  Looks OK to me.

breerly:
What kind of speedup are you seeing?

belak:
In my (very unscientific) tests, I was a cold start time of 1.8 seconds, a warm start time of ~0.5 seconds, and a cached start time of ~0.01 seconds.

belak:
I've been using this with a lower cache time (to force invalidations) for most of the day and it's been working fine, so I'm going to merge this. If you see any problems, feel free to ping me.

modules/completion/init.zsh
```diff
@@ -27,9 +24,21 @@ setopt PATH_DIRS           # Perform path search even on command names with slas
...
+setopt EXTENDED_GLOB       # Needed for file modification glob modifiers with compinit
 unsetopt MENU_COMPLETE     # Do not autoselect the first completion entry.
 unsetopt FLOW_CONTROL      # Disable start/stop characters in shell editor.
 
+# Load and initialize the completion system ignoring insecure directories with a
+# cache time of 20 hours, so it should almost always regenerate the first time a
+# shell is opened each day.
+autoload -Uz compinit
+compfiles=(${ZDOTDIR:-$HOME}/.zcompdump(Nm-20))
+if [[ $#compfiles > 0 ]]; then
```
sorin-ionescu: Use `(( ... ))` for math expressions.

belak: Cool. Thanks for the tip.

indrajitr: #1369 does the needful.

PR #1317, opened 2017-05-04
git stash show with untracked files

### LINKED ISSUE #1211: gsd: show untracked files in git stashes
It's easy to lose untracked files in some stash since they are never listed anywhere.
It would be nice if the `gsd` alias (or e.g. `gsD`) also included untracked files.

vogler:
Fixes #1211 

## Proposed Changes

Add git alias `gsD`, which also shows staged untracked files.


belak:
This just seems like a whole lot of magic. How does this actually work?

samcv:
Looks like the answer is in this stackoverflow answer: https://stackoverflow.com/a/22819771

That hash is the 'magic' hash for an empty tree that every git repository has.

PR #1328, opened 2017-06-08
Fix the pacu alias in the pacman module.

luismilanooliveira:
Since the deprecation of abs [1], the call for the `pacu` alias was resulting in a rsync error.
This PR just updates the alias with the equivalent command, `asp update`.

[1] https://www.archlinux.org/news/deprecation-of-abs/


belak:
Thanks for the contribution! Looks good!

PR #1330, opened 2017-06-13
Use unique tmpdir

### LINKED ISSUE #1319: Setting TMPDIR using mktemp disrupts tmux use
This commit: https://github.com/sorin-ionescu/prezto/commit/3a6ae0fc6c73963e5f6b0b77f32d219677c840b5 disrupts tmux use.  Because each login session gets a unique `$TMPDIR` when you logout and login again tmux can't find it's socket file as tmux uses `$TMPDIR` if it's set.

My solution for now is to set `$TMPDIR` in `.zshrc` but I'm wondering if there is a better fix for this?

fd0:
Fixes #1319

## Proposed Changes

  - Use a non-random `TMPDIR` which is unique for each user
  - Use a random tempfile for each shell for async theme data

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -87,7 +87,7 @@ function prompt_sorin_setup {
   unsetopt XTRACE KSH_ARRAYS
   prompt_opts=(cr percent subst)
   _prompt_sorin_precmd_async_pid=0
-  _prompt_sorin_precmd_async_data="${TMPPREFIX}-prompt_sorin_data"
+  _prompt_sorin_precmd_async_data=$(mktemp "${TMPDIR}/sorin-prompt-async-XXXXXXXXXX")
 
   # Load required functions.
   autoload -Uz add-zsh-hook
--- a/runcoms/zprofile
+++ b/runcoms/zprofile
@@ -74,5 +74,3 @@ if [[ ! -d "$TMPDIR" ]]; then
   mkdir "$TMPDIR"
   chmod 700 "$TMPDIR"
 fi
-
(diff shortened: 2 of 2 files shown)


runcoms/zprofile
```diff
@@ -65,9 +65,12 @@ fi
 #
 # Temporary Files
 #
+#
+if [[ -z "$TMPDIR" ]]; then
+  export TMPDIR="${TMPDIR:-/tmp/}/zsh-$UID"
```
belak: Because we checked that TMPDIR was empty, we can just use /tmp

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -87,7 +87,7 @@ function prompt_sorin_setup {
   unsetopt XTRACE KSH_ARRAYS
   prompt_opts=(cr percent subst)
   _prompt_sorin_precmd_async_pid=0
-  _prompt_sorin_precmd_async_data="${TMPPREFIX}-prompt_sorin_data"
+  _prompt_sorin_precmd_async_data="$(mktemp -p '' sorin-prompt-async-XXXXXXXXXX)"
```
belak: I'm not sure this works on both OSX and Linux as their mktemp utilities are very different.

runcoms/zprofile
fd0: Okay

modules/prompt/functions/prompt_sorin_setup
fd0: I'll use a call to `mktemp` which works on both systems.

unphased:
This unbroke me on Ubuntu 14.04

Thank you

PR #1334, opened 2017-06-19
Make ssh-add use $SSH_ASKPASS

### LINKED ISSUE #1320: SSH module asks for passphrase (ksshaskpass)
With `SSH_ASKPASS=/usr/bin/ksshaskpass`, get asked for my passphrase. The SSH module currently invokes ssh-add like that, so i open a shell to the “Enter passphrase” prompt with it active:

```console
$ ssh-add
Enter passphrase for ~/.ssh/id_rsa:
```

But when invoked without a terminal attached to stdin, `ssh-add` invokes ksshaskpass:

```console
$ ssh-add </dev/null
```

flying-sheep:
Fixes #1320

## Proposed Changes

- Make `ssh-add` use `$SSH_ASKPASS` instead of using terminal input

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -40,9 +40,9 @@ fi
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
   if (( ${#_ssh_identities} > 0 )); then
-    ssh-add "$_ssh_dir/${^_ssh_identities[@]}" 2> /dev/null
+    ssh-add "$_ssh_dir/${^_ssh_identities[@]}"  < /dev/null 2> /dev/null
   else
-    ssh-add 2> /dev/null
+    ssh-add  < /dev/null 2> /dev/null
   fi
 fi
 


belak:
Thanks for your contribution! I've been playing around with this without setting SSH_ASKPASS and it still seems to work alright.

flying-sheep:
great!

crivotz:
I noticed that at each ssh session I was prompted for the passphrase.
I've removed the changes made by @flying-sheep and everything is back to work properly (passphrase required only once).
To make it work with the edit I had to install `ssh-askpass`

Can be a good thing add it to the module readme?

indrajitr:
Yes, good idea. A PR would be appreciated :)

crivotz:
No problem for PR but am I the only one that encountered this type or problem?

indrajitr:
I am guessing the problem might be existing, just that most are not encountering it either because of the workflow ('x11-ssh-askpass' or a variant is already installed) or is on macOS or ssh module isn't enabled.
I am usually on macOS where this doesn't happen. And I frequently ssh into Arch Linux, Ubuntu etc. but I don't have prezto there.

crivotz:
@indrajitr thanks. I will do the PR. Just for  information I am on Debian

crivotz:
PR #1408 created. Thanks

PR #1335, opened 2017-06-19
Change print to printf in init.zsh error

samcv:
When my zsh had issues, and I was debugging it, this line would
end up being run but no output would be made to the screen.

Changing it from print to printf caused the error to properly be
displayed.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -12,7 +12,7 @@
 # Check for the minimum supported version.
 min_zsh_version='4.3.17'
 if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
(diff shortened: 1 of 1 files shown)


belak:
Alright, makes sense. :) Thanks for your contribution!

samcv:
I'm not sure exactly why this happened, maybe it has something to do with `printf` existing as a standalone program as well as a builtin? Hmm.

belak:
I originally assumed that print didn't send a newline but that doesn't appear to be the case. I don't completely understand why it fixes things either, but it's a small change which should be supported pretty much everywhere.... and I like printf better anyway. :)

PR #1338, opened 2017-06-21
Move virtualenv auto-switch cwd hook optional loader out from inside unrelated if statement

sgtsquiggs:
It seems that a bad merge somewhere placed the python virtualenv auto-switch loader code under the pip if statement. It shouldn't be there so I've moved it to directly under the cwd hook code.

I missed this in #1300.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -76,6 +76,12 @@ function _python-workon-cwd {
   fi
 }
 
+# Load auto workon cwd hook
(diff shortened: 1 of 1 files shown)


belak:
I suppose this makes sense, though there may be unintended side effects if you don't have virtualenvwrapper and want to use this. That may be worth investigating in the future.

Anyway, thanks for the fix!

PR #1339, opened 2017-06-21
Add cwd hook to change node version via nvm

sgtsquiggs:
This is based on the python virtualenv auto-workon cwd hook.

It isn't *perfect*: if you change your .nvmrc file it will not pick up the change until you cwd outside of the project, which in turn returns you to the 'default' alias in nvm. This imperfection is because nvm's resolving of .nvmrc is too slow to fire on every cwd. Instead, I'm caching the absolute path of the relevant .nvmrc.

### DIFF
--- a/modules/node/init.zsh
+++ b/modules/node/init.zsh
@@ -28,6 +28,48 @@ elif (( ! $+commands[node] )); then
   return 1
 fi
 
+# Auto node version switch cwd hook
+function _node-nvm-switch-cwd {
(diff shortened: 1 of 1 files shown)


belak:
Hm, it seems like the project-root stuff could be refactored into a function if that's going to be shared between a number of other tools... and I could see it as being useful elsewhere.

I'm also a terrible person to test this because I don't use node very often.

PR #1341, opened 2017-06-21
Fix #1337 by removing top alias to htop.

### LINKED ISSUE #1337: modules/utility htop is aliased to top
### Description

The names of existing programs should not be aliased to *other* programs. For something that aims to be an easy to install and configure zsh configuration, it should not go about aliasing one program to another. If I want to run top I run top. When I want to run htop I run htop. If the user wants to add an alias and *overwrite* an existing program that is fine, but the default setup should not do this.

### Expected behavior

Expected top to run.

### Actual behavior

htop ran

### Versions

  - Prezto commit: a75c637
  - ZSH version: 5.3.1
  - OS information: Linux x64

diraol:
- Remove the `alias top=htop` behavior from the utility module.

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -107,11 +107,9 @@ Aliases
 
   - `df` displays free disk space using human readable units (aliases to `pydf`, if installed).
   - `du` displays disk usage using human readable units.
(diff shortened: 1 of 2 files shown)


belak:
Thanks for your contribution! I appreciate it. :)

PR #1342, opened 2017-06-22
Update external repos for syntax-highlighting, autosuggestions, and completion

samcv:
Changes pulled in have been added as a shortlog in each commit's body.


belak:
Thanks for the contribution!

PR #1344, opened 2017-06-23
[zprezto-update] Add convenience function to update zprezto

samcv:
This function checks if there is any update to zprezto, and if so
will pull in the changes. It will not attempt a pull unless
it is fastforwardable. It also makes sure the user is on the master
branch before attempting.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -18,28 +18,51 @@ fi
 unset min_zsh_version
 
 # zprezto convenience updater
-function zprezto-update {
-    local OWD="$(pwd)"
-    cd -- "${ZPREZTODIR}"
-    local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
-    if [[ "$orig_branch" == "master" ]]; then
-        git fetch
-        if [[ "$(git status -uno | grep behind)" ]]; then
-            printf "There is an update availible. Trying to pull"
-            if git pull --ff-only; then
-                printf "Syncing submodules\n"
-                git submodule update --recursive
-            else
-                printf "Unable to fast-forward the changes. You can fix this by running 'cd %s' and then 'git pull' to manually pull and possibly merge in changes\n"
-            fi
-        else
-            printf "There are no updates.\n"
-        fi
+# The function is surrounded by ( ) instead of { } so it starts in a subshell
+# and won't affect the environment of the calling shell
+function zprezto-update () (
+  function cannot-fast-forward {
+    local STATUS="$1"
+    [[ "${STATUS}" ]] && printf "%s\n" "${STATUS}"
+    printf "Unable to fast-forward the changes. You can fix this by "
+    printf "running\ncd '%s' and then\n'git pull' " "${ZPREZTODIR}"
+    printf "to manually pull and possibly merge in changes\n"
+  }
+  cd -q -- "${ZPREZTODIR}" || return 7
+  local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
+  if [[ "$orig_branch" == "master" ]]; then
+    git fetch || return "$?"
+    local UPSTREAM=$(git rev-parse '@{u}')
+    local LOCAL=$(git rev-parse @)
+    local REMOTE=$(git rev-parse "$UPSTREAM")
+    local BASE=$(git merge-base @ "$UPSTREAM")
+    if [[ $LOCAL == $REMOTE ]]; then
+      printf "There are no updates.\n"
+      return 0
+    elif [[ $LOCAL == $BASE ]]; then
+      printf "There is an update availible. Trying to pull.\n\n"
+      if git pull --ff-only; then
+        printf "Syncing submodules\n"
+        git submodule update --recursive
+        return $?
+      else
+        cannot-fast-forward
+        return 1
+      fi
+    elif [[ $REMOTE == $BASE ]]; then
+      cannot-fast-forward "Commits in master that aren't in upstream."
+      return 1
     else
-        printf "zprezto install at '%s' is not on the master branch (you're on '%s') Unable to automatically update.\n" "${ZPREZTODIR}" "${orig_branch}"
+      cannot-fast-forward "Upstream and local have diverged."
+      return 1
     fi
-    cd -- "${OWD}"
-}
+  else
+    printf "zprezto install at '%s' is not on the master branch " "${ZPREZTODIR}"
+    printf "(you're on '%s')\nUnable to automatically update.\n" "${orig_branch}"
+    return 1
+  fi
+  return 1
+)
 #
 # Module Loader
 #
--- a/init.zsh
+++ b/init.zsh
@@ -17,6 +17,29 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
 fi
 unset min_zsh_version
 
+# zprezto convenience updater
+function zprezto-update {
+    local OWD="$(pwd)"
+    cd -- "${ZPREZTODIR}"
+    local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
+    if [[ "$orig_branch" == "master" ]]; then
+        git fetch
+        if [[ "$(git status -uno | grep behind)" ]]; then
+            printf "There is an update availible. Trying to pull"
+            if git pull --ff-only; then
+                printf "Syncing submodules\n"
+                git submodule update --recursive
+            else
+                printf "Unable to fast-forward the changes. You can fix this by running 'cd %s' and then 'git pull' to manually pull and possibly merge in changes\n"
+            fi
+        else
+            printf "There are no updates.\n"
+        fi
+    else
+        printf "zprezto install at '%s' is not on the master branch (you're on '%s') Unable to automatically update.\n" "${ZPREZTODIR}" "${orig_branch}"
+    fi
+    cd -- "${OWD}"
+}
 #
 # Module Loader
 #
--- a/README.md
+++ b/README.md
@@ -50,9 +50,16 @@ window or tab.
 Updating
 --------
 
-Pull the latest changes and update submodules.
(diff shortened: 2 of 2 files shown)


Yoshyn:
Pretty cool PR. 
Just one question : I have modification on theses files : 
  * runcoms/zpreztorc (adding module, & co ...)
  * runcoms/zshrc        (export EDITOR & VISUAL...

So, if i run this it fail due to : 
```
error: cannot pull with rebase: You have unstaged changes.
error: please commit or stash them.
```

Manually, i stash it and reapply after.
I think that a lot people modify theses files. Do you plan to manage it?

diraol:
Following up @Yoshyn comment, what could have been done here is something like what is done on `spf13-vim` distro.
We could have a `zprestorc.local` and a `zshrc.local` files that are imported on the end of the `zprestorc` and `zshrc` files. Doing so, the `.local` won't be versioned and then the users personalized configurations wouldn't get in conflict with the main config files, making this updates easier.
Does it make sense?

Also, what do you think about printing, as output from the update command, the messages from the commits that were pulled, so the user can know what was updated? Does it make sense, at least with an optional argument?

Yoshyn:
Well, initially i though about someting like that : 
```
printf "Try to stash before pulling..."
local zprezto_stash="zprezto-update-`date +%s`"
git stash save "$(echo $zprezto_stash)" --include-untracked
if git pull --ff-only; then
    git submodule update --recursive
    git stash pop
else
    printf "Unable to fast-forward the changes. You can fix this by running 'cd %s' and then 'git pull' to manually pull and possibly merge in changes\n"
fi
```

But it's just unmaintainable.  
After reading @diraol, i'm completely convinced. A local setup not versioned will be awesome

init.zsh
```diff
@@ -17,6 +17,29 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
 fi
 unset min_zsh_version
 
+# zprezto convenience updater
+function zprezto-update {
+    local OWD="$(pwd)"
+    cd -- "${ZPREZTODIR}"
```
belak: `pushd` and `popd` would probably be better for this. Then we don't need to track `OWD`. Also note that we should use `-q` to ensure the chpwd functions are not run.

init.zsh
```diff
@@ -17,6 +17,29 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
+    local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
+    if [[ "$orig_branch" == "master" ]]; then
+        git fetch
+        if [[ "$(git status -uno | grep behind)" ]]; then
+            printf "There is an update availible. Trying to pull"
+            if git pull --ff-only; then
+                printf "Syncing submodules\n"
+                git submodule update --recursive
+            else
+                printf "Unable to fast-forward the changes. You can fix this by running 'cd %s' and then 'git pull' to manually pull and possibly merge in changes\n"
```
belak: Looks like this is missing the param for the `%s`

init.zsh
```diff
@@ -17,6 +17,29 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
 unset min_zsh_version
 
+# zprezto convenience updater
+function zprezto-update {
+    local OWD="$(pwd)"
+    cd -- "${ZPREZTODIR}"
+    local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
+    if [[ "$orig_branch" == "master" ]]; then
+        git fetch
+        if [[ "$(git status -uno | grep behind)" ]]; then
```
belak: This check seems a bit error prone. Maybe something like what's mentioned here would be better: https://stackoverflow.com/a/3278427 (on a side note, I had no idea `@` even existed in git)

samcv: Took that suggestion and implemented it in my latest commit.

init.zsh
samcv: Fixed this in the latest commit.

init.zsh
samcv: Fixed.

samcv:
Fixed @belak's suggestions in my latest commit and cleaned up a few things regarding the error handling. As stated in the commit description.

init.zsh
```diff
@@ -17,6 +17,45 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
+function zprezto-update {
+   function cannot-fast-forward {
+       local STATUS="$1"
+       [[ "${STATUS}" ]] && printf "%s\n" "${STATUS}"
+       printf "Unable to fast-forward the changes. You can fix this by "
+       printf "running\ncd '%s' and then\n'git pull' " "${ZPREZTODIR}"
+       printf "to manually pull and possibly merge in changes\n"
+   }
+   pushd -q
+   cd -- "${ZPREZTODIR}"
```
belak: This would be better as `pushd -q -- "${ZPREZTODIR}"` so cd won't cause chpwd hooks to run (cd also has a similar option, but they can be combined, so it makes sense to)

init.zsh
```diff
@@ -17,6 +17,45 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
 fi
 unset min_zsh_version
 
+# zprezto convenience updater
+function zprezto-update {
+   function cannot-fast-forward {
+       local STATUS="$1"
+       [[ "${STATUS}" ]] && printf "%s\n" "${STATUS}"
```
belak: Should this be `[[ -n "${STATUS}" ]]`?

init.zsh
```diff
@@ -17,6 +17,45 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
+       elif [[ $REMOTE == $BASE ]]; then
+           cannot-fast-forward "Commits in master that aren't in upstream."
+       else
+           cannot-fast-forward "Upstream and local have diverged."
+       fi
+   else
+       printf "zprezto install at '%s' is not on the master branch " "${ZPREZTODIR}"
+       printf "(you're on '%s')\nUnable to automatically update.\n" "${orig_branch}"
+   fi
+   popd -q
```
belak: It would be nice to have proper return codes if this fails. `popd` could be moved to an atexit function (`trap 'popd -q' EXIT`) which would make it easier to return without having to copy the popd everywhere.

init.zsh
samcv: `[[ -n "${STATUS}" ]]` is the same thing as `[[ "${STATUS}" ]]`.

Should I explicitly add the -n? (this works without -n even in posix shell by the way)

belak: This is fine, I think... because we control the value of STATUS. If we didn't the -n would make sense.

init.zsh
samcv: Okay, I had to make a change and not run pushd -q earlier on to make popd -q work inside the trap. Otherwise it would take me back to the wrong folder.

runcoms/zprofile
```diff
@@ -65,12 +65,9 @@ fi
 #
 # Temporary Files
 #
-#
-if [[ -z "$TMPDIR" ]]; then
```
belak: This change seems unrelated...

init.zsh
```diff
@@ -17,6 +17,51 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
+# zprezto convenience updater
+function zprezto-update {
+  function cannot-fast-forward {
+    local STATUS="$1"
+    [[ "${STATUS}" ]] && printf "%s\n" "${STATUS}"
+    printf "Unable to fast-forward the changes. You can fix this by "
+    printf "running\ncd '%s' and then\n'git pull' " "${ZPREZTODIR}"
+    printf "to manually pull and possibly merge in changes\n"
+  }
+  cd -- "${ZPREZTODIR}" || return 7
```
belak: This will still trigger chpwd callbacks (which we don't want to run). There's an option to cd which won't trigger them... Also, I'm a bit confused why pushd didn't work...

samcv:
I've added the `-q` option to `cd`. Instead of using pushd and trying to popd on exit, I changed it to use `( )` around the function which keeps the contents in the function outside the external scope and avoids the issue of having to change back. Since the `cd` happens inside a different scope.

belak:
Alright, a subshell works. I'd like to see a comment mentioning that (because it's easy to miss that `()` was used rather than `{}`.

Once that's done and a small blurb is added to the readme (or somewhere relevant) I think this is ready.

samcv:
@belak I've added information to the README and added a comment to `init.zsh` about it being in a subshell.

init.zsh
```diff
@@ -17,6 +17,52 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
...
+    local STATUS="$1"
+    [[ "${STATUS}" ]] && printf "%s\n" "${STATUS}"
+    printf "Unable to fast-forward the changes. You can fix this by "
+    printf "running\ncd '%s' and then\n'git pull' " "${ZPREZTODIR}"
+    printf "to manually pull and possibly merge in changes\n"
+  }
+  cd -q -- "${ZPREZTODIR}" || return 7
+  local orig_branch="$(git symbolic-ref HEAD 2>/dev/null | cut -d '/' -f 3)"
+  if [[ "$orig_branch" == "master" ]]; then
+    git fetch
```
belak: Sorry, forgot about this one. We should probably return if git fetch fails.

samcv: Done.

belak:
Merged! Thanks for your work on this!

samcv:
We can probably have an enhancement based on @Yoshyn's suggestion to possibly stash things and then reapply but I'd think it'd be best to keep it simple for at least a little while and make sure there's no issues with things as they have been merged in currently. Since at least we know it won't make any mistakes if it aborts in case of conflicts.

belak:
Yep! I figured, since this is new and not used by many people it's alright to have a few non-critical bugs... and this PR has had enough feedback on it. :)

PR #1345, opened 2017-06-23
[editor] Set Delete key in vicmd mode to delete character

samcv:
Delete key deletes character in vimcmd cmd mode instead of weird
default functionality. The default functionality in vicmd mode
is the Delete key will change the case of many of the characters
on the screen, which is not the default thing that vim does.

This could be confusing and frustrating to users, so set it to delete
a character instead.


belak:
This makes sense. I'm not really sure why it's not set by default. Thanks!

PR #1346, opened 2017-06-23
Update zpreztorc

Yoshyn:
Make a .zpreztorc.local that are not versioned in order to make update on prezto easier to update.

This PR is a discuss proposal related to https://github.com/sorin-ionescu/prezto/pull/1344
This must not be merged.

How can we manage the other files in runcoms?

### DIFF
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -7,6 +7,13 @@
 
 #
 # General
+# This file define the mains settings for zprezto
+# If you want customization follow theses steps :
+#   * cd "${ZDOTDIR:-$HOME}"
+#   * cp .zpreztorc .zpreztorc.local
+#   * Then edit the .zpreztorc.local
+# It will avoid you to have trouble when trying to update
+# Prezto
 #
 
 # Set case-sensitivity for completion, history lookup, etc.
@@ -186,3 +193,8 @@ zstyle ':prezto:module:prompt' theme 'sorin'
 
 # Set the default session name:
 # zstyle ':prezto:module:tmux:session' name 'YOUR DEFAULT SESSION NAME'
+
+# Load zpreztorc.local
+if [[ -s "${ZDOTDIR:-$HOME}/.zpreztorc.local" ]]; then
+  source "${ZDOTDIR:-$HOME}/.zpreztorc.local"
+fi


diraol:
@Yoshyn I think work doing the same with the other files that may be used to setup the user environment, namely:

  * zlogin
  * zlogout
  * zprestorc
  * zprofile
  * zshrenv
  * zshrc

Does it make sense?

Yoshyn:
I'm not sure about that. 
For exemple zshenv source ~/.zprofile so I think that we don't have to manage it.

What should be great is avoid to define ours rules but take the rules of others projects that have already reflect on it.

For exemple concerning the naming : .zpreztorc.local is pretty ugly (at least for me)
Maybe .zpreztorc-profile would be better (probably cause i thought to bash-profile)

diraol:
That makes sense! =)

belak:
As far as I know, `.local` is fairly standard. It may make sense to add these to most of the runcoms. I personally haven't run into this because I don't actually use any of the runcoms (shh, don't tell anyone ;)) and I load modules with antigen.

`.bash-profile` was intended to be `.profile` (something from sh) for bash, so it's not quite the same thing. If we move forward with this, I'd prefer `.local`.

Also thanks for the idea! This solves some of my long running complaints with the runcoms.

samcv:
@Yoshyn if you want to look at my other PR https://github.com/sorin-ionescu/prezto/pull/1352 I don't know if we would want to have those variables stored inside the .local file? though I'd assume we'd want that run *afterwards*? Wanted to know what you think. Though personally I think this thing is best handled with hand merges but that may not work for people who don't know git very well. So I'm at a loss for how to handle zpreztorc and zpreztorc.local, since things that load in zpreztorc need the data available beforehand, and .local is usually loaded afterward?

Yoshyn:
* if you want to look at my other PR #1352 I don't know if we would want to have those variables stored inside the .local file? though I'd assume we'd want that run afterwards? Wanted to know what you think. 
=> I do not understand the advantage of storing theses value into an array. It's not a bad think but i do not see how it help for the current concern.

* I think this thing is best handled with hand merges but that may not work for people who don't know git very well.
=> Exactly my point.

* So I'm at a loss for how to handle zpreztorc and zpreztorc, since things that load in zpreztorc need the data available beforehand, and .local is usually loaded afterward?
=> For me zpreztorc should load zpreztorc.local at the end of the file and override value thank to zstyle but maybe i did not well understand the purpose of zstyle.

diraol:
@Yoshyn about what should be done with the `.local`, what if there is a 'default' behavior on the `zprestorc` file that the user does not want (load a module, for example)? That should be a problem right, since that the `zprestorc.local` will not unload the already loaded module.

Being so, may be the `zprestorc` should check for each variable on the `zprestorc.local` file before taking any action?

belak:
> what if there is a 'default' behavior on the `zprestorc` file that the user does not want (load a module, for example)? That should be a problem right, since that the `zprestorc.local` will not unload the already loaded module.

zpreztorc doesn't actually do any of the loading, so if you override the zstyle setting, it will be fine. The actual loading happens in init.zsh after zpreztorc has been loaded.

The rest of this is mostly a mind dump with what I'd like to see happen to the runcoms. Suggestions and comments are welcome.

The way I see it, there are a few ways of "solving" this issue which I'll do my best to outline below:

* Add `.local` files for all the runcoms. This lets us change defaults and add new options easily but some defaults will be missed because they would be silently overridden in the `.local` files.
* Stop recommending that runcoms should be symlinked. This raises some problems about how to change defaults or let people know new options have been added.
* Leave it as is.
* Rework the runcoms so most of them just bootstrap into prezto (removing the random `fortune` calls and what not) and zpreztorc is used for all settings.

The problems with the runcoms I would like to see solved:

* Make it easier for people who don't understand git and merging to use prezto (without sacrificing the flexibility we currently have for power users)
* Make it possible to change defaults without breaking and (maybe) inform the user
* Make it possible to add options without breaking (maybe) and inform the user
* The changes cannot be extremely inconvenient or annoying... I don't want to get issues filed because people are super annoyed with how verbose it gets. I'd like to avoid another issue like the controversial oh-my-zsh auto-updater.

The solutions I listed above don't solve all these problems. At most, they solve one or two. Are there any suggestions (or variations on the above solutions) which aim to solve as many of these as possible without sacrificing too much?

diraol:
> zpreztorc doesn't actually do any of the loading, so if you override the zstyle setting, it will be fine. The actual loading happens in init.zsh after zpreztorc has been loaded.

ok, no problem with that so! =)

> Make it possible to change defaults without breaking and (maybe) inform the user
> Make it possible to add options without breaking (maybe) and inform the user

Just trying to understand... what do  you mean by 'without breaking' here? I mean, how would these problems occur? Any examples so we can try to think on solutions?
I'm not a so much advanced user on how zsh/prezto works under the hood, that's why I'm asking ;)

belak:
As an example, there was a push to change the default alias of `rm` (and similar tools) to remove `-i` by default. However, I don't like changing defaults in a way that could break people. It would be nice if there was a way to notify people of the change without them having to check the commit log.

Note that the list above is more of a wish list... I realize we probably can't fix all of them, but I'd like a solution which solves as much as possible.

samcv:
@belak we could maybe add a hook into `zprezto-update` that would give certain notifications? Or even like zprezto-news or something which could show them whenever they wanted to find out, but also show them when they update with zprezto the notification?

Or we could have a folder in $ZPREZTODIR/news which would contain text files that are news entries and if there are files in that folder we could show a notification of unread messages. That would allow people to either use our yet to be made interface for reading news or just clearing the folder.

diraol:
I remember that I read at somewhere a discussion about defining some release-rules for prezto. Considering this, I think it would be nice to have a CHANGELOG file in order to keep track of the changes.

If we have this CHANGELOG file, after a zpresto update then we can just print the changelog diff on the screen (or a formatted version of this "diff").

To ease the tracking of versions we can store the 'current commit hash' on a '.current_hash' file. When the users start a new zpresto term, then we compare the latests commit hash (from git log) to this '.current_hash'. If they are different, then we show the CHANGELOG diff (formatted) and update the '.current_hash' file with the latest commit hash.
This file would be ignored on `.gitignore`.

diraol:
Hey all, any new ideas over here?

I've seem some discussions going around some issues/PRs that would benefit from some changes proposed on this PR. =)

jeeftor:
I second the question @diraol I kind of like the `.local` idea because it seems like every time i update there is a good number of changes.

johnpneumann:
I'm in favor of a .local, however, I'm against giving users notifications as a default. That needs to be an opt-in feature as we don't want to be sending data across the network without the user *explicitly* saying it's ok to do so. In terms of a news/changelog, that needs to be discussed in a different issue otherwise this is going to off the rails with features.

belak:
@johnpneumann The notifications which were described above would be local only and would happen when they pull. There would be no network traffic. I would also like to avoid the oh-my-zsh auto-update controversy if possible.

One thing that comes to mind... if we have a `zpreztorc.local`, what is the `zpreztorc` going to be used for? It seems entirely superfluous. Would it be better to remove zpreztorc entirely (because it's all commented out default settings) and/or have users copy it?

diraol:
@johnpneumann as @belak said, there wouldn't be any network traffic for the notifications to go on. After an active request by the user to update its own prezto, the update would be handled by git versioned information.

@belak about the `zprestorc` vs `zpreztorc.local`, I've taken the idea of a `.local` from [`vim-spf13`](https://github.com/spf13/spf13-vim). Over there, the `.vimrc` file is used to store the default configs (not commented versions of it). It can be seen [here](https://github.com/spf13/spf13-vim/blob/3.0/.vimrc).

One advantage of keep using `zprestorc` is to join all defaults at the same place. This would make the life of prezto uses easier because they do not need to look for defaults they didn't even know the existence for all the files around.

Another point I would like to bring is that I wouldn't leave the defaults commented out on the `zprestrc` file. Sometimes it can be confusing for the user to understand if the commented version is the default or one option that can be used if uncommented.

diraol:
Just adding one more info regarding #1354.

By using `zprestorc` and setting there all the 'defaults', it would also make it easier to document the differences between two releases, in terms of settings/options. We will only need to diff between the previous stable release and the 'current one being released'.

jeeftor:
Does this remove the need for the unit function where the runcomms version of the files are then pre populated so you can do a dif?

belak:
Good suggestions, @diraol. I think adding all the defaults to the zpreztorc and suggesting that users use zpreztorc.local would be a good option.

Still not sure what to do with the other runcoms, but this should be a good starting point.

diraol:
About the other runcoms, I would keep them just if they are "executed" on distinct moments and they need to be executed at that moments. Otherwise, I would merge them with zprestorc (maybe splitted by "sections" [just comments to organize the content]).

belak:
What about removing some of the oddities from the runcoms? The calls to fortune and printing messages (in zlogin and zlogout respectively) seem a bit strange to put into a config framework.

belak:
Also, thinking about it... setting all the values in the zpreztorc will make it harder for people using plugin managers (as normally you set values in your zshrc then load prezto).

diraol:
@belak, I do not use plugin managers, so can you give some examples why the proposal can be problematic with this tools?

wadkar:
Quick question: I decided to drop zgen and use prezto directly, and I am not sure where to put my existing prezto customization. Should I directly edit the runcoms, specifically, .zpreztorc? I have no issues managing git conflicts.

If yes, can we update the root level README.md to instruct new users to directly edit the runcoms while we make progress on this PR?

diraol:
Hi all, any news about this issue? What do we need to move forward here? Can I help with something? =)

Yoshyn:
This issue was more a discussion thread than a real issue from the beginning.
I think we can close it and reopen another more specific if needed.

belak:
Definitely. I'd be happy to continue this discussion in an issue if this is something people want.

PR #1347, opened 2017-06-23
[runcoms/zpreztorc] Suggest changing the key timeout for vi mode users

samcv:
Set the time for zsh to wait for additional keypresses. In vi mode you
should set this to 1 (0.1s) to remove the delay from entering and
exiting vicmd mode. The default is 0.5s.
Recommended in vi mode, but setting too low in emacs mode will not allow
time to enter additional keypresses.


belak:
Most of the info I can find about this seems to imply that setting KEYTIMEOUT might mess with other things. https://superuser.com/a/648046

It may be worth looking into some of those workarounds as well.

samcv:
On Wednesday, June 28, 2017 2:08:05 PM PDT Kaleb Elwert wrote:
> Most of the info I can find about this seems to imply that setting KEYTIMEOUT might mess with other things. https://superuser.com/a/648046
> 
> It may be worth looking into some of those workarounds as well.
> 
> 
That answer is not applicable because in zprezto escape key *is* bound to something. Entering vicmd mode. And so there aren't any more problems other than not having enough time to enter other commands if there are other commands bound to the escape key.

For me I have:

	"^[," _history-complete-newer
	"^[/" _history-complete-older
	"^[OA" history-substring-search-up
	"^[OB" history-substring-search-down
	"^[[A" up-line-or-history
	"^[[B" down-line-or-history

So maybe we want to make sure these are set to non-escape keys, and that may help the issue? Though I still think it may be useful to put a suggestion for KEYTIMEOUT though maybe setting it to 1 is too low? We could maybe just say the default is 40, and if they have problems with keyboard lag they can lower it?

hlecuanda:
There is a very good reason for the KEYTIMEOUT to be set at 500ms, for `vi` mode, namely  **Multiple stroke keybindings** and **In-seq / Out-seq keybindings**

Consider the following cases:
 
### `[ lead ]` like bindings to group similar functions on a logical prefix
```zsh
bindkey -M vicmd ",s" prepend-sudo           # bring me a sandwich https://xkcd.com/149/
bindkey -M vicmd ",c" vi-pound-insert        # distilled commenting awesomeness
bindkey -M vicmd ",p" copy-prev-shell-word   # use shell parsing instead of space matching to find prev word
```
This sort of bindings are very common, specially if you find that your mnemonic space starts having collisions, so how to reconcile standard `[ p ]ut register` with `[ p ]revious-shell-word`? by expanding the mnemonic memory space via prefixes or "sub modes" below 200ms or 300s of a second and you'll begin mis-triggering some  bindings, because of timeouts. It's worse if you consider `-s` or In-sequence/Out-sequence bindings that can have an arbitrary length as long as you can enter the complex binding in the span of time between typing the first character and the moment the first sequence character is read and `KEYTIMEOUT`

### sequence-in bindings 
this kind of bindings can be "overloaded" by triggering other bindings, as the out-sequence is interpreted as user keyboard input. If timeout is too short, it can become difficult to properly trigger something like 
    **[ esc ][ - ] [ ! ][ ! ]** 
as **[ ! ]** and **[ - ]** are on opposite ends of the keyboard and both require a **[ Shift ]** modifier on most laptops. 1/10th of a second is plainly not enough, and it doesn't get better with user's age , i know n.n
```zsh
bindkey -M vicmd -s  "!!" "\e\e0Csudo !! "           # prepend sudo to the previous history command
bindkey -M vicmd -s ",!!" "\e\egg0isudo \eG$a"       # prepend sudo on a multi-line command by vi'ing your way around
bindkey -M vicmd -s "-!!" "\e\e sudo !!:0 !!:2 !!:1" # prepend sudo to prev history while swapping arguments
```
### stock gold standard
`vim` timeout is way more granular controlled via `'timeout' 'ttimeout' 'timeoutlen'' and 'ttimeoutlen parameters, but defaults to stock `vi`'s default, which is immutably set at compile time at 1000ms. this is to allow for slower terminals. 5 years ago even 500ms  were a good compromise. but today's workloads are shifting to mobile, the bottleneck is not on the serial cable but on the soft-keyboard's response and tiny keys that make hitting the right key fast enough somewhat difficult. 

### TL;DR
IMHO, `KTIMEOUT` should be set between 800 and the default 1000 mostly to accommodate mobile workloads. (prezto makes shell usage ridiculously easy with keybindings, tmux and mosh automation)

Issue #1331, opened 2017-06-15
prompt sorin theme broken on FreeBSD TMPDIR doesn't get set


samcv:
### Description
The problem is that TMPDIR does not get set. I added a print inside `modules/prompt/functions/prompt_sorin_setup` to print out `TMPDIR`, and it ended up being empty.
I made a folder ~/tmp and then added a line at the top of `modules/prompt/functions/prompt_sorin_setup` to set `TMPDIR="$HOME/tmp"` this resolved the errors.

### Expected behavior

No error.

### Actual behavior

```
mktemp: mkstemp failed on /sorin-prompt-async-2sa6cJ5bWe: Permission denied`
prompt_sorin_precmd_async:4: no such file or directory:`
```

### Steps to Reproduce

1. Use zprezto with sorin theme
2. No further steps required
3. Message shows after every command, or display of prompt (pressing enter with no text entered)

### Versions

  - Prezto commit: 2f20a80f7255bffcf878de70192f6df126c6069f
  - ZSH version: zsh 5.3.1
  - OS information: FreeBSD 11.0

belak:
Do you have all the runcoms set up? TMPDIR is currently set in zprofile, so if you're only using zshrc it may not be getting set properly.

However, this seems a bit error prone so I'll look at improving it, maybe by falling back to /tmp in the prezto theme.

samcv:
Yes, I have .zpreztorc, .zshrc, and it's all setup and loading. So maybe the problem is not specifically in that theme, but everything except TMPDIR being set are working fine.

belak:
I've updated the sorin theme to default to /tmp if it isn't set, though I'm not sure why it wouldn't be getting set properly in .zprofile.

Issue #1180, opened 2016-08-20
Are non-login, non-interactive shells being detected correctly?


leoj3n:
Regarding this test: https://github.com/sorin-ionescu/prezto/commit/3a5fff9c6d8e62032f0173476aed73afcd00578e#diff-9b19914007506af01febc6dabc80fd47R9

The command:

```
zsh -c '[[ "$SHLVL" -eq 1 && ! -o LOGIN ]] && \
echo "I think I am non-interactive, non-login!" \
|| echo "I think not"'
```

will echo `I think not`. However, if wrapped in a subshell like:

```
(zsh -c '[[ "$SHLVL" -eq 1 && ! -o LOGIN ]] && \
echo "I think I am non-interactive, non-login!" \
|| echo "I think not"')
```

the command will echo `I think I am non-interactive, non-login!`.

What is going on here? I'm not exactly sure if this is the correct behavior.

leoj3n:
I guess I am confused as to why `$SHLVL` doesn't increment for a shell spawned in a subprocess of another shell and whether or not that is a problem or bug for prezto. In more reduced code terms:

Why does
```
(zsh -c 'echo $SHLVL') && zsh -c 'echo $SHLVL'
```

Result in:

```
1
2
```

?

leoj3n:
The original commit's extended message says:


> A top level shell is usually a login shell, but can also be a
non-interactive, non-login shell in certain cases, such as when
executing an SSH remote command.

What are the "certain cases" other than SSH?

leoj3n:
@sorin-ionescu Would you accept a change like this?

```sh
# Ensure that a non-script, non-interactive, non-login shell has a defined environment.
if [[ $ZSH_EVAL_CONTEXT =~ :file$ && ! -o INTERACTIVE && ! -o LOGIN ]]; then
  [[ -s "${ZDOTDIR}/.zprofile" ]] && source "${ZDOTDIR}/.zprofile"
fi
```

leoj3n:
Check this out https://github.com/zsh-users/zsh/blob/master/Functions/Chpwd/chpwd_recent_dirs#L29-L36

leoj3n:
```sh
ssh localhost 'echo "$SHLVL $ZSH_SUBSHELL $ZSH_EVAL_CONTEXT $-"'
# 1 0 cmdarg 569X
```

```sh
ssh localhost -t
[...]

PR #1350, opened 2017-06-26
[runcoms/zprofile+zshenv] Ensure TMPDIR is always set

samcv:
In some cases TMPDIR may not be set, in which case it could cause issues
for zsh modules.

This change was prompted by issue #1331.

Also when creating a missing TMPDIR, use mkdir --mode=700 instead
of creating the directory and then chmoding it afterward.

### DIFF
--- a/runcoms/zprofile
+++ b/runcoms/zprofile
@@ -66,11 +66,11 @@ fi
 # Temporary Files
 #
 #
-if [[ -z "$TMPDIR" ]]; then
-  export TMPDIR="/tmp/zsh-$UID"
+# Set TMPDIR if the variable is not set/empty or the directory doesn't exist
+if [[ -z "${TMPDIR}" ]]; then
+  export TMPDIR="/tmp/zsh-${UID}"
 fi
 
-if [[ ! -d "$TMPDIR" ]]; then
-  mkdir "$TMPDIR"
-  chmod 700 "$TMPDIR"
+if [[ ! -d "${TMPDIR}" ]]; then
+  mkdir -m 700 "${TMPDIR}"
 fi
--- a/runcoms/zshenv
+++ b/runcoms/zshenv
@@ -6,6 +6,6 @@
 #
 
 # Ensure that a non-login, non-interactive shell has a defined environment.
-if [[ "$SHLVL" -eq 1 && ! -o LOGIN && -s "${ZDOTDIR:-$HOME}/.zprofile" ]]; then
+if [[ ( "$SHLVL" -eq 1 && ! -o LOGIN || -z "${TMPDIR}" ) && -s "${ZDOTDIR:-$HOME}/.zprofile" ]]; then
   source "${ZDOTDIR:-$HOME}/.zprofile"
 fi


belak:
`--mode` doesn't exist for the BSD mkdir used on macOS though there is a `-m` which appears to be shared between the GNU and BSD implementations

runcoms/zprofile
```diff
@@ -66,11 +66,11 @@ fi
 # Temporary Files
 #
 #
-if [[ -z "$TMPDIR" ]]; then
-  export TMPDIR="/tmp/zsh-$UID"
+# Set TMPDIR if the variable is not set or the directory doesn't exist
+if [[ ! "${TMPDIR}" ]]; then
```
belak: Why doesn't -z work? That should check if the string is zero length, which for an undefined or empty variable, it should be.

runcoms/zshenv
```diff
@@ -6,6 +6,6 @@
 #
 
 # Ensure that a non-login, non-interactive shell has a defined environment.
-if [[ "$SHLVL" -eq 1 && ! -o LOGIN && -s "${ZDOTDIR:-$HOME}/.zprofile" ]]; then
+if [[ "$SHLVL" -eq 1 && ! -o LOGIN && -s "${ZDOTDIR:-$HOME}/.zprofile" || ! "${TMPDIR}" ]]; then
```
belak: This check shouldn't be needed. -s checks if a file exists and has a size greater than zero, so it will be sourced if it exists.

runcoms/zprofile
samcv: `-z` works fine. `[[ ! "$var" ]]" is excatly equivalent to `[[ -z "$var ]]`. I can change it back though. I find the ! clearer but that's just me. I can change it to `-z` if people are happier.

belak: I'd like to start moving towards the Google Shell Style Guide which [recommends explicit use of -z and -n](https://google.github.io/styleguide/shell.xml?showone=Testing_Strings#Testing_Strings)

runcoms/zshenv
samcv: It will only be sourced if that file exists AND `[[ "$SHLVL" -eq 1 && ! -o LOGIN ]]`

Looking at it again though, this could cause it to try and source a non-existant file. I will make it so it requires .zprofile even when TMPDIR isn't set.

belak: There's another issue somewhere about changing this check because it doesn't completely make sense... but what you mentioned should be good for now (I don't know what happened to those other changes)

runcoms/zprofile
samcv: Works for me. Will do that going forward and ignore my personal style :)

samcv:
I also changed mkdir to use `-m` so it will work on freebsd and macos as well.

runcoms/zshenv
```diff
@@ -6,6 +6,6 @@
 #
 
 # Ensure that a non-login, non-interactive shell has a defined environment.
-if [[ "$SHLVL" -eq 1 && ! -o LOGIN && -s "${ZDOTDIR:-$HOME}/.zprofile" ]]; then
+if [[ "$SHLVL" -eq 1 && ! -o LOGIN || -z "${TMPDIR}" ]] && [[ -s "${ZDOTDIR:-$HOME}/.zprofile" ]]; then
```
belak: This could be defined in a single block using ()

samcv: Fixed to use a single block.

runcoms/zshenv
samcv: What's the issue with changing this check?

belak:
Thanks! Looks good! It'll be nice to get this TMPDIR stuff done with.

runcoms/zshenv
belak: No issue with changing it. Just mentioning it as something to keep in mind. https://github.com/sorin-ionescu/prezto/issues/1180

PR #1352, opened 2017-06-28
[zpreztorc] Use an array to load the plugins for ease of use

samcv:
Previously we used a newline separated list of arguments which had
the downside of requiring a backslash at the end of each newline,
and in many cases made it difficult to comment out lines to disable
particular plugins. By using an array we can easily comment them out
and not have to deal with using backslashes at the end of lines.

Note: it may be a good idea to list all of the plugins, but leave some
commented out, making it even easier for people to enable plugins?
Even if not, this would make it much easier to do things like store
the array in another file if we decide to changeup the setup of 
zpreztorc and have a zpreztorc.local or something.

### DIFF
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -22,17 +22,19 @@ zstyle ':prezto:*:*' color 'yes'
 # zstyle ':prezto:load' zfunction 'zargs' 'zmv'
 
 # Set the Prezto modules to load (browse modules).
-# The order matters.
-zstyle ':prezto:load' pmodule \
-  'environment' \
-  'terminal' \
-  'editor' \
-  'history' \
-  'directory' \
-  'spectrum' \
-  'utility' \
-  'completion' \
-  'prompt'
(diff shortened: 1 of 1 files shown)


indrajitr:
> Note: it may be a good idea to list all of the plugins, but leave some
commented out, making it even easier for people to enable plugins?

Yes, having them listed (but commented out) also makes it easier for people to figure out the relative ordering of the respective plugins, should they chose to enable.

samcv:
I've added at least the modules I use commented out. I have also converted all the other settings which
used to take multiple arguments split across lines with backslashes to continue onto the next line.

belak:
I like the general idea behind this change but I think it might be best to put this on hold while we figure out what we're doing with runcoms (including zpreztorc)

samcv:
I agree, we can hold off until we figure out what we are doing with the runcoms.

sorin-ionescu:
The reason why a zstyle was used is to keep the settings format consistent.

PR #1355, opened 2017-07-02
Fix README example on tmux's default session name

lfilho:


belak:
Thanks for your contribution! Merged!

Issue #1125, opened 2016-04-12
can't (easily) disable noglob on scp


ChrisJefferson:
I was confused for a while why I couldn't type:

`scp *.pdf remotehost.com:dir/`

Until I discovered that zprezto does a 'noglob' unconditionally on scp. That seems like slightly strange behaviour to me -- I can't think why I'd want that disabling?

I could (and will) maintain my own branch without the noglob, but I was wondering about the reasoning.

sorin-ionescu:
I don't use `scp` often. So, this may be a relic from OMZ. `noglob` is for programs that do their own globbing like `find`, which allows the user to use it without quotes.

ChrisJefferson:
Ah, this is annoying (I think I know where it comes from)

You wouldn't want globbing for ```scp remotehost.com:dir/* .``` But you do want globbing for ```scp *.pdf remotehost.com:dir/```. I'll have a think / google, to see if there is a cleverer option.

ChrisJefferson:
This stackoverflow answer has a clever answer, to check which parameters are remote, and which are local:

http://unix.stackexchange.com/questions/106980/scp-command-working-in-bash-doesnt-work-in-zsh

Would you (in principle) accept a patch which tidied this up / applied it to other commands (like rsync for example)

sorin-ionescu:
We can turn it into a function. Try that function, see if it works for you.

trustlix:
Here's a workaround I use:

`\scp *.txt remote:~/`

It will bypass the alias and call the scp command directly.

samcv:
This may be the same thing causing this, but I am getting an issue where using scp on *another* machine causes the command to fail. Because the computer i'm rsyncing **to** has zprezto installed on it…

```
$ rsync -az --chown=www-data:www-data --chmod=644 myfolder user@computer:/loc/ation  ✘ 127
zsh:1: no matches found: --usermap=*:www-data
rsync: connection unexpectedly closed (0 bytes received so far) [sender]
rsync error: error in rsync protocol data stream (code 12) at io.c(226) [sender=3.1.2]
```

samcv:
I've made a pull request which fixes this bug 😄 

https://github.com/sorin-ionescu/prezto/issues/1356

belak:
This should be fixed (in a slightly different way) thanks to @samcv.

Remote paths have glob disabled, while local paths should have them enabled.

PR #1356, opened 2017-07-02
[utility] scp/rsync glob local paths but don't glob remote paths

samcv:
Adds a function which wraps rsync and scp so that remote paths are not globbed
but local paths are globbed. This is because the programs have their own
globbing for remote paths. The wrap function globs args starting in / and ./
and doesn't glob paths with : in it as these are interpreted as remote paths
by these programs unless the path starts with / or ./

Fixes issue #1125

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -41,8 +41,26 @@ alias ftp='noglob ftp'
 alias history='noglob history'
 alias locate='noglob locate'
 alias rake='noglob rake'
-alias rsync='noglob rsync'
-alias scp='noglob scp'
+alias rsync='noglob rsync_scp_wrap rsync'
+alias scp='noglob rsync_scp_wrap scp'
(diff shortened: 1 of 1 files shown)


belak:
Black magic! But pretty awesome black magic. Nice improvement over the samples I've seen elsewhere as well!

samcv:
Yeah, I saw some alternatives but they had the issue that they would glob when *anything* had a `:` colon in it, so I played around with the commands and it seemed they recognized `:` specially as well, but having `./` or `/` overrode that. Wanted it to work in practically all conditions instead of just most all.

chaos95:
I just hit this very problem. After searching around for a better solution than having to remember `\scp` when I wanted globbing on local paths, I found this PR and realised that I last fetched prezto 3 days ago and must have _just_ missed out on this merge.

Anyway, long story short, thanks for this, it's a great fix :D

PR #1357, opened 2017-07-04
[editor] set forward/backward word for Ctrl+Right Ctrl+Left

samcv:
Set forward/backward word to be mapped the same as they are in vim,
instead of having weird functionality when unset.


belak:
With all these improvements to vi mode, I might have to start using it!

PR #1358, opened 2017-07-04
[editor] Add bindkey-all function to show all bound keys

samcv:
Function which allows you to see all the bound keys for all of
the different keymaps. It does accept arguments and passes them
through to bindkey -m "$keymap" so you could set a shortcut to
all keymaps as well if you wished.

Makes it much easier to see and grep what keys are bound to each of
the different keymaps.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -78,7 +78,15 @@ zle -N edit-command-line
 #
 # Functions
 #
-
+# Runs bindkey but for all of the keymaps. Running it with no arguments will
+# print out the mappings for all of the keymaps.
+function bindkey-all {
+  local keymap=''
+  for keymap in $(bindkey -l); do
+    [[ "$#" -eq 0 ]] && printf "#### %s\n" "${keymap}" 1>&2
+    bindkey -M "${keymap}" "$@"
+  done
+}
 # Exposes information about the Zsh Line Editor via the $editor_info associative
 # array.
 function editor-info {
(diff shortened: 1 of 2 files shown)


modules/editor/init.zsh
```diff
@@ -78,7 +78,15 @@ zle -N edit-command-line
 #
 # Functions
 #
-
+# Runs bindkey but for all of the keymaps. Running it with no arguments will
+# print out the mappings for all of the keymaps.
+function bindkey-all {
+  local keymap=''
+  for keymap in $(bindkey -l); do
+    printf "#### %s\n" "${keymap}" 1>&2
```
belak: Perhaps this should only be printed if there are no arguments?

samcv:
I was thinking of having `bindkey -q` be quiet and not print out the headings? Bindkey has no `-q` option. Or would that be too much?

```sh
function bindkey-all {
  local keymap=''
  local quiet=''
  if [[ "$1" == '-q' ]]; then quiet=1; shift; fi
  for keymap in $(bindkey -l); do
    [[ -z "$quiet" ]] && printf "#### %s\n" "${keymap}" 1>&2
    bindkey -M "${keymap}" "$@"
  done
}
```

modules/editor/init.zsh
samcv: That could work too :-) I like that idea.

samcv:
It has been updated to only print the keymap headings if there are no arguments.

PR #1359, opened 2017-07-04
[gitmodules] ignore dirty modules to make it easier for users

samcv:
When trying to add changes line by line by running `git add -p` the
user is prompted every time for each of the submodules, even if
there are no real changes to any of their directories.

Ignore them if dirty to be less annoying to people trying to add
changes by line.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,21 +1,28 @@
 [submodule "modules/autosuggestions/external"]
 	path = modules/autosuggestions/external
 	url = https://github.com/zsh-users/zsh-autosuggestions
+	ignore = dirty
 [submodule "modules/history-substring-search/external"]
 	path = modules/history-substring-search/external
 	url = https://github.com/zsh-users/zsh-history-substring-search.git
+	ignore = dirty
 [submodule "modules/syntax-highlighting/external"]
 	path = modules/syntax-highlighting/external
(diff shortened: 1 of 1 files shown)


.gitmodules
```diff
@@ -1,21 +1,28 @@
...
 	url = https://github.com/zsh-users/zsh-history-substring-search.git
+	ignore = dirty
 [submodule "modules/syntax-highlighting/external"]
 	path = modules/syntax-highlighting/external
 	url = https://github.com/zsh-users/zsh-syntax-highlighting.git
+	ignore = dirty
 [submodule "modules/completion/external"]
 	path = modules/completion/external
 	url = https://github.com/zsh-users/zsh-completions.git
+  ignore = dirty
```
belak: Indentation doesn't appear to match here

samcv:
Maybe it's just me but I seem to have files end up in all of these repositories, I'm not sure why they ended up there, but they cause them to be dirty since there are files that have not yet been added to the repository. Hmm.
```
$ cat .zprezto/modules/syntax-highlighting/external/titles
zsh-users/zsh-syntax-highlighting.git
```
I have files like this in all the folders but I'm not sure what created them.

samcv:
Looks like it was something I did that got those files there. Closing this PR since they have not reappeared.

PR #1360, opened 2017-07-04
Add zsh-help function for easily searching the zsh documentation

samcv:
Looks up things in the zsh documentation.
Usage: zsh-help [--all] search term(s)

Option --all will seach for the term anywhere, not just at the start of a
line. When not using --all it will search nicely for terms at the beginning
of the line, which in the zsh man pages is where terms that are explained
are located, allowing you to search the zsh man pages easily.

@belak do you think this is the best place to put this function?
Or should it go in another file. This function I have found myself
using all the time now that I have defined it, and I'm sure would be
very helpful for other people as well.

### DIFF
--- a/modules/utility/functions/zsh-help
+++ b/modules/utility/functions/zsh-help
@@ -1,0 +1,59 @@
+# Function to search zsh's man pages in a convenient way
+function zsh-help {
+    function zsh-help-join { # Joins the arguments into a string delimited by $separator
+        local separator=$1;
+        local arr=$*;
+        arr=${arr:${#separator}+1}; # < Line needed so result doesn't start with
+        arr=${arr// /$separator};   # a separator.
+        <<<$arr
+    }
+    local case='-i'
+    local section='zshall'
+    local debug=''
+    local pattern=''
+    # by default search only things at start of line
+    local prefix='^\s*'
+    if [[ $# < 1 || $1 == "--help" ]]; then
+        printf "Looks up things in the zsh documentation.\n"
+        printf "Press 'n' to search forward 'N' to search backwards for the term.\n"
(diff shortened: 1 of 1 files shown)


indrajitr:
Historically, these nifty functions have been going into `utility/functions` so that they become automagically available when 'utility' module is enabled.

What do you think?

NB: Also, a quick and simple completion helper would be cool :)

belak:
I think utility/functions would be a good place. Thanks for the suggestion!

I don't think a completion helper is needed, but it would be nice.

samcv:
I have moved the function into `utility/functions/zsh-help`. Is it acceptable for merge now?

modules/utility/functions/zsh-help
```diff
@@ -0,0 +1,59 @@
+# Function to search zsh's man pages in a convenient way
```
belak: The header format should be one of the following:

```
#
# Description
#
```

```
#
# Description
#
# Authors:
#   Author <authoremail>
#
```

modules/utility/functions/zsh-help
```diff
@@ -0,0 +1,59 @@
...
+function zsh-help {
```
belak: Because this will be an autoloaded function, the file should either call `zsh-help "$@"` at the end or comment out the function declaration (and closing brace).

See `modules/utility/functions/diff` for an example of the former and `modules/utility/functions/prep` for the latter.

I don't think we've decided on a specific format for prezto, so either should be fine for now.

samcv:
I followed how the other functions were defined with commented out function declaration and closing brace along with author/email and description using the convention. I made some improvements as well. Let me know if I need to change anything else.

PR #1361, opened 2017-07-04
[prompt/sorin] Fix branch code exploit vulnerability

samcv:
Escape all $ except the first $. Escape all backtick `'s. This prevents
variable names or shell expansions placed as branch names from
remotely exploiting code.

Fixes issue #1267 for sorin prompt.


belak:
Can we get a small comment added, saying why this exists? To someone who just jumps into this prompt and sees this section might be tempted to remove it because they don't understand it. Other than that, I guess this works, though the better fix would be to eventually switch to zsh-async so we don't need the temp file.

samcv:
Comment added!

belak:
Merged. Thanks for the fix!

PR #1364, opened 2017-07-06
Update functions to all use POSIX syntax

### LINKED ISSUE #692: Use the POSIX function syntax

belak:
I'm not a huge fan of this change, as I've made it known. However, this
is semi-recommended by the style guide we've chosen to use and is a more
common syntax.

I used the command Eriner originally posted and manually fixed some of the lines which were accidentally modified and tried to double check the rest.

Fixes #692.

CC @Eriner @sorin-ionescu

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -18,10 +18,10 @@ fi
 unset min_zsh_version
 
 # zprezto convenience updater
-# The function is surrounded by ( ) instead of { } so it starts in a subshell
+# The function is surrounded by ( ) instead of() { } so it starts in a subshell
 # and won't affect the environment of the calling shell
-function zprezto-update () (
-  function cannot-fast-forward {
+zprezto-update() (
+  cannot-fast-forward() {
     local STATUS="$1"
     [[ -n "${STATUS}" ]] && printf "%s\n" "${STATUS}"
     printf "Unable to fast-forward the changes. You can fix this by "
@@ -68,7 +68,7 @@ function zprezto-update () (
 #
 
 # Loads Prezto modules.
-function pmodload {
+pmodload() {
   local -a pmodules
   local pmodule
   local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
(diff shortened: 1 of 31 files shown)


belak:
This *will* be going in before any other PRs which change actual shell code so we don't need to spend another year waiting on someone to re-do these changes.

I've been using this for a day with no problems, so I'll merge this soon unless anyone has problems with it.

samcv:
I don't see the point of changing this? This is a zsh project and I don't see what this helps. If it were not a zsh project this could be an okay change. But in the end it really doesn't matter as long as there are no differences between defining functions the different ways...

samcv:
The style guide says this:
`Lower-case, with underscores to separate words. Separate libraries with ::. Parentheses are required after the function name. The keyword function is optional, but must be used consistently throughout a project.`

I personally like the function keyword as it makes it easier to grep for functions and easier to see where functions are. Though in the end it's really a stylistic choice.

belak:
I honestly prefer the `function x` syntax... but in the mentioned issue, Sorin and Eriner both weighed in, seemingly preferring the posix style syntax. It was me and one other person preferring the function keyword.

One thing worth noting: there are currently 3 function definition syntaxes being used in prezto:

```
function x
function x ()
x ()
```

I prefer the first, the middle should never be used and the third is posix. As long as we're consistent I'm fine with either the first or last syntaxes.

Anyway, there's a ton of discussion and quoting man pages and stuff in the issue. I can drop this if you'd like to comment there... I just want to be done with this issue.

samcv:
Also I'm most against this because:

Says bash on posix mode:
```
12. Function names must be valid shell names. That is, they may not contain characters other than letters, digits, and underscores, and may not start with a digit. Declaring a function with an invalid name causes a fatal syntax error in non-interactive shells.
```

We have plenty of functions which use dashes so we are already non-POSIX there. And there seems no real reason to change, though we should remove any `()` if we use the function keyword. Though this is all just my opinion and in the end it really doesn't matter that much...

belak:
I'm dropping this in favor of #1366

PR #1365, opened 2017-07-06
Add a .editorconfig file to setup tabstyle preferences

samcv:
With editors that support it it will setup preferences so that it
will use 2 spaces as the indent, except in .gitignore and .gitmodules
which in our files use tabs.

For more info see http://EditorConfig.org

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+[{.gitignore,.gitmodules}]
+indent_style = tab


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# Unix-style newlines with a newline ending every file
```
belak: I'm nit-picking at this point, but this comment probably isn't needed

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+[{.gitignore,.gitmodules}]
```
belak: Another nit: it would be nice to have a blank line before this section

indrajitr: In nit-picking race with @belak :)

Maybe:
```
[.git{attributes,ignore,modules}]
```

belak:
I'm going to merge this and make a few tweaks. Thanks for your contribution!

PR #1367, opened 2017-07-07
[git] Fix 'git-hub-shorten-url' helper

indrajitr:
## Proposed Changes
- Use git.io over 'https'
- Add extra check to conform to *.github.com URLs
- Use built-in _urls function for completion
- Update readme with GitHub blog URL


belak:
Thanks for your contribution! Looks good!

PR #1368, opened 2017-07-07
[editor] Fix undefined results with unbound keys in viins/vicmd

samcv:
Bind the remaining unbound keys in the main mode so that they
don't result in undefined functionality when pressed. (Often this
will change the casing of one/most/all of the characters you've
typed or even stranger things). In emacs mode this usually just
inserts a tilde, but this will fix that as well so nothing happens
when unbound keys are pressed.

Since there isn't any binding which does "nothing", create a noop
ZLE widget and bind them to that.

The user or other modules are still able to override these bindings
either by overriding the main keymap or setting a keymap in one
of the other modes which will take precedence over the main fallback.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -282,6 +282,32 @@ fi
 # Emacs and Vi Key Bindings
 #
 
+# Unbound keys in vicmd and viins mode will cause really odd things to happen
+# such as the casing of all the characters you have typed changing or other
+# undefined things. In emacs mode they just insert a tilde, but bind these keys
+# in the main keymap to a noop op so if there is no keybind in the users mode
+# it will fall back and do nothing.
+function _prezto-zle-noop {  ; }
+zle -N _prezto-zle-noop
+local unbound_keys=(
+  "${key_info[F1]}"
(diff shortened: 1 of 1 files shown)


modules/editor/init.zsh
```diff
@@ -278,6 +278,31 @@ else
...
+  "${key_info[F8]}"
+  "${key_info[F9]}"
+  "${key_info[F10]}"
+  "${key_info[F11]}"
+  "${key_info[F12]}"
+  "${key_info[PageUp]}"
+  "${key_info[PageDown]}"
+)
+for keymap in $unbound_keys; do
+  bindkey -M vicmd "${keymap}" vi-cmd-mode
```
belak: Could these be bound to `noop` rather than the vi-mode you're currently in, or is there different behavior when that's done?

samcv: Great suggestion! I made a noop ZLE widget since there was no `noop` and this is a better solution.

samcv:
Changed them to use `main` keymap so that we only have to bind them once and this will also fix problems users may get in every mode and solve the problem once and for all.

PR #1369, opened 2017-07-08
Prefer evaluation over test for arithmetic expression

indrajitr:
Usage of `(( ... ))` over `[[ ... ]]` is preferred for arithmetic expression since
the former is less error prone.

Also, unset local variable whenever possible.

Ref: [word of wisdom](https://github.com/sorin-ionescu/prezto/pull/1314#discussion_r115126654) :)


jeffcox:
Tested this out on macOS and FreeBSD, works great!

belak:
Thanks for your contribution!

PR #1371, opened 2017-07-08
[git] Support short-code and improve completion in 'git-hub-shorten-url'

indrajitr:
### Proposed Changes
- Add optional short-code support
- Improve completion for github.com URL (`http(s)://*.github.com` only)
- Return with non-zero exit code appropriately

_**Edit:** This is a quick followup to #1367, for the sake of completeness._


belak:
I'm not sure why I missed this until now. Thanks!

PR #1372, opened 2017-07-09
[completion] Minor cleanup of zstyle completion 'tag-order'

indrajitr:
### Proposed Changes
- De-duplicate common zstyle `tag-order` for ssh, scp and rsync.


belak:
Looks good to me. Thanks for your contribution! :)

Issue #205, opened 2012-06-19
Do not use rm's -i options by default


ghost:
I don't think that it is a good idea to [alias](https://github.com/sorin-ionescu/oh-my-zsh/blob/master/modules/utility/init.zsh#L53) `rm` to `rm -i` by default, because it is annoying and therefore users will just enter `y` without reading the name of the prompted file, or they will just start using the -f options. 
I personally think that this is very counterproductive.

herrbischoff:
Also, sorry to to necro-post but I just switched to prezto and this default is the single biggest annoyance I found with it. Also, due to its rather thin documentation, it took me a while to find this issue. Just wanted to comment on it.

sorin-ionescu:
I am inclined to keep it. I shot myself too many times in the foot. There is not a big protest against it.

herrbischoff:
Well, no problem. I managed to find it eventually. Apparently there are more people out there that need to be safeguarded from their own actions than I thought... ;P

mateogianolio:
I was also really annoyed that it's like this by default.

joaquinlpereyra:
+1 to disabling it by default or at least make it obviously configurable in the settings somewhere, as some plugins are in the .zprestorc file.

excetara2:
Y+1 disable or making obvious the default. I was confused by this for awhile as well. At least state it somewhere obvious because I couldn't figure out why it kept prompting me when typing rm without the -i.

sorin-ionescu:
It stays. Don't be [this guy](http://www.independent.co.uk/life-style/gadgets-and-tech/news/man-accidentally-deletes-his-entire-company-with-one-line-of-bad-code-a6984256.html).

joaquinlpereyra:
@sorin-ionescu There are saner settings available to avoid that. For example, warn if user tries to rm -rf any folder not in his home. Besides, this settings would do anything against rm -rf, the *force* would do it anyway. 

If you feel so strongly about it (which is a matter of personal taste, can't be discussed and, after all, you're the project's owner), I'd recommend at least documenting it clearly somewhere in the settings. Less intrusive things are stated there.

sorin-ionescu:
Asking for confirmation is good user experience.

mateogianolio:
Changing the default behaviour of a common bash function without mentioning
how to reverse it is bad user experience. However, it is of course entirely
your decision as project owner.

sorin-ionescu:
[...]

PR #1375, opened 2017-07-11
[utility] Add safe-ops zstyle option to allow disabling rm -i alias

samcv:
Currently rm is aliased to rm -i so that it will prompt before removing files.
Some people would not like this functionality and wish for an easy way to
disable it. This adds a new option:

zstyle ':prezto:module:utility' safe-ops yes/no

It is enabled by default even if zstyle is not set, but can be set to no
to disable rm from being aliased to rm -i.

This should resolve issue #205


belak:
Looks good. Thanks for your work!

PR #1376, opened 2017-07-12
[command-not-found] Load brew 'command-not-found' if homebrew tap is available

indrajitr:
Homebrew has _official_ tap that supports `command-not-found`. Enable it when possible.

### DIFF
--- a/modules/command-not-found/init.zsh
+++ b/modules/command-not-found/init.zsh
@@ -3,6 +3,7 @@
 #
 # Authors:
 #   Joseph Jon Booker <joe@neoturbine.net>
+#   Indrajit Raychaudhuri <irc+code@indrajit.com>
 #
 
 # Load command-not-found on Debian-based distributions.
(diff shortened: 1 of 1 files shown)


modules/command-not-found/init.zsh
```diff
@@ -11,6 +12,9 @@ if [[ -s '/etc/zsh_command_not_found' ]]; then
 # Load command-not-found on Arch Linux-based distributions.
 elif [[ -s '/usr/share/doc/pkgfile/command-not-found.zsh' ]]; then
   source '/usr/share/doc/pkgfile/command-not-found.zsh'
+# Load command-not-found on Mac OS X when homebrew tap is configured.
+elif (( $+commands[brew] )) && [[ "$(brew command command-not-found-init 2> /dev/null)" ]]; then
```
belak: This check should probably be:

```
elif (( $+commands[brew] )) && brew command command-not-found-init > /dev/null 2>&1; then
```

This is closer to the recommended command in the tap.

indrajitr: It might ever so slightly deviate from some places in prezto. But sure, let's follow what the upstream recommends.

belak:
Awesome! Thanks for your work on this!

Issue #1378, opened 2017-07-14
`rsync` and `scp` completion broken


indrajitr:
### Description

`rsync` and `scp` completion broke after #1356.

### Expected behavior

Tab completion should work for `rsync` and `scp`

### Actual behavior

Tab completion doesn't work anymore

### Versions

  - Prezto commit: a876890afd14304fee06dd87d7ba70412e4780d2
  - ZSH version: 5.3.1
  - OS information: macOS 10.12

belak:
This is the patch I'm playing with:

```
diff --git a/modules/utility/init.zsh b/modules/utility/init.zsh
index e9c20b2..1157bcb 100644
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -41,8 +41,8 @@ alias ftp='noglob ftp'
 alias history='noglob history'
 alias locate='noglob locate'
 alias rake='noglob rake'
-alias rsync='noglob rsync_scp_wrap rsync'
-alias scp='noglob rsync_scp_wrap scp'
+alias rsync='noglob rsync_wrap'
+alias scp='noglob scp_wrap'
 # This function wraps rsync and scp so that remote paths are not globbed
 # but local paths are globbed. This is because the programs have their own
 # globbing for remote paths. The wrap function globs args starting in / and ./
@@ -61,6 +61,15 @@ function rsync_scp_wrap {
   esac; done
   command $cmd "${(@)args}"
 }
+function rsync_wrap {
+  rsync_scp_wrap "rsync" "$@"
+}
+function scp_wrap {
+  rsync_scp_wrap "scp" "$@"
+}
+compdef -a _rsync rsync_wrap
+compdef -a _scp scp_wrap
+
 alias sftp='noglob sftp'

 # Define general aliases.
```

I'd love to see a better fix, so I'll ping a few people and see if any ideas come up.

belak:
Fixed in 90071d3ad062c4969b1d2e8512e9238524b90981. I'd like to get a better solution in (if it exists) but I wanted to get this in before I take off this weekend.

indrajitr:
I was actually planning to work on it tonight. But sure, good to have something out!

OkanEsen:
Unfortunately, the error still occurs for me and I think that this is related:

```
/Users/okan/.zprezto/modules/utility/init.zsh:70: command not found: compdef
/Users/okan/.zprezto/modules/utility/init.zsh:71: command not found: compdef
```

[...]

PR #1380, opened 2017-07-17
[utility] Refactor rsync_scp wrapper and fix completion

indrajitr:
Bit of refactoring and cleanup of `rsync_scp_wrap` and add proper completion support

### Changes
- Rename `rsync_scp_wrap` to `smartglob` and make it more generally usable. The rename is to signify that it can be generally used like any other built-in precommands (like `time`, `noglob`, `nocorrect` etc.)
- Tighten up internal variable usages within `rsync_scp_wrap`
- Update function documentation
- Enable generic completion support (and make `smartglob` usable with any other command)

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -57,8 +57,8 @@ Aliases
   - `history`
   - `locate`
   - `rake`
-  - `rsync`
-  - `scp`
+  - `rsync` (selectively enabled for local files)
+  - `scp` (selectively enabled for local files)
   - `sftp`
 
 ### General
@@ -139,6 +139,7 @@ Functions
   - `mkdcd` makes a directory and changes to it.
   - `popdls` pops an entry off the directory stack and lists its contents.
   - `pushdls` pushes an entry onto the directory stack and lists its contents.
+  - `noremoteglob` enable local path globbing but disable remote path globbing.
 
 ### Developer
 
--- a/modules/utility/functions/_noremoteglob
+++ b/modules/utility/functions/_noremoteglob
@@ -1,0 +1,11 @@
+#compdef noremoteglob
+#autoload
+
+#
+# Completes noremoteglob.
+#
+# Authors:
+#   Indrajit Raychaudhuri <irc+code@indrajit.com>
+#
+
+_precommand
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -41,33 +41,8 @@ alias ftp='noglob ftp'
 alias history='noglob history'
 alias locate='noglob locate'
 alias rake='noglob rake'
-alias rsync='noglob rsync_wrap'
-alias scp='noglob scp_wrap'
-# This function wraps rsync and scp so that remote paths are not globbed
-# but local paths are globbed. This is because the programs have their own
-# globbing for remote paths. The wrap function globs args starting in / and ./
-# and doesn't glob paths with : in it as these are interpreted as remote paths
-# by these programs unless the path starts with / or ./
-function rsync_scp_wrap {
-  local args=( )
-  local cmd="$1"
-  shift
-  local i
-  for i in "$@"; do case $i in
-    ( ./* ) args+=( ${~i} ) ;; # glob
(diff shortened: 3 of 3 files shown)


indrajitr:
@belak, @samcv: Let's discuss couple of things before merging this:
* As you can see, the intent is to make the wrapper more widely usable beyond just `rsync` and `scp`. Does the name sound intuitive enough?
* I couldn't figure out the examples of these cases:
    ```sh
    ( ./* ) argo+=( ${~arg} ) ;; # glob
    (  /* ) argo+=( ${~arg} ) ;; # glob
    ```
  Would be useful to have examples added for these. The doc needs updated accordingly.
* Adjusted the wordings of the documentation a bit, let me know what you think.

samcv:
Those comments just make it more clear which things get globbed and which don't. So ./ paths and / paths get globbed. Makes it easy for a person to read the switch even if they may not know that `${~foo}` means that the variable gets globbed.

belak:
* I like the idea, though I'm not sure smartglob is the right name because this is limited to a specific use case.
* These attempt to match things we know are local paths so we can glob those before falling back to noglob on the arguments containing a : (which *might* be remote).
* The current wording makes it sound like rsync and scp might only work for local files which isn't really what this does.

indrajitr:
* Yes, I am on the fence regarding the name as well. Ideas?
* Regarding the match, what I am wondering is, are we missing anything if we replace this
    ```sh
    ( ./* ) argo+=( ${~arg} ) ;; # glob
    (  /* ) argo+=( ${~arg} ) ;; # glob
    ( *:* ) argo+=( ${arg}  ) ;; # remote, noglob
    (  *  ) argo+=( ${~arg} ) ;; # glob
    ```
    with just this?
    ```sh
    ( *:* ) argo+=( ${arg}  ) ;; # noglob remote
    (  *  ) argo+=( ${~arg} ) ;; # glob everything else
    ```
   Local paths might not necessarily have `/` afterall (for example: `scp *.md  user@localhost:foo/`). So where all would we need explicit special case-ing for `./*` and `/*`?
* Sure, the wordings need to get better. Let me think about it. Suggestions?

belak:
* Perhaps localglob, or something like that? Though I hesitate to fully recommend that because people may confuse it with local variables.
* Those are matching strings which start with / or ./ which should always be local, not strings which contain them. So, something would be lost.
* Because this makes rsync a bit saner and the changes are mostly under the alias section (which is mostly unrelated to this change) I'd vote for not changing the docs or putting it separate from the aliases.

samcv:
@indrajitr the specialcasing for `./` and `/` are so that they get choosen first before the `:` case. If they were afterward then `./:foo` would not go through globbing. Also note that rsync and scp will not accept files that start with a `:` unless there is a `/` or a `./` in front of it. So at the moment it may glob incorrectly if you did `scp foo:* user@host:foo/`, but it handles at least using `./` or `/` and the filename having a `:` in it. Also the default *is* glob, so `scp *md` works perfectly fine. It doesn't handle if you did `scp fo:o*md` though and wouldn't glob that. Let me know if you think of a *safe* way to do this.

indrajitr:
@samcv, thanks a lot for the clarification. Indeed, your option is way better. I have adjusted the documentation capturing your explanation. Let me know if you want to have that tweaked a bit more.

@belak, response to your comments:

> Perhaps localglob, or something like that? Though I hesitate to fully recommend that because people may confuse it with local variables.

Maybe `lpathglob`, `remotenoglob` or something?

> Because this makes rsync a bit saner and the changes are mostly under the alias section (which is mostly unrelated to this change) I'd vote for not changing the docs or putting it separate from the aliases.
> ... the diff was really misleading and I missed that it was under the noglob heading.

I am guessing you are okay having this moved to function section. The reason for the relocation was this function would have to be made visible anyway (because of the way it is being used). We might as well go ahead and make an attempt to document it and formalize its usage.

belak:
How does `noremoteglob` sound?

indrajitr:
> How does `noremoteglob` sound?
Yes, much better. Updating PR.

indrajitr:
Fixes #1378

PR #1383, opened 2017-07-17
archive: fix unrar-free and add support for unar

corna:


modules/archive/functions/lsarchive
```diff
@@ -43,9 +43,11 @@ while (( $# > 0 )); do
       || lzcat "$1" | tar x${verbose:+v}f - ;;
     (*.tar) tar t${verbose:+v}f "$1" ;;
     (*.zip|*.jar) unzip -l${verbose:+v} "$1" ;;
-    (*.rar) unrar &> /dev/null \
+    (*.rar) unrar -V &> /dev/null \
```
belak: It would probably be better to use a check like `(( ${+commands[unrar]} ))` here and for the other commands.

corna:
Done

belak:
I'll try to take a look at this again soon. Thanks!

The only thing I saw at a glance is that the first option for unrar isn't surrounded by `()` so it's not as easy to follow as I'd like, but it's not a deal breaker.

belak:
Could you add a note that the `archive` function won't work for rar files if you only have `lsar` and `unar` (and don't have any of the other alternatives).

Are there any other filetypes this would work with, since as far as I can tell these commands are from the Unarchiver (maybe not something for this PR, but it may be worth exploring in the future).

corna:
Done.

Yes, The Unarchiver can be also used for other files.

belak:
Looks good! Thanks!

PR #1388, opened 2017-07-20
make the pydf alias optional

gmmeyer:
pydf is not a true replacement for df, as it cannot stat remote disks.
This alias being a default is inappropriate, especially without making it very clear in the documentation.

pydf should not stat remote disks because it is a single threaded application, and stating a remote disk can hang. However, it's simply disabled (and cannot be enabled) on pydf. While the color output is nice, it's not a simple drop in replacement the way that htop is for top.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -175,9 +175,13 @@ elif (( $+commands[wget] )); then
   alias get='wget --continue --progress=bar --timestamping'
 fi
 
-# Resource Usage
-if (( $+commands[pydf] )); then
-  alias df=pydf
+if zstyle -s ':prezto:module:utility' pydf 'yes'; then
+  # Resource Usage
+  if (( $+commands[pydf] )); then
+    alias df=pydf
+  else
+    alias df='df -kh'
+  fi
 else
   alias df='df -kh'
 fi


modules/utility/init.zsh
```diff
@@ -175,9 +175,13 @@ elif (( $+commands[wget] )); then
   alias get='wget --continue --progress=bar --timestamping'
 fi
 
-# Resource Usage
-if (( $+commands[pydf] )); then
-  alias df=pydf
+if zstyle -s ':prezto:module:utility' pydf 'yes'; then
```
indrajitr: * Usually, we prefer `zstyle -t` or `zstyle -T` for boolean
* Would you consider `use-pydf` instead of `pydf`?
* We'll also need doc for this in `utility/README.md` and in default `runcoms/depreztorc` with the default commented out.

modules/utility/init.zsh
```diff
@@ -175,9 +175,13 @@ elif (( $+commands[wget] )); then
...
+  # Resource Usage
+  if (( $+commands[pydf] )); then
+    alias df=pydf
+  else
+    alias df='df -kh'
+  fi
```
indrajitr: Possible to combine to two `if` conditions into a _single_ one? It might avoid unnecessary nested `if` and duplicate `alias df='df -kh'` in multple `else` blocks.

modules/utility/init.zsh
belak: Alternatively we could use `if zstyle -t ':prezto:module:utility:df' color` which lines up with how it's handled in `diff`

gmmeyer: I think the `if zstyle -t ':prezto:module:utility:df' color` is the same as the original error, it doesn't make explicit what the option is doing. pydf is a partial reimplementation of df in python that has colors rather than a wrapper that just colors the output.

modules/utility/init.zsh
gmmeyer: Yes, I will use an `||`. Good spot!

modules/utility/init.zsh
gmmeyer: (`colordiff` is a wrapper for `diff`)

belak: I'm tempted to take the sledgehammer approach and just say remove pydf in general. We killed branches in `diff` which fell back to commands which didn't support the same options... maybe it would be worth doing the same here.

belak:
I've removed support for pydf as the command line options don't line up with df and for such an common/important tool, I'd rather avoid the confusion. Thanks to everyone for weighing in!

PR #1389, opened 2017-07-21
[general] Switch code block formatting to use fence formatting

indrajitr:
### Changes:
* Indented code block doesn't support syntax highlighting, use fenced
formatting (```) instead for better syntax highlighting
* Wrap commands/functions in backticks
* Typo fixes


indrajitr:
@belak, thoughts?

PR #1391, opened 2017-07-21
[utility] Simplify enabling 'diff' color conditions in wrapper function

indrajitr:
Nested `if` can be removed for simple cases like these.

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -144,7 +144,7 @@ Functions
 
 ### Developer
 
-  - `diff` highlights diff output (requires `colordiff` or `Git`).
+  - `diff` highlights diff output (requires `colordiff`).
   - `make` highlights make output (requires `colormake`).
   - `wdiff` highlights wdiff output (requires `wdiff `or `Git`).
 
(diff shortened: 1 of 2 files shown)


belak:
This was originally left in because there were multiple options for getting the colored diffs. Would it make more sense to make it a simple alias, similar to df?

indrajitr:
That's a good point, we can make it an alias. However, we'll lose completion for `colordiff` (unless we have a wrapper completion function `_colordiff`).

FWIW, As we saw [elsewhere](https://github.com/sorin-ionescu/prezto/issues/1378#issuecomment-315507030), we cannot simply use `compdef _diff colordiff` in `utility` module because `compinit` happens later in `completion` module which in turn is loaded _after_ `utility` module.

indrajitr:
Let me think of other ways.

* For one we can probably use legacy `compctl`, but it might not be ideal for prezto (as we always use the new completion style). Also, `_foo` will win over `compctl` style by default anyway.
* Another option could be introduce an array of _`helper`_ functions that can be tried based of user  configurable `zstyle` (say `zstyle ':prezto:module:utility:diff' helper foo bar`).

belak:
Not sure why I delayed merging this... thanks for submitting it!

PR #1392, opened 2017-07-21
[utility] Reuse interactive aliases when 'safe-ops' is set

indrajitr:
When `safe-ops` is set, we can reuse the aliases that are already available.


belak:
Makes sense. Nice small change. :)

PR #1393, opened 2017-07-22
ssh: update README to reflect macOS changes

slarew:
## Proposed Changes
  - Remove statement in README of SSH module that states the module won't load on macOS.

### DIFF
--- a/modules/ssh/README.md
+++ b/modules/ssh/README.md
@@ -3,10 +3,6 @@ SSH
 
 Provides for an easier use of [SSH][1] by setting up [ssh-agent][2].
 
-This module is disabled on Mac OS X due to custom Apple SSH support rendering it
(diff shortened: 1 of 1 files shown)


indrajitr:
This seems to be related to #1394. @slarew, would you consider merging the two PR's into a single one?
We haven't done this consistently so far, but it might be useful to have README (or related doc update) together with the actual code change.

belak:
Normally when we do doc updates later, it's because we missed the doc update. It would be good to see these in the same PR if they're related.

slarew:
This commit is partially related to #1394, but it's actually more of a fix for 658fffb3a67a1f951d9c413802844694359caa7e when the README wasn't updated.

belak:
That makes sense. Thanks for the clarification!

PR #1394, opened 2017-07-22
ssh: remove persistent socket; use macOS Keychain

slarew:
## Proposed Changes

* Remove persistent SSH auth socket.
* Make ps|grep more robust and POSIX compliant.
* On macOS, use `-A` switch to "add identities to the agent using any
passphrase stored in the user's keychain."


belak:
Looks like there's a conflict now. Would you mind fixing that?

slarew:
I think I fixed the conflict.

belak:
Does this still work on linux and does it connect to the same agent between different shell sessions? I'm not very familiar with this, unfortunately.

slarew:
I've been running this on several Ubuntu 16.04 machines. Connects to same ssh-agent. Works in tmux.

jeffwidman:
@indrajitr / @johnpneumann if either of you get a chance, can you take a look? I am in the same boat as @belak, that it looks good but I'm not very familiar with this and there's a number of changes in here. 

Alternatively, the consequences of this breaking probably aren't the end of the world as it's unlikely to introduce a security risk, just a breakage risk so we'd hear about it immediately and fix it, so personally I'm fine with merging. Will give the other folks a day or two to weigh in though...

johnpneumann:
I'll take a look tonight

crivotz:
I tested the modification on Debian v9.1 - zsh 5.3.1 and found no problems.

modules/ssh/init.zsh
```diff
@@ -16,36 +16,32 @@ _ssh_dir="$HOME/.ssh"
...
-  export SSH_AUTH_SOCK="$_ssh_agent_sock"
-fi
-
 # Load identities.
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
   # Check for Linux system and ssh-askpass presence
   if [[ "$OSTYPE" == linux* ]] && [[ ! -a /usr/lib/ssh/x11-ssh-askpass ]]; then
     ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}"  2> /dev/null
+  elif [[ "$OSTYPE" == darwin* ]]; then
```
johnpneumann: The problem I have with this is what if the user doesn't put their keys into the keychain and simply define them within the identities? What happens is that the identities that are defined are ignored entirely, so they're now being forced to put their keys into their keychain.

slarew: I believe I have addressed your concern with 8430fb8.

slarew: @johnpneumann Could you review my changes in 8430fb8 to see if that fixes your concerns?

PR #1395, opened 2017-07-22
ssh: add trust checks before sourcing env vars

slarew:
## Proposed Changes

* Add ownership test before attempting to source environment variables in SSH module.  If the file is not owned by the current user, its contents should not be trusted.  The default path is predictable and thus subject to abuse.


modules/ssh/init.zsh
```diff
@@ -16,6 +16,19 @@ _ssh_dir="$HOME/.ssh"
 # Set the path to the environment file if not set by another module.
 _ssh_agent_env="${_ssh_agent_env:-${TMPDIR:-/tmp}/ssh-agent.env.$UID}"
 
+# Due to the predictability of the env file, check the env file exists and is
+# owned by current EUID before trusting it.
+if [ -f "$_ssh_agent_env" -a ! -O "$_ssh_agent_env" ]; then
```
belak: Please use `[[` rather than `[`

slarew: fixed

Shea690901:
Since commit ff91c8d4 unnecessary.
From this commit on, the env-file is stored within the `XDG_CACHE_HOME` directory (alternative in `${HOME}/.cache`) and as such isn't be stored in a world writeable location (at least I hope nobody has **that** location writable for anybody except oneself).
The same goes for the `gpg-agent.env` which, depending of the configuration, might be used instead of the ssh one.

PR #1396, opened 2017-07-22
[python] Detect actual `pip` command available for generating completion cache

indrajitr:
We cannot always assume availability of `pip`, it can be `pip2` or `pip3`
instead. We detect the first available one and use it for generating the
completion cache.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -106,16 +106,21 @@ if zstyle -T ':prezto:module:python' skip-virtualenvwrapper-init; then
 fi
 
 # Load PIP completion.
-if (( $+commands[pip] )); then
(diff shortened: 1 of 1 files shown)


belak:
I feel like we should prefer pip3 over pip2 if possible

indrajitr:
Currently, completion stub generated by `pip3` is identical to the one generated by `pip2` as would be evident from:
```console
$ diff -s <(pip2 completion --zsh) <(pip3 completion --zsh)
Files /dev/fd/11 and /dev/fd/12 are identical
```
And given that in most cases `pip` actually points to `pip2` out of the box (although changing fast), I just left it _as is_ with the conservative default. Of course, if `pip` points to `pip3`, that would always win.

PR #1400, opened 2017-07-22
[java] Add basic support for java module

indrajitr:
WIP


belak:
What are your plans for this? If you don't have the time to finish it up, maybe it would be worthwhile putting in contrib for now?

PR #1402, opened 2017-07-24
Fix unbound keys in vicmd/viins mode

samcv:
Previously I made a change to try and bind these unbound keys to the main
keymap in thoughts that in vicmd or viins mode it would fallback if it wasn't
bound in vicmd/viins mode. This turned out not to be the case.

Explicitly bind the keys in both viins and vicmd mode as works properly.


belak:
Thanks for the fix!

PR #1403, opened 2017-07-25
[python] Improve and document `virtualenvwrapper` initialization flow

indrajitr:
### Changes:
* Simplify zstyle name `skip-virtualenvwrapper-init` to `initialize`
  avoiding double negation in name
* Always perform `eval (pyenv virtualenv-init -)` at initialization
* Prefer `virtualenvwrapper_lazy` over `virtualenvwrapper` when available
* Honor `VIRTUALENVWRAPPER_VIRTUALENV` if it is defined.
* Document about `VIRTUALENVWRAPPER_PYTHON` and `VIRTUALENVWRAPPER_VIRTUALENV`
  (this would be particularly important in macOS after recent homebrew update)
* Add additional documentation for `initialize` in _README.md_ and _runcoms/zpreztorc_
* Add aliases `py2`, `py3` as shortcut for `python2`, `python3` respectively

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -56,6 +56,20 @@ is used. Replace *Developer* with your projects directory.
 export PROJECT_HOME="$HOME/Developer"
 ```
 
+The variable `VIRTUALENVWRAPPER_PYTHON` tells virtualenvwrapper to use the
+specified full path of `python` interpreter overriding the `$PATH` search.
+
+```sh
+export VIRTUALENVWRAPPER_PYTHON=/usr/local/bin/python
+```
+
+The variable `VIRTUALENVWRAPPER_VIRTUALENV` tells virtualenvwrapper to use the
+specified full path of `virtualenv` binary overriding the `$PATH` search.
+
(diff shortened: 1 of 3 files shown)


indrajitr:
A quick note about the admittedly contentious change of the `zstyle` name `skip-virtualenvwrapper-init` to `initialize`:
We earlier had:
```
zstyle -T ':prezto:module:python' skip-virtualenvwrapper-init
```
This should probably be `-t` instead of `-T`. So given that we intend to toggle the meaning, and it isn't documented yet, we might want to take the opportunity to avoid confusing _double negation_ and use a different name.
In case this doesn't sound right, or in case the name `initialize` isn't apt, I am open to suggestion.

jeffwidman:
> This should probably be -t instead of -T. 

What is the difference?

> So given that we intend to toggle the meaning, and it isn't documented yet, we might want to take the opportunity to avoid confusing double negation and use a different name.

Should an issue be opened so this doesn't get forgotten?

PR #1406, opened 2017-07-26
[completion] Support ignoring '/etc/hosts' entries in host completion

### LINKED ISSUE #1387: Speeding up completion when /etc/hosts is big with many blacklisted domains
### Description

Speeding up completion when /etc/hosts is big.
I want to use https://github.com/StevenBlack/hosts for blacklisting many domains but this will make autocompletion slow.

https://github.com/imanel/hosts_updater/#slow-auto-complete
> As a lot of hosts are added to your system, autocomplete might slow down. To solve this in zsh prezto one could replace `cat /etc/hosts (|)(N)` with `sed /0\.0\.0\.0/d`. This prevents the domains to be loaded for completion.

This change could be made in https://github.com/sorin-ionescu/prezto/blob/6e1f4ddd19dca2a02a79658fc8b9f8d230b591d0/modules/completion/init.zsh#L108

I don't know if you would like to make this change globally for all users?
And if not maybe some configuration option would be nice.
Personally I don't want to have any changes in prezto itself, only in runcoms.

### Expected behavior

Do not autocomplete domains blacklisted in `/etc/hosts` (`0.0.0.0 domain.com`).

### Actual behavior

Autocompletion is working for all domains in `/etc/hosts`.

### Steps to Reproduce

1. sudo vim /etc/hosts
2. Add `0.0.0.0 domain.com`
3. Restart terminal
4. `ssh dom<Tab>`

### Versions

  - Prezto commit: 18c96095f6df204efc6b2c69a0b6733d6de27ea6
  - ZSH version: zsh 5.1.1 (x86_64-ubuntu-linux-gnu)
  - OS information:
           Kernel: 4.8.0-58-generic x86_64 (64 bit gcc: 5.4.0)
           Desktop: Cinnamon 3.2.7 (Gtk 3.18.9-1ubuntu3.3)
           Distro: Linux Mint 18.1 Serena

indrajitr:
Entries from static */etc/hosts* can now be ignored via 'zstyle' configuration in *zpreztorc*. Both IP address and corresponding hostname will be ignored during host completion. However, some of
the entries ignored from */etc/hosts* still might appear during completion because of their presence in *ssh* configuration or history).

Fixes #1387

### DIFF
--- a/modules/completion/README.md
+++ b/modules/completion/README.md
@@ -6,6 +6,23 @@ the [zsh-completions][1] project.
 
 This module must be loaded **after** the *utility* module.
 
+Settings
+--------
+
+### Ignore */etc/hosts* Entries
+
+To ignore certain entries from static */etc/hosts* for host completion, add the
+following lines in *zpreztorc* with the IP addresses of the hosts as they
+appear in */etc/hosts*. Both IP address and the corresponding hostname will be
+ignored during host completion. However, some of the entries ignored from
+*/etc/hosts* still might appear during completion because of their presence in
+*ssh* configuration or history).
+
+```sh
+zstyle ':prezto:module:completion:*:hosts' etc-host-ignores \
+  '0.0.0.0' '127.0.0.1'
+```
+
 Contributors
 ------------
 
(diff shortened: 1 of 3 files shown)


indrajitr:
@matrixik, would appreciate if you can try this out.

PR #1408, opened 2017-07-27
ssh: update README for ssh-askpass

crivotz:
## Proposed Changes
Added _Requirements_ section for `ssh-askpass` on Linux machine

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -39,10 +39,11 @@ fi
 # Load identities.
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
-  if (( ${#_ssh_identities} > 0 )); then
-    ssh-add "$_ssh_dir/${^_ssh_identities[@]}"  < /dev/null 2> /dev/null
+  # Check for Linux system and ssh-askpass presence
+  if [[ "$OSTYPE" == linux* ]] && [[ ! -a /usr/lib/ssh/x11-ssh-askpass ]]; then
+    ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}"  2> /dev/null
   else
-    ssh-add  < /dev/null 2> /dev/null
+    ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}"  < /dev/null 2> /dev/null
   fi
 fi
 


indrajitr:
Thanks @crivotz!

Also wondering if we should have _guards_ around implicit calls to `ssh-askpass` so that we don't rely on users to read the _README_ explicitly.

Historically, prezto hasn't documented explicit requirement for a feature to work. It is meant to _just work&trade;_ if the requirements are available.

NB: I am not against explicitly documenting _Requirements_, but there isn't any right now (so this might be a _first_). And we don't want to end up being sloppy over time and deviate from _just works&trade;_ principal by relying more on users having to read _README_ to get started.

crivotz:
Thanks @indrajitr 
I am agree with you. I thought a lot about the word 'requirements' and I looked in the other modules to find alternatives.
Without the package you have no error but the agent will not work properly and will continue to require. the passphrase.
I try to look deep in the ssh-askpass ;)

crivotz:
I hope to be on the right track
I cleaned the README and only for linux machines if askpass is not present I use the old way.

diraol:
Hi @crivotz ,

in order to reduce code lines and checks, I think that you can first create a "local variable" (`ssh_id_path`) for example, by checking if there is any `_ssh_identities` (if there isn't, this local variable could be just 'empty'.

Then you check the SO and check for ASKPASS and use this variable.

I'm not good at all with shell/zsh syntax, but it would be something like:

```shell
  if (( ${#_ssh_identities} > 0 )); then
    local_ssh_id = "$_ssh_dir/${^_ssh_identities[@]}"
  else
    local_ssh_id = ""
  fi
  # Check for Linux system and ssh-askpass presence
  if [[ $(uname -s) == Linux* ]] && [[ ! -a /etc/X11/app-defaults/SshAskpass ]]; then
    ssh-add "$local_ssh_id"  2> /dev/null
  else
    ssh-add "$local_ssh_id"  < /dev/null 2> /dev/null
  fi
```

Sorry if it does not make sense, just trying to help. ;)

modules/ssh/init.zsh
```diff
@@ -40,9 +40,19 @@ fi
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
   if (( ${#_ssh_identities} > 0 )); then
-    ssh-add "$_ssh_dir/${^_ssh_identities[@]}"  < /dev/null 2> /dev/null
+    # Check for Linux system and ssh-askpass presence
+    if [[ $(uname -s) == Linux* ]] && [[ ! -a /etc/X11/app-defaults/SshAskpass ]]; then
```
indrajitr: * In zsh world, we do `if [[ "$OSTYPE" == linux* ]]`
* `/etc/X11/app-defaults/SshAskpass` might be too distro-specific, could there be something more generic?

modules/ssh/init.zsh
```diff
@@ -40,9 +40,19 @@ fi
...
+      ssh-add "$_ssh_dir/${^_ssh_identities[@]}"  2> /dev/null
```
indrajitr: You can probably replace `if (( ${#_ssh_identities} > 0 ))` block by inlining the check with something like:
```
ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}"  < /dev/null 2> /dev/null
```
I haven't tested though :)

modules/ssh/init.zsh
crivotz: - Ok I fix with $OSTYPE
- I can check for `/usr/bin/ssh-askpass` (symbolic link to `/etc/alternatives/ssh-askpass`) or for `/usr/lib/ssh/x11-ssh-askpass`

modules/ssh/init.zsh
crivotz: Work fine ;)

crivotz:
Changes made

indrajitr:
Looks good, thanks! Would you please squash the changes into a single commit?

crivotz:
@indrajitr done

indrajitr:
> ssh: update README for ssh-askpass

The commit message might have to be adjusted a bit. For example, the first line mentions updating README, which isn't the case anymore :)

crivotz:
Sorry I left the standard message from the squash just for history.
I fix with a single and clear commit

indrajitr:
Thank you very much @crivotz for your contribution and for dealing with these review requests patiently :)

crivotz:
Thanks @indrajitr for your advice and for your help

PR #1411, opened 2017-07-30
Introduce the 'kubectl' module.

brunomcustodio:
This PR introduces a `kubectl` module which enables completion and provides a few aliases for most commonly used commands.

`README.md` is included. I intend to expand on the module whenever I have the time.

### DIFF
--- a/modules/kubectl/init.zsh
+++ b/modules/kubectl/init.zsh
@@ -1,0 +1,40 @@
+#
+# Provides 'kubectl' aliases and utiities.
+#
+# Authors:
+#   Bruno Miguel Custodio <brunomcustodio@gmail.com>
+#
+
+# Return if requirements are not found.
+if (( ! ${+commands[kubectl]} )); then
+  return 1
+fi
+
+# Enable completion for 'kubectl'.
+cache_file="${0:h}/cache.zsh"
+kubectl_command="${commands[kubectl]}"
+
+if [[ "${kubectl_command}" -nt "${cache_file}" || ! -s "${cache_file}" ]]; then
+  ${kubectl_command} completion zsh >! "${cache_file}" 2> /dev/null
+fi
+
+source "${cache_file}"
+unset cache_file kubectl_command
+
+#
+# Aliases
+#
+
+alias kc='kubectl'
+
+alias kca='kubectl apply'
+alias kcc='kubectl create'
+alias kcD='kubectl delete'
+alias kcd='kubectl describe'
+alias kce='kubectl exec'
+alias kcf='kubectl port-forward'
+alias kcg='kubectl get'
+alias kcl='kubectl logs'
+alias kcr='kubectl run'
+
+alias kclf='kubectl logs --follow'
--- a/modules/kubectl/README.md
+++ b/modules/kubectl/README.md
@@ -1,0 +1,24 @@
+# `kubectl` Aliases and Utilities
+
+Provides [`kubectl`][1] aliases and utilities.
+
+## Aliases
+
+- `kc` is an alias for `kubectl`
+- `kca` is an alias for `ubectl apply`
+- `kcc` is an alias for `kubectl create`
+- `kcD` is an alias for `kubectl delete`
+- `kcd` is an alias for `kubectl describe`
+- `kce` is an alias for `kubectl exec`
+- `kcf` is an alias for `kubectl port-forward`
+- `kcg` is an alias for `kubectl get`
+- `kcl` is an alias for `kubectl logs`
+- `kcr` is an alias for `kubectl run`
+- `kclf` is an alias for `kubectl logs --follow`.
+
+## Authors
+
+*The authors of this module should be contacted via the [issue tracker][2].*
+
+[1]: https://kubernetes.io/
+[2]: https://github.com/sorin-ionescu/prezto/issues


modules/kubectl/init.zsh
```diff
@@ -0,0 +1,32 @@
...
+#   Bruno Miguel Custodio <brunomcustodio@gmail.com>
+#
+
+# Return if requirements are not found.
+if (( ! $+commands[kubectl] )); then
+  return 1
+fi
+
+# Enabled completion for 'kubectl'.
+source <(kubectl completion zsh)
```
indrajitr: Couple of things:
* We'll need to source this conditionally _iff_ zsh completion for `kubectl` hasn't been loaded already. For example, in homebrew-ed kubernetes-cli, `_kubectl` is provided by default.
* In case, we really have to source this, we'll have to _cache_ this upfront so that we don't perform the costly operation each time zsh is loaded.

brunomcustodio:
@indrajitr thanks for taking the time 👍 I've amended the commit and a cache file is now used just like in `npm`, `python` and other modules.

belak:
I'm also a little hesitant to add a single character alias for a less common module like this... But that may be up for debate depending on what the other maintainers think

brunomcustodio:
@belak I wonder why do you consider this a _less common module_... Are there any criteria or usage stats that can distinguish a _less common module_ from a "more common" one? Also, last time I checked module loading is optional.

I was more than happy to contribute to `prezto`, but I wouldn't ever think contributions are "ranked". If you are hesitant to accept this seemingly minor contribution, I'll happily retire the PR and keep this in my fork only.

belak:
Sorry about the novel below. The short version is that I think a 2-character alias (like `kc` or `ku`) would be better, especially when trying to be as consistent as we can with the rest of the project.

I definitely think this module is useful. I believe someone submitted a PR a few weeks ago which just had the kubectl completion stuff, and I asked for a little bit more... this is exactly what I was looking for - the completion on top of aliases.

I think "common" may have been the wrong term. Maybe a better way of putting it would have been "language or framework-specific". `git` is one of the only modules we have which I think deserves the single-character alias it has... and that's because of how universal it is. I don't have any specific stats, but I'd like to avoid single character aliases for anything that's less universal than git (outside specific utilities which are added like `fasd` which was added to be a `jump` command and was given `j`). Docker is close, but even that uses `dk`.

The problem isn't the alias itself. It's that I'd like to avoid a "first-come first-served" mentality on single character aliases... we try to make sure all modules in prezto can work together and that includes conflicting aliases, so if we accept `k` as an alias for `kubectl` it means we can't use `k` as an alias anywhere else in prezto. As an example, this would have been problematic if we gave `p` to `perl` then someone added a python module and wanted the same alias (this didn't happen - they have `pl` and `py` respectively).

I'm sorry it's still a bit subjective, but I hope my reasoning is a little bit clearer now.

Would switching to something like `kc` as the main alias (and the base for the other aliases) be an option for you? It's still quite short, but allows prezto to have more options in the future. Most of the other language or framework specific modules define their base command as a 2 or 3 character alias, which would also apply to this module.

If you still have problems with this, I'd be happy to answer any other questions you've got. I do appreciate you sticking with it for this!

brunomcustodio:
@belak OK, that definitely sounds better now. 🙂  I get your point, and with `kc` we go then. I've amended the commit.

modules/kubectl/init.zsh
```diff
@@ -0,0 +1,40 @@
...
+# Return if requirements are not found.
+if (( ! ${+commands[kubectl]} )); then
+  return 1
+fi
+
+# Enable completion for 'kubectl'.
+cache_file="${0:h}/cache.zsh"
+kubectl_command="${commands[kubectl]}"
+
+if [[ "${kubectl_command}" -nt "${cache_file}" || ! -s "${cache_file}" ]]; then
```
indrajitr: What's the purpose of `"${kubectl_command}" -nt "${cache_file}"`? I saw this on node module as well, but couldn't recall why it is there. @belak any recollection?

brunomcustodio: @indrajitr if I understand correctly, this serves as a means to check whether the completion cache file is up-to-date (as different tool versions could generate different completion files).

brunomcustodio:
@indrajitr I'm no expert in `zsh`, to be honest. Can you please point out to me how to check for that?

johnpneumann:
I'm against this being merged. The aliases defined here can be added in via other means, without having to add in a new module. Create a `.zsh_aliases` directory and name it `10_kubectl_aliases.zsh` (10 assumes that you want other things loaded before it). Inside of your `.zshrc` after it says `# Customize to your needs`, add the following:

```sh
for zsh_config in $(HOME/.aliases/*.zsh) source ${zsh_config}
```

Most of the other modules within are generic and they serve common use cases for everyone. They are either based on things that exist within most systems (ruby, python, yum, dpkg, homebrew) or allow better history, nicer prompts, etc (fasd, history-substring-search, spectrum, utility). Docker is the outlier and my feelings on it existing within prezto are the same (ignoring dnf since it's been there for 2 years). It's not that I don't think that this (or the docker module) brings value. It does; It's just too specific of a use case for me and can be loaded via other means.

To answer the question about what the `cache_file` if statement is doing: @brunomcustodio what you found was correct. It determines, whether the file for kubectl is newer than the cache or if the cache file does not exist or is empty (zero size). If any of that is true, then it writes out the [completion file](https://kubernetes.io/docs/user-guide/kubectl/kubectl_completion/#completion)  to the cache. For `npm`, it makes sense to cache it because it's slow, however I'd be surprised to see that same issue here.

brunomcustodio:
Gosh, I didn't expect the entry barrier for a module to be this steep, especially when module loading is optional.

> Create a .zsh_aliases directory and name it 10_kubectl_aliases.zsh (10 assumes that you want other things loaded before it). Inside of your .zshrc after it says # Customize to your needs, add the following:

Isn't the whole purpose of OSS to _share_? I could of course keep this to myself in my fork and stay quiet, but I thought the whole idea behind OSSing something is contributing and accepting contributions from those who want to give-back. But if there's no interest in this I'll happily close the PR myself.

johnpneumann:
Yes the point of OSS is to share and we're happy to have contributions. However, expecting any pull request sent to simply be accepted without discussion of merit, doesn't happen; Not just here, but on any OSS project. If it did nothing would get done. 

To address your other points: While module loading is optional, additions to the code base require support, which requires time - which is the only resource we can never get back. Again, I said that **I** disagree with adding this. I am one person within this group and if there's enough consensus around allowing this in, then it'll go in. Other contributors may disagree with me and arguments over adding in features and having a barrier to entry has been had before:

https://github.com/prezto-inactive-community-fork/prezto/issues/44#issuecomment-288772976
https://github.com/prezto-inactive-community-fork/prezto/issues/44#issuecomment-285606886

belak:
I think my main argument against allowing all modules is in oh-my-zsh. Right now it has over 700 open PRs, many to fix bugs in existing modules and many to add new modules. Once a framework starts accepting everything, it gets very hard to maintain and make sure everything works together. I don't want to see prezto get to that point. 

That being said, I do think there's value here, and it's definitely worth starting a discussion about the module even if we don't eventually decide to merge it in.

However, it's a bit of a fine line too... Because it's very hard to gauge real interest, as people have a tendency to game the system when you try to quantify what qualifies as "enough people" who would use a module. This has happened before and it's made me very hesitant to try and come up with a method for handling it as my previous attempts have been unsuccessful.

Please bear in mind that we're somewhat new to maintaining this as a group as well, so we've got some stuff to figure out when maintainers disagree, as we're seeing here. Please bear with us if we're moving slower than you'd like. We're really just trying to make the best software we can.

beloso:
I find this interesting. How is the `haskell` module a common tool?

As far as I know haskell is not a widely used programming language (and yet I use it).

You guys should clarify your modules section, and what you want to see as PRs for modules.

There are quite some modules there that are not used except by specific people. Not all of us develop in ruby/python/ocaml/node like you assume. 

The point in accepting "all" modules, is to give them visibility. Otherwise newer users will not know that there is a Kubernets module available. 
The point to not accept "any" modules, is like it was said, to not exhaust maintainers with less common tools. But dare I ask why is (again) the haskell module not updated for three years? By your criteria it is a common tool. Or am I wrong?

I come here just to say that I use only the git module, and all the other ones are irrelevant to me, I am not slowed down by the other modules being there.

belak:
1. I agree we should try to come up with a formalized module acceptance process, but I don't know how to do that without opening us up for abuse (votes and such are easy to fake and/or get your friends to do). Anyway this is a discussion we're going to need to have sooner rather than later. I've opened #1424 and will update that with more info when I get to a real keyboard (currently on my phone).
2. JavaScript, Python and Ruby are both in the top 10 languages used on GitHub. I don't think it's hard to justify those modules existing. Haskell and ocaml were both added before there were any other maintainers so I'm not sure as to the justification there. Haskell is still in a number of top 20 lists so I can understand it, but ocaml is a bit obscure and probably would have been met with the same resistance as this PR if someone wanted to add it today.
3.  I think most modules are useful, but I don't think all of them should be in prezto core. The fact of the matter is that when a module goes into prezto, we are responsible for bugs which get filled and PRs which are opened which will make it harder for maintainers, even if we don't use it. I don't think that will be a huge issue with this module as it stands now, but if the module grows, I'm not sure. I know the intention is not to exhaust maintainers, but unfortunately the intention doesn't change what actually happens.
4. The Haskell module has not been updated in 3 years because this repo went through a period of inactivity for quite a while and just earlier this year got a few more maintainers added. Just as a counter-example, the history module hasn't been updated in 3 years and I don't think we should be removing that. 

Maybe cleaning up and removing old/unused modules would be worth while as well. I'm not sure.

Hope this clears up some of your questions. :)

beloso:
I will not say what you must or must not do. I also don't know.

There are a couple of things that come to mind. You can add modules, but you don't actually need to promise to support them. In fact like you said prezto went for quite a time without an active maintainer and it is still the most popular ZSH framework, IMO.

You can add a repo next to this one like `prezto-extras` with more modules other than the `core` ones. With instructions on how to use them.

Quite honestly I don't feel that programming languages should be `core` for prezto. Most surely iOS development is quite popular (I am an iOS developer), and I don't need an xcode/objective c/swift module.

Right next to this PR is a java module (#1400), java is quite popular I agree, but do you guys really need a module with 2 allias for java, and a few more for maven (you should consider `gradle` btw, it will replace maven in the next years)? Why not two modules? One for java and another for maven? Do you spend that much amount of time building java applications in the command line? I am really curious about this, I thought Java IDE's were quite advanced these days.

What I think should be core to prezto are modules that facilitate life in the command line. Be it alias or whatever else.

belak:
I like your idea of a `prezto-extra`/`prezto-contrib` repo with a lower barrier of entry. We can have less of a promise of support for those modules and have a much lower barrier of entry. Then we can focus on what we actually deem the "core" modules, once we figure out what that means.

I'm not sure if @sorin-ionescu would be willing to create a prezto-contrib repo or if we'd have to put it elsewhere... would other maintainers agree with that change?

johnpneumann:
I like the idea of a prezto-extras, but we can make it simple and simply provide a readme with links and instructions for adding them to their current setup. We can then find a consensus around what it is that we want to do with modules that won't be part of core. If we can get it into zsh-users I'd prefer that - just so that it's owned by an org instead of an individual. 

In regards to the other responses about the addition or certain modules over others: I'm not going to continue the conversation. 

We are here to make sure that this remains an open project with contributions where they make sense. We continue to grow and want to encourage users to provide value where they can, but we are not going to merge every pull request submitted. Since I've already provided quotes directly from sorin (which I agree with) and have been explicit about the reasoning, I'm fairly certain that the horse has been beaten to death. No, 3 line aliases should not be merged. No, things specific to particular use cases should not be merged. The ability to load modules outside of having them be a part of core exists. Create a repo with the module and we can add a link to the readme mentioned above and then it's available for all. If someone wants to submit a PR for the Readme as well, we'd be happy to review it and get that moving forward until we can decide on a long-term solution.

brunomcustodio:
Enough drama. I'll do as I was told and keep this to myself in my fork. Nonetheless, and as life should be simple, I'm leaving this here for anyone coming across this PR:

https://gist.github.com/brunomcustodio/1814ed7688765f9a21f8410569d76c4c

Thanks to everyone who contributed with their opinion.

belak:
I'm working getting the contrib repo started while waiting to hear back about where it will be. I've included this as a new module in the contrib repo.

https://github.com/belak/prezto-contrib/commit/a2c460868ae81f730aa616ec2c7a9fc7b9dc4c37

PR #1420, opened 2017-08-07
Update prompt_sorin to show python virtual env

jeeftor:
Hopefully his is a quick and simple change.  

Adds a simple `(virtualenv) ` in front of the git tag if there is a virtual env specified.

* I did not use async tasks because `python-info` appears to query the environment so this should be super fast

Fixes #

## Proposed Changes

  -
  -
  -


belak:
Nice small change, though if there isn't a virtualenv set, this will add an extra space at the beginning of the rprompt, which changes how many characters are allowed before the rprompt disappears.

Issue #1415, opened 2017-08-04
Extra else in secondary prompt


jgdye:
### Description
When showing the secondary prompt during an if, else, etc., there is always an extra else in the prompt

### Expected behavior

Only the commands actually entered should be listed

### Actual behavior

![image](https://user-images.githubusercontent.com/4614540/28977032-66126822-790e-11e7-87d5-a7db5901bc5c.png)

This happened after updating a rather old install (based on 4f19700) that didn't show this property, but also occured on a brand new install with a brand new user (Image is from new user)

### Steps to Reproduce

1. Clone repo and install
2. Input `"` (or other item that will create a secondary prompt and hit return
### Versions

  - Prezto commit: 7b5196c
  - ZSH version: 5.1.1 (Ubuntu) and  5.3.1 (Raspbian)
  - OS information: Ubuntu 16.04 and Raspbian 9

indrajitr:
1. What's the output of `print $PROMPT2`?
2. What prompt theme are you using?

I am not sure if we set secondary prompt in default theme viz., _prompt_sorin_theme_ (which would be the case with brand new setup). Is it possible that you have a different source (~/.zshrc, ~/.zprofile, /etc/zprofile etc.) which is bringing this in?

belak:
I just tried this on Ubuntu 16.04 with ZSH 5.1.1 and wasn't able to reproduce it. Do you have your dotfiles anywhere public so we can test this? Or is it just your prezto fork?

indrajitr:
@jgdye, just looked at your prezto fork.

1. Is it possible that you are using [joshua theme](https://github.com/jgdye/prezto/blob/master/runcoms/zpreztorc#L95)?
2. The theme seems to be [customizing `PS2`](https://github.com/jgdye/prezto/blob/8557ed5e99eeb739d9afdc14a42ba218642be811/modules/prompt/functions/prompt_joshua_setup#L140-L143).

I am closing this based on the assumption that you are using custom theme (joshua) which has `PS2` customization _outside_ what prezto provides. Please feel encouraged to re-open if you consider my assumption incorrect and/or you mange to find some other prezto issue.

jgdye:
[...]

PR #1421, opened 2017-08-07
Update prompt_sorin_setup (python support)

jeeftor:
Added python support

Fixes #
Deals with half of issue #924

## Proposed Changes

  - Add `python-info` into the `precmd` (does not require async because is looking at env vars 
  - Made the virtual env color orange
  
sorin2

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -131,7 +131,7 @@ function prompt_sorin_setup {
 
 
   # Set python-info format
-  zstyle ':prezto:module:python:info:virtualenv' format '%f%F{208}(%v)%F{7}'
+  zstyle ':prezto:module:python:info:virtualenv' format '%f%F{208}(%v)%F{7} '
   
   # Get the async worker set up
   async_start_worker async_sorin_git -n
@@ -143,7 +143,7 @@ function prompt_sorin_setup {
 
   # Define prompts.
   PROMPT='${SSH_TTY:+"%F{9}%n%f%F{7}@%f%F{3}%m%f "}%F{4}${_prompt_sorin_pwd}%(!. %B%F{1}#%f%b.)${editor_info[keymap]} '
-  RPROMPT='$python_info[virtualenv] ${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}${_prompt_devel_git}'
+  RPROMPT='$python_info[virtualenv]${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}${_prompt_devel_git}'
   SPROMPT='zsh: correct %F{1}%R%f to %F{2}%r%f [nyae]? '
 }
 
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -131,7 +131,7 @@ function prompt_sorin_setup {
 
 
   # Set python-info format
-  zstyle ':prezto:module:python:info:virtualenv' format '%f%F{208}(%v)%F{7} '
+  zstyle ':prezto:module:python:info:virtualenv' format '%f%F{3}(%v)%F{7} '
   
   # Get the async worker set up
   async_start_worker async_sorin_git -n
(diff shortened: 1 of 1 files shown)


belak:
Please avoid creating a new PR every time a change is made, though I do appreciate the additional information. You can push commits to your branch and they will update the PR without needing you to make a new one.

Anyway. this is a nice small change... though if there isn't a virtualenv set, this will add an extra space at the beginning of the rprompt, which changes how many characters are allowed before the rprompt disappears. Would it be possible to remedy that?

jeeftor:
So maybe i'm doing something wrong but I don't actually see the `()`'s when i deactivate

sorin3

Are you sure its an issue?  In a previous version I had a function that would handle what you are talking about

```
# Returns the Virtual Environment string if it exists
function check_python_virtual_environment() {
  if [[ -z "${VIRTUAL_ENV// }" ]]; then
    current_python_virtual_environment=
  else
    current_python_virtual_environment="(${VIRTUAL_ENV[(ws:/:)-1]})"
  fi
}
```

but I wasn't actually using `zstyle` and now that i switched i don't see that behavior any more.

jeeftor:
The only bug I do see is maybe i need to nix the space before `${editor_info` 
```
RPROMPT='$python_info[virtualenv] ${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}${_prompt_devel_git}'
```

I'll make that update - because in the case there is no pyenv i think $python_info is unset but i'm still adding a space after it so thats bad.

jeeftor:
So to summarize things thus far

* When a virtual env is set prompt will get `(virtualenv) ` prepended to the `RPROMPT`
* When the virtual env is not set the prompt will have nothing prepended
* The `() ` is set inside a `zstyle` command so there will be no empty `()`'s in the case of no prompt (as far as i know)

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -133,7 +143,7 @@ function prompt_sorin_setup {
 
   # Define prompts.
   PROMPT='${SSH_TTY:+"%F{9}%n%f%F{7}@%f%F{3}%m%f "}%F{4}${_prompt_sorin_pwd}%(!. %B%F{1}#%f%b.)${editor_info[keymap]} '
-  RPROMPT='${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}${_prompt_sorin_git}'
+  RPROMPT='$python_info[virtualenv]${editor_info[overwrite]}%(?:: %F{1}⏎%f)${VIM:+" %B%F{6}V%f%b"}${_prompt_devel_git}'
```
indrajitr: Where is `_prompt_devel_git` coming from? I am guessing that's a typo.

modules/prompt/functions/prompt_sorin_setup
```diff
@@ -123,6 +129,10 @@ function prompt_sorin_setup {
   zstyle ':prezto:module:git:info:keys' format \
     'status' '%b %p %c:%s%A%B%S%a%d%m%r%U%u'
 
+
+  # Set python-info format
+  zstyle ':prezto:module:python:info:virtualenv' format '%f%F{208}(%v)%F{7} '
```
indrajitr: We have to be careful with the colors that we use (particularly for default theme). Best to stick to the 16 terminal colors for maximum compatibility. We might have to have ellipsis for very long string (again, as safety for default theme).

jeeftor:
@indrajitr  I made the two changes you asked for.  Good finds.  Also moving forward I'm happy to work on a generalized env setup for the prompt.  If the python changes are acceptable - at least i'm hoping we can push those.  

(I actually wont be able to test the changes until tomorrow - because i'm limited to web browser access only - but I think from looking over it - it should work)

modules/prompt/functions/prompt_sorin_setup
jeeftor: Updated in last commit

modules/prompt/functions/prompt_sorin_setup
jeeftor: This was a type-o on my end. Fixed in last commit.

johnpneumann:
Would you mind taking one of the scripts (docker or vagrant, doesn't matter to me) from this comment https://github.com/sorin-ionescu/prezto/issues/1415#issuecomment-320886220 and post screen shots of some basic interactions from whichever one you choose. Some things I'd like to see:

- if/then/else
- nested if/then/else
- for/do/done
- nested for/do/done
- history search
- with python installed
- with python installed and a virtualenv activated
- without python installed

We've had a few issues with themes over the last few months, so I want to make sure that we aren't breaking existing functionality. Thanks for the help with this. The code looks good to me based on my quick glance through. Can dive deeper, but it looks like @indrajitr has already given some feedback, so I'll let him roll with it.

indrajitr:
Didn't go all the way into trying in a dockerized instance. But looks good so far.
Further enhancements can happen as followup.

Thanks @jeeftor for your first PR! ✨

PR #1432, opened 2017-08-17
Add thefuck module

lucianomagrao:


modules/thefuck/init.zsh
```diff
@@ -0,0 +1,24 @@
+#
+# Integrates TheFuck
+#
+# Author:
+#   Luciano Mores <leslau@gmail.com>
+#
+
+if [[ -z $commands[thefuck] ]]; then
```
belak: This should be `if (( $+commands[thefuck] )); then`

modules/thefuck/init.zsh
```diff
@@ -0,0 +1,24 @@
...
+
+fuck-command-line() {
+    local FUCK="$(THEFUCK_REQUIRE_CONFIRMATION=0 thefuck $(fc -ln -1 | tail -n 1) 2> /dev/null)"
+    [[ -z $FUCK ]] && echo -n -e "\a" && return
+    BUFFER=$FUCK
+    zle end-of-line
+}
+zle -N fuck-command-line
+# Defined shortcut keys: [Esc] [Esc]
+bindkey "\e\e" fuck-command-line
```
belak: So, this is the same binding which was requested for the sudo plugin? I guess that makes sense since it does pretty much the same thing with some additional tracking.

leslau: Yeah, it does the same thing with some plus :)

leslau:
Fell free to take the file to another repo if u dont want to merge this on core, i'm fine.

modules/thefuck/init.zsh
```diff
@@ -0,0 +1,24 @@
...
+#   Luciano Mores <leslau@gmail.com>
+#
+
+if (( ! $+commands[thefuck] )); then
+    echo 'thefuck is not installed, you should "pip install thefuck" first'
+    return 1
+fi
+
+# Register alias
+eval "$(thefuck --alias)"
```
sergiorussia: shouldn't it be `eval $(thefuck --alias)`? see https://github.com/nvbn/thefuck

leslau: it does the same thing

LeoJweda:
@lucianomagrao This has been inactive for a while. Is it OK with you if I create a PR in contrib repo?

lucianomagrao:
@LeoJWeda sure, go ahead and do it 😁

lucianomagrao:
> I think we'd like to keep this out of prezto-core (I'm trying to move as much functionality out of core as possible), but it would be a good candidate for the [contrib repo](https://github.com/belak/prezto-contrib). I'd be happy to accept it there if you want to resubmit it.

Added to contrib repo, close this please

PR #1433, opened 2017-08-17
[python] Edit pyenv virtualenv-init check to support non-brew installations

### LINKED ISSUE #1428: Python module failing to run pyenv virtualenv-init
### Description
It seems like the Python module is incorrectly using the `commands` array, by trying to find a command named `pyenv-virtualenv-init`, causing the module to not find and run `pyenv virtualenv-init`.
Line in question: https://github.com/sorin-ionescu/prezto/blob/master/modules/python/init.zsh#L96

But looking at the code history, the Python module has been using this since April of 2017 1050a0a290324f3b96256d3b1fefe4a84ab402d0, so I feel like I must be doing something wrong.

I have pyenv installed with the virtualenv plugin, but when I run `$+commands[pyenv-virtualenv-init]` in my shell I get `0`

### Expected behavior

virtualenv is loaded via pyenv

### Actual behavior

virtualenv is not loaded, and an error is printed on shell open

```
/home/zwhaley/.zprezto/modules/python/init.zsh:source:109: no such file or directory: 
```

### Steps to Reproduce

1. Enable Python module `zstyle ':prezto:module:python:virtualenv' auto-switch 'yes'`
2. Open shell

### Versions

  - Prezto commit: dbf96d2d3749b2905927a08c55a8c9eba12c8425
  - ZSH version: zsh 5.3.1 (x86_64-redhat-linux-gnu)
  - OS information: Fedora release 26 (Twenty Six) x86_64

zachwhaley:
Fixes #1428 

## Proposed Changes
If the pyenv virtualenv plugin is installed using the [pyenv-installer][1]
app script or directly via a [Git clone][2], then the pyenv-virtualenv-init
executable, that the `pyenv virtualenv-init` command uses, will not
exist in the user's PATH and therefore cannot be found using `$commands[pyenv-virtualenv-init]`.
Installing the pyenv-virtualenv plugin in this manner is common among Linux users.

Using the pyenv `commands` command, which lists all commands pyenv can
run, we can find if the virtualenv-init command is available to pyenv
and subsequently the virtualenv plugin, without relying on
pyenv-virtualenv-init to exist in the user's PATH.

[1]: https://github.com/pyenv/pyenv-installer
[2]: https://github.com/pyenv/pyenv-virtualenv#installing-as-a-pyenv-plugin

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -93,7 +93,9 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
   # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] && $+commands[pyenv-virtualenv-init] )); then
+  if (( $+commands[pyenv] )) && \
+     pyenv commands | command grep -q virtualenv-init
+  then
     eval "$(pyenv virtualenv-init -)"
     # Optionall activate 'virtualenvwrapper' with 'pyenv' is available.
     if (( $#commands[(i)pyenv-virtualenvwrapper(_lazy|)] )); then


modules/python/init.zsh
```diff
@@ -93,7 +93,9 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
   # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] && $+commands[pyenv-virtualenv-init] )); then
+  if (( $+commands[pyenv] )) && \
+     pyenv commands | command grep -q virtualenv-init
+  then
     eval "$(pyenv virtualenv-init -)"
     # Optionall activate 'virtualenvwrapper' with 'pyenv' is available.
     if (( $#commands[(i)pyenv-virtualenvwrapper(_lazy|)] )); then
```
diraol: Shouldn't the same type of check from line 97 be done here? (`pyenv-virtuanenvwrapper`)

zachwhaley: I believe so, but I don't have that plugin installed and I didn't want to over complicate this PR.

If you'd like me to look into this, I can install virtualenvwrapper and give it a go.

indrajitr: Yes, I think so. Essentially, we can assume pyenv plugins to be present in `$PATH` in all cases. We'll have to revisit the treatment for pyenv plugin detection in general.

zachwhaley: @indrajitr Sorry, I don't think I understood.  Yes, I should make the same change to pyenv-virtualenvwrapper?

And

> we can assume pyenv plugins to be present in `$PATH`

Did you mean _cannot_, because that is the reason for this change.

indrajitr: You got me :)

modules/python/init.zsh
```diff
@@ -93,7 +93,9 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
   # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] && $+commands[pyenv-virtualenv-init] )); then
+  if (( $+commands[pyenv] )) && \
+     pyenv commands | command grep -q virtualenv-init
+  then
     eval "$(pyenv virtualenv-init -)"
     # Optionall activate 'virtualenvwrapper' with 'pyenv' is available.
```
belak: Unrelated issue - this should be "Optionally"

modules/python/init.zsh
```diff
@@ -93,7 +93,9 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
   # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] && $+commands[pyenv-virtualenv-init] )); then
+  if (( $+commands[pyenv] )) && \
+     pyenv commands | command grep -q virtualenv-init
```
belak: This should probably be a check for `[[ -n $(pyenv commands | command grep -q virtualenv-init) ]]` (https://google.github.io/styleguide/shell.xml?showone=Testing_Strings#Testing_Strings) and the `then` should be on the same line as the conditional (https://google.github.io/styleguide/shell.xml?showone=Loops#Loops)

modules/python/init.zsh
indrajitr: And `s/with/when/` :)

modules/python/init.zsh
zachwhaley: The first line is not checking a string, but the return code of the command.

I'll fix the then conditional.

modules/python/init.zsh
zachwhaley: So after installing virtualenvwrapper and making those changes, I'm left wondering what all the conditionals on line [88](https://github.com/zachwhaley/prezto/blob/6ed549ba8f3595bc4b1ab39ae1aad925910c7c37/modules/python/init.zsh#L88) are testing for.  Escpecially `zstyle -T ':prezto:module:python:virtualenv' initialize`, which seems to be used to set whether the user wants to initialize virtualenvwrapper or not.

In my case, I would not, but I would want to enable virtualenv.

So I'm wondering if this block of code needs a little reworking to allow for virtualenv without initializing virtualenvwrapper.

modules/python/init.zsh
belak: You're right. My mistake. I wasn't paying attention to the -q... also pipelines in if statements usually trip me up. I tend to think it's the first command in the pipeline, but it's actually the last... one of the reasons I try to avoid pipes in conditionals if possible is that it's easy to mess that up.

modules/python/init.zsh
zachwhaley: I'm gonna walk myself back on this since it seems like a safe assumption that if you installed the virtualenvwrapper plugin, you probably want to use it.

indrajitr:
Personally, I am not a fan of `|` and `grep` noises since these can be mostly achieved directly using zsh.
That said, I am going ahead and merging the PR  because this is a decent attempt at fixing the initial problem (which is an obvious bug to me) with awesome analysis in commit log.

I'll make a followup PR little later with relevant tweaks.

gpanders:
One other change I'd like to suggest: the check for the `:prezto:module:python:virtualenv' initialize` zstyle seems to be wrong. Currently, it is using

`zstyle -T ':prezto:module:python:virtualenv' initialize`

This is returning `true` even if the relevant line in the `.zpreztorc` is commented out. It should be

`zstyle -t ':prezto:module:python:virtualenv' initialize 'yes'`

But this PR just got merged, so does this need to be a new PR now?

indrajitr:
And thanks @zachwhaley for the first PR! ✨

belak:
@gpanders It's a bit confusing, but the values in zpreztorc are supposed to be the defaults, even when commented out, so I believe this is expected behavior. Thanks for mentioning it though!

gpanders:
@belak Interesting. So where is that value set, then, if not in zpreztorc?

indrajitr:
As [inlined doc](https://github.com/sorin-ionescu/prezto/blob/cab4ac54ece92bb6793c255fe2dc1851586a916e/modules/python/init.zsh#L85-L86) says, it defaults to `yes` and thus enabled.

So to disable, you'll have to explicitly specify `no` in _~/.zpreztorc_:
```
zstyle ':prezto:module:python:virtualenv' initialize 'no'
```
This has been done to stay compatible with previous behavior (when it wasn't backed by a `zstyle` setting).

PR #1414, opened 2017-08-04
[python] Detect availability of virtualenv pyenv plugins as well as pyenv

indrajitr:
Just having 'pyenv' available doesn't imply availability of virtualenv pyenv plugins. Check for availability of virtualenv plugins as well before attempting to use pyenv wrapper for virtualenv.

Fixes #1413

PR #1434, opened 2017-08-18
[python] Refine pyenv based virtualenv management

indrajitr:
### Proposed Changes

Now that pyenv plugins availability is detected by directly probing
'pyenv', we need to rely on the same mechanism consistently.

Further, we perform available plugin scan with zsh native techniques
instead of relying on external commands.

Followup to #1433

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -92,18 +92,25 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
   # Disable the virtualenv prompt.
   VIRTUAL_ENV_DISABLE_PROMPT=1
 
-  # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] )) && \
-     pyenv commands | command grep -q virtualenv-init
-  then
+  # Create a sorted array of available virtualenv related 'pyenv' commands to
+  # look for plugins of interest. Scanning shell '$path' isn't enough as they
+  # can exist in 'pyenv' synthesized paths (e.g., '~/.pyenv/plugins') instead.
+  local -a pyenv_plugins
+  if (( $+commands[pyenv] )); then
+    pyenv_plugins=(${(@oM)${(f)"$(pyenv commands --no-sh 2>/dev/null)"}:#virtualenv*})
+  fi
+
+  if (( $pyenv_plugins[(i)virtualenv-init] <= $#pyenv_plugins )); then
+    # Enable 'virtualenv' with 'pyenv'.
     eval "$(pyenv virtualenv-init -)"
-    # Optionall activate 'virtualenvwrapper' with 'pyenv' is available.
-    if (( $#commands[(i)pyenv-virtualenvwrapper(_lazy|)] )); then
-      pyenv "${${(@O)commands[(I)pyenv-virtualenvwrapper(_lazy|)]}[1]#pyenv-}"
+
+    # Optionally activate 'virtualenvwrapper' plugin when available.
+    if (( $pyenv_plugins[(i)virtualenvwrapper(_lazy|)] <= $#pyenv_plugins )); then
+      pyenv "$pyenv_plugins[(R)virtualenvwrapper(_lazy|)]"
     fi
   else
-    # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper in '$path'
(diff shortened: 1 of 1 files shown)


belak:
Quality as always. Looks good to me at a glance @indajitr.

modules/python/init.zsh
```diff
@@ -92,22 +92,31 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
...
+    unset virtenv_plugins
   else
     # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper in '$path'
     # and other known locations on a Debian based system.
     virtenv_sources=(
       ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
       /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
     )
+
     source "${virtenv_sources[1]}"
```
zachwhaley: FYI, this gets called whether this virtualenvwrapper file exists or not. 

So if you have pyenv installed, but not the virtualenv plugin or the virtualenvwrapper plugin, then you get a"file not found" error at login

modules/python/init.zsh
```diff
@@ -92,22 +92,31 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
...
 
-  # Enable 'virtualenv' with 'pyenv'.
-  if (( $+commands[pyenv] )) && \
-     pyenv commands | command grep -q virtualenv-init
-  then
-    eval "$(pyenv virtualenv-init -)"
-    # Optionall activate 'virtualenvwrapper' with 'pyenv' is available.
-    if (( $#commands[(i)pyenv-virtualenvwrapper(_lazy|)] )); then
-      pyenv "${${(@O)commands[(I)pyenv-virtualenvwrapper(_lazy|)]}[1]#pyenv-}"
+  if (( $+commands[pyenv] )); then
```
zachwhaley: I think this might be a regression for @diraol, since this might revert the changes in #1414

modules/python/init.zsh
```diff
@@ -92,22 +92,31 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
...
+    # Grab sorted array of available 'pyenv' plugins by probing 'pyenv'
```
zachwhaley: I'm not sure this comment matches the code pre se.  It looks like the code is returning all pyenv commands that start with "virtualenv", and not all pyenv plugins, correct?

modules/python/init.zsh
```diff
@@ -92,18 +92,26 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
...
+  # Create a sorted array of available virtualenv related 'pyenv' commands to
+  # look for plugins of interest. Scanning shell '$path' isn't enough as they
+  # can exist in 'pyenv' synthesized paths (e.g., '~/.pyenv/plugins') instead.
+  local -a virtenv_plugins
+  if (( $+commands[pyenv] )); then
+    virtenv_plugins=(${(@oM)${(f)"$(pyenv commands --no-sh 2>/dev/null)"}:#virtualenv*})
+  fi
+
+  if (( $+commands[pyenv] && \
+        $virtenv_plugins[(i)virtualenv-init] <= $#virtenv_plugins )); then
```
diraol: @indrajitr if `virtenv_plugins` is "empty", this expression (`$virtenv_plugins[(i)virtualenv-init] <= $#virtenv_plugins`) returns False?

I'm asking it because if there is no `pyenv` installed, then `virtenv_plugins` is at its initial state (`local -a virtenv_plugins`). Then, may be we do not need to re-check for `$+commands[pyenv]` here right?

modules/python/init.zsh
indrajitr: Good find, thanks! Applied the fix separately.

modules/python/init.zsh
indrajitr: Right you are! Clearly, I wasn't in right mind. Thanks.

modules/python/init.zsh
indrajitr: Reworded, does the new one make sense?

indrajitr:
Thanks for the review guys!

PR #1436, opened 2017-08-19
Drop `easy_install`, everyone now uses `pip`

jeffwidman:
Everyone now uses `pip`, drop reference to `easy_install`

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -28,8 +28,7 @@ documentation.
 
 ### Usage
 
-Install packages into the per user site directory with `easy_install --user` or
-`pip install --user`.
+Install packages into the per user site directory with `pip install --user`.
 
 virtualenvwrapper
 -----------------


indrajitr:
What do we gain by dropping reference to something that still works? Most tools still allow `easy_install` based installation as documented method, don't they?

johnpneumann:
So the other reason to keep it around is because pip cannot install eggs, so if you're working with legacy systems or have a specific requirement to install from an egg, you'd still need it. I would say that pip should be featured first, with an additional note that easy_install is available. It's not anything I'm passionate about though, so I'm fine with this. Within the confines of folks I know, everyone utilizes pip and the only reason anyone used easy_install was to install pip (before get_pip.py was a thing).

jeffwidman:
Keep in mind that this is purely a docs change, not a functional change. 

I work primarily in the python ecosystem, and pip has become the _blessed_ package manager in the community. No one I know uses `easy_install` for anything other than when they need to bootstrap `pip`. 

Anyone who actually needs `easy_install` for things like installing eggs will likely be aware of it (and if not, this one-line change will not help them understand the difference between eggs, wheels, etc.)

So my thought on this change was to reduce confusion for 99% of users who don't need to know about `easy_install`.

belak:
Honestly, that's my preference as well. There isn't much use in keeping around install information for a bunch of package managers when the people who need to know about it already do.

johnpneumann:
Agree with everything you said and understand - as someone who also uses
python for his "day job". As I said originally, because it's a doc not a
functional change I see nothing wrong with this.

belak:
Alright, merged. Thanks to everyone for weighing in!

johnpneumann:
:) I only know because I've had the "fortune" of working on legacy systems,
but agree that OSS involvement definitely expands my knowledge about things
I'd never had known (issues with prompts being a good example since I'd
never looked at the zsh man pages up to that point).

PR #1450, opened 2017-08-29
[command-not-found] Document support for macOS

jeffwidman:
This adds docs for b0c85eddf6bee4d75ca9920c9c6180ab7ff1962a which was added in https://github.com/sorin-ionescu/prezto/pull/1376/


belak:
This seems like a useful update. Thanks!

PR #1455, opened 2017-08-29
[python] cleanup readme formatting

jeffwidman:
Anyone want some backticks? :-)


indrajitr:
FWIW, @sorin-ionescu preferred to use *italics* for pathnames and `backticks` for command/code snippets. I don't have any strong opinion one way or the other, but we certainly need to have a consistent convention.

PR #1457, opened 2017-08-31
Add powerlevel9k theme

Twanislas:
Hi ! :)

## Proposed Changes
- Add the "[powerlevel9k](https://github.com/bhilburn/powerlevel9k)" theme as a submodule

Here's how it looks :

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,3 +25,6 @@
 [submodule "modules/prompt/external/async"]
 	path = modules/prompt/external/async
 	url = https://github.com/mafredri/zsh-async.git
+[submodule "modules/prompt/external/powerlevel9k"]
+	path = modules/prompt/external/powerlevel9k
+	url = https://github.com/bhilburn/powerlevel9k.git
--- a/modules/prompt/functions/prompt_powerlevel9k_setup
+++ b/modules/prompt/functions/prompt_powerlevel9k_setup
@@ -1,0 +1,1 @@
+../external/powerlevel9k/powerlevel9k.zsh-theme


indrajitr:
Submodule might be relatively easier to include. @belak, what do you think?

belak:
I was looking at including this as a submodule a while back but never got around to it.

I'd be happy to merge this, but I'd like to see it pinned to the latest release (v0.6.4) rather than to the latest commit on master.

Twanislas:
I'd be really happy to do that, however that goes beyond my knowledge with Git. Are you referring to branches, tags or release of this repo ? I don't really understand, sorry !

belak:
powerlevel9k tags releases. This page should give you some info on how to change what the submodule is pointing to: http://www.vogella.com/tutorials/GitSubmodules/article.html#submodules_track

Note that you can use `git checkout` on tags as well as branches. :)

diraol:
Since you only have one commit on this PR, I think you can just ammend a small change to it in order to track the v0.6.4 tag. :)

You can do:
```
cd <powerlevel9k_dir>
git checkout v0.6.4
cd ..
git add powerlevel9k
git commit --amend --no-edit
git push -f
```

I just want to point out that `git push -f` can be dangerous, on the future use it with extremely caution and only if you know what you are doing. ;)

But I do recommend you to read the link suggested by @belak  .

Twanislas:
Thanks guys for your comments, another day with great lessons ! :)

indrajitr:
Thanks @Twanislas for your first contribution!

Twanislas:
You're very welcome ! Thanks for your help in setting this up ;)
Prezto is awesome, keep up the great work !

diogoazevedos:
Is the follow error related with this merge?

```
Couldn't read file /Users/diogoazevedos/.zgen/sorin-ionescu/prezto-master/modules/prompt/functions/prompt_powerlevel9k_setup containing theme powerlevel9k.
```

diogoazevedos:
I just reinstall everything and it works again 😝

diraol:
@diogoazevedos Yes, it was. after doing the `git pull` you should have run `git submodule update powerlevel9k` in order to "git pull" the submodule content. =)

rjcoelho:
This prompt is very slow on git repo directories. 
Using zsh 5.4.1 (x86_64-apple-darwin16.7.0)

diraol:
yeah, @rjcoelho I do have some problems with that too. The bigger the git-repo the slowest. =(

Twanislas:
The best way I found to measure it was pasting the following in my terminal :
```sh
date +"%T.%N"
date +"%T.%N"
# note the extra line here to send a carriage return
```
Here's the result :
```
  ~ 
 date +"%T.%N"                                                                                                         
22:18:55.704987000

  ~ 
 date +"%T.%N"                                                                                                         
22:18:56.964205500

  ~ 
 cd .dotfiles                                                                                                          

  ~/.dotfiles     master 
 date +"%T.%N"                                                                                                         
22:19:25.194600500

  ~/.dotfiles     master 
 date +"%T.%N"                                                                                                         
22:19:30.096161100
```

We can see that when not in a git dir, it takes ~1s but when inside a git dir it adds up to ~5s. That's quite slow.

Also it's worth noting that these numbers come from running it on my PC with a normal hard drive. I have almost no slowdown when running it on my Macbook Pro which has a very fast SSD. That makes me think it comes from how powerlevel9k scans the git repo to populate the prompt infos.

That would be worth investigating and finding a better way to run this test, with and without prezto (vanilla zsh) so we can be sure it's the theme's fault before opening an issue to powerelevel9k's repo :).

Issue #1424, opened 2017-08-10
Decide on official criteria for new modules


belak:
### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

### Versions

  - Prezto commit:
  - ZSH version:
  - OS information:

jeffcox:
I think it would make the most sense to keep prezto lightweight and instead direct people to a separate "modules" project which could be more accepting.  This would keep the base install reasonable while encouraging and centralizing modifications.

belak:
There was a bit of discussion in #1411 (and specifically, I'd like to mention https://github.com/sorin-ionescu/prezto/pull/1411#issuecomment-321683521)

I've already sent @nicoulaj a message about hosting it under the zsh-users project and he seemed open to it as long as we have an identified set of leaders and we can promise it won't be a temporary project like last time.

johnpneumann:
So my opinion on this is the same as what @jeffcox stated, however, since "lean" was brought up, I'm going to (do what apparently is my favorite thing at this point) and quote @sorin-ionescu, because I believe the relevance to the questions is there:

> I will not limit collaborators as long as they remain true to the project by not turning into something it is not.
> I expect collaborators to continue the work in the same manner it has been done, for it is the reason why we are all here.

> Prezto must remain lean.

To provide additional context, rather than just quotes, I'll give a summation of my thoughts (which generally means it'll be long, wordy and you'll fall asleep halfway through, so enjoy). 

- Unless it is believed to benefit 80% of the community, it should not be merged. To determine how we'd gauge that, we'd need to track users and we're not going to do that. So it will always be something somewhat subjective. Even something as universal as git (the irony of posting this on here is not lost on me) is only used by a certain userbase. Can we say with a level of certainty that 80% of people using prezto are using git. Yes. I'm fairly certain we can put that at 100% (even if it's only so they could pull it down, though I genuinely can't imagine doing that). I think there needs to be additional discussion around this in particular, because without data, you cannot make an informed decision. Maybe an anonymous survey of some sort where we just ask for what modules people are loading would suffice. I'm not sure how I feel about that, but I know that when other OSS projects ask for my usage info in a survey, I willingly give it (especially given that my benefit is generally greater than the time I spend on the survey).

- People can setup their own aliases fairly easily and utilizing others is easily done as well, so I don't think this is something we want to support.

[...]

PR #1458, opened 2017-09-01
Allow modules to be loaded from multiple places

belak:
This is initial work for the contrib repo, mentioned in #1424

## Proposed Changes

  - Allow for loading additional modules from `$ZPREZTODIR/contrib`
  - Note that this changes how functions from modules are loaded if multiple modules are loaded at once... this was done to make the code a bit simpler, but it could be changed back if need be. Previously we added all the modules being loaded to the fpath then loaded the modules one by one. The new code adds a module to the fpath then tries to load the module, following that pattern for each module.

## Additional Ideas
  - Add a zstyle option to add additional module dirs, so someone could add ~/.zmodules if they wanted to
  - A simple way of cloning the contrib repo (which currently doesn't exist)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.zwc
 *.zwc.old
 modules/*/cache.zsh
+contrib
--- a/init.zsh
+++ b/init.zsh
@@ -89,9 +89,6 @@ function pmodload {
   for pmodule in "$pmodules[@]"; do
     if zstyle -t ":prezto:module:$pmodule" loaded 'yes' 'no'; then
       continue
-    elif [[ ! -d "$ZPREZTODIR/modules/$pmodule" ]]; then
-      print "$0: no such module: $pmodule" >&2
-      continue
     else
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(/FN)})
       if (( ${#locations} > 1 )); then
--- a/init.zsh
+++ b/init.zsh
@@ -78,9 +78,15 @@ function pmodload {
   local pmodule_location
   local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
 
-  # Any dirs which prezto modules can be loaded from. Currently this is only the
-  # main modules dir and the optional contrib dir.
-  pmodule_dirs=("$ZPREZTODIR/modules" "$ZPREZTODIR/contrib")
+  # Load in any additional directories and warn if they don't exist
+  zstyle -a ':prezto:load' pmodule-dirs 'user_pmodule_dirs'
+  for user_dir in "$user_pmodule_dirs[@]"; do
+    if [[ ! -d "$user_dir" ]]; then
+      echo "$0: Missing user module dir: $user_dir"
+    fi
+  done
+
+  pmodule_dirs=("$ZPREZTODIR/modules" "$ZPREZTODIR/contrib" "$user_pmodule_dirs[@]")
 
   # $argv is overridden in the anonymous function.
   pmodules=("$argv[@]")
--- a/init.zsh
+++ b/init.zsh
@@ -72,26 +72,18 @@ function zprezto-update {
 # Loads Prezto modules.
 function pmodload {
   local -a pmodules
+  local -a pmodule_dirs
+  local -a locations
   local pmodule
+  local pmodule_location
   local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
+
+  # Any dirs which prezto modules can be loaded from. Currently this is only the
+  # main modules dir and the optional contrib dir.
+  pmodule_dirs=("$ZPREZTODIR/modules" "$ZPREZTODIR/contrib")
 
   # $argv is overridden in the anonymous function.
   pmodules=("$argv[@]")
-
-  # Add functions to $fpath.
-  fpath=(${pmodules:+$ZPREZTODIR/modules/${^pmodules}/functions(/FN)} $fpath)
-
-  function {
-    local pfunction
-
-    # Extended globbing is needed for listing autoloadable function directories.
-    setopt LOCAL_OPTIONS EXTENDED_GLOB
-
-    # Load Prezto functions.
-    for pfunction in $ZPREZTODIR/modules/${^pmodules}/functions/$~pfunction_glob; do
-      autoload -Uz "$pfunction"
-    done
-  }
 
   # Load Prezto modules.
   for pmodule in "$pmodules[@]"; do
@@ -101,6 +93,33 @@ function pmodload {
       print "$0: no such module: $pmodule" >&2
       continue
     else
+      locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(/FN)})
+      if (( ${#locations} > 1 )); then
+        print "$0: conflicting module locations: $locations"
+        continue
+      elif (( ${#locations} < 1 )); then
+        print "$0: no such module: $pmodule"
+        continue
+      fi
+
+      # Grab the full path to this module
+      pmodule_location=${locations[1]}
+
+      # Add functions to $fpath.
+      fpath=(${pmodule_location}/functions(/FN) $fpath)
+
+      function {
+        local pfunction
+
+        # Extended globbing is needed for listing autoloadable function directories.
+        setopt LOCAL_OPTIONS EXTENDED_GLOB
+
+        # Load Prezto functions.
+        for pfunction in ${pmodule_location}/functions/$~pfunction_glob; do
+          autoload -Uz "$pfunction"
+        done
+      }
+
       if [[ -s "$ZPREZTODIR/modules/$pmodule/init.zsh" ]]; then
         source "$ZPREZTODIR/modules/$pmodule/init.zsh"
       fi
@@ -109,7 +128,7 @@ function pmodload {
         zstyle ":prezto:module:$pmodule" loaded 'yes'
       else
         # Remove the $fpath entry.
-        fpath[(r)${ZPREZTODIR}/modules/${pmodule}/functions]=()
+        fpath[(r)${pmodule_location}/functions]=()
 
         function {
           local pfunction
@@ -119,7 +138,7 @@ function pmodload {
           setopt LOCAL_OPTIONS EXTENDED_GLOB
 
           # Unload Prezto functions.
-          for pfunction in $ZPREZTODIR/modules/$pmodule/functions/$~pfunction_glob; do
+          for pfunction in ${pmodule_location}/functions/$~pfunction_glob; do
             unfunction "$pfunction"
           done
         }
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -15,6 +15,9 @@
 # Color output (auto set to 'no' on dumb terminals).
 zstyle ':prezto:*:*' color 'yes'
 
+# Add additional directories to load prezto modules from
+# zstyle ':prezto:load' pmodule-dirs $HOME/.zprezto-contrib
+
 # Set the Zsh modules to load (man zshmodules).
 # zstyle ':prezto:load' zmodule 'attr' 'stat'
 


belak:
Ping @indrajitr, @johnpneumann, @jeffwidman, @facastagnini 

Does anyone have an opinions on this, or better ways of accomplishing it?

indrajitr:
Will review today, sorry for the delay.

belak:
No worries, thanks for following up! :)

belak:
No worries, thanks for following up! :)

belak:
What's the status on a review here?

jeffwidman:
Afraid I really don't know enough here to have a thoughtful opinion. :-(

belak:
That's alright! I just pinged all the current maintainers I'm aware of... so no worries. :)

johnpneumann:
Sorry. I started looking at this yesterday and got pulled away. I’ll take a look this week and respond over the weekend.

belak:
@johnpneumann, @indrajitr Any updates on this? I'd like to start pushing some modules and themes towards a contrib repo and this is currently the bottleneck.

rjcoelho:
Most zsh plugins use <module>.plugin.zsh as the entry source file to be sourced instead of init.zsh, so we could
``
if [[ -s "$ZPREZTODIR/modules/$pmodule/init.zsh" ]]; then
          source "$ZPREZTODIR/modules/$pmodule/init.zsh"
elif [[ -s "$ZPREZTODIR/modules/$pmodule/$pmodule.plugin.zsh" ]]; then
          source "$ZPREZTODIR/modules/$pmodule/$pmodule.plugin.zsh"
fi
``

This way i can git submodule include zsh plugins.

belak:
I like the idea of allowing more common plugin extensions, but I think that should be done in a separate PR. Feel free to file an issue so I don't forget about it.

rjcoelho:
That PR would depend on this one. That is why I raised the issue here.

belak:
Yep, I understand why. And I appreciate you mentioning it... however, I don't want to add it to this PR, as the main reason for this is #1483, something which is blocking quite a few other pull requests.

I've raised #1484 for another place to discuss this and I've assigned it to myself so I remember to look into it later.

johnpneumann:
Any update on some test steps and external modules to utilize to test this? I've got my test environments fairly well automated for testing PR's, so I can get to it this week and test on CentOS, Debian and Ubuntu. 

For those waiting on this: I'd like some defined testing outcomes (and an external module for testing) or at least some requirements defined so that we can say, "yes this works as expected" or "no it doesn't and here's what's broken". We've had quite a few regressions since all of us were added as contributors and I'd like to make sure that doesn't happen when we're dealing with, what equates to, the core of how everything works.

belak:
Sorry about the delay - this is what I've been using to test this so far: https://github.com/belak/prezto-experimental-contrib

This repo has two modules:

- `prompt-contrib` This should be loaded before `prompt` and adds the belak and gentoo prompts
- `wakeonlan` This is here only to conflict with the built-in wakeonlan module. Loading this should cause an error and neither conflicting module should be loaded.

I'm open to adding any other possible edge cases if need be, but I'm not sure if it'll be around much after this PR is merged.

johnpneumann:
@belak - So I ran this through the ringer and had a bunch of issues, but that appears
to be due to the fact that I hadn't pulled the latest (specifically it was bombing on
what you fixed in a18ec622df2d3787d7abfeec78a44ead6f297d97). After testing with the
legit version, this looks good, but I'm going to run it through other OS' before I
approve it (I had a Ubuntu 16.04 box available, but didn't want to try and pull
another down while I was on the train). The wakeonlan module did bomb where it was
supposed to and the themes available were loaded up and I was able to set them up.

One thing that I did notice is that, if users clone from contrib, and want to keep
their modules or load from other locations, they're unable to keep that as part of
their current setup due to the fact that the entire directory is in the
``.gitignore``. How can we alleviate this issue, giving users the ability to utilize
submodules without it becoming a pain for us? My initial thoughts are to do one or all
of the following things:

1. Put the contrib directory outside of `$ZPREZTODIR` and just specify that, they'll
need to put any additional modules inside of that directory. For instance, on first
pull, check if there's a `.zprezto-contrib` directory and if not, create it and
that's where users can put their modules that they want to load.
2. Add a new `$ZPREZTO_CONTRIB` environment variable that when unset is ignored. When
it's set, that directory is used to check for modules.

belak:
One of the things I was looking at possibly adding was a method of adding additional module dirs with zstyle options. This would let users add additional module locations, but still allow for a standard location for the contrib repo to be cloned.

https://github.com/sorin-ionescu/prezto/pull/1458/files#diff-70e532346dbd4347d1a49387eb6c552cR83

belak:
I threw together support for additional module dirs (30a6e88), but the current plan is still to keep the official contrib repo in $ZPREZTODIR/contrib.

diraol:
@johnpneumann I couldn't understand your point over here:

> One thing that I did notice is that, if users clone from contrib, and want to keep
> their modules or load from other locations, they're unable to keep that as part of
> their current setup due to the fact that the entire directory is in the
> .gitignore. How can we alleviate this issue, giving users the ability to utilize
> submodules without it becoming a pain for us?

What do you mean by `keep their modules`?
In regard to loading modules from other locations, or loading modules that are not distributed from prezto-contrib repo, I think @belak 's suggestion may solve the problem.

belak:
I believe the idea behind that was that a user couldn't just create a contrib directory and add their own modules to it because it was in gitignore. Loading modules from user defined locations was my way of working around that... letting them store their modules wherever they want.

belak:
@johnpneumann Have you had the chance to try this on other systems?

johnpneumann:
Alright, I made time for running through this, this evening (morning now... whatever). Ran all of it through `Vagrant`. I used the default zsh that came with apt/yum or the os (in the case of OSX). I noted the issues I had below which are few and out of our control. 

I'd like a second pair of eyes on this for whatever configuration folks may have, since this list is non-exhaustive, nor is it indicative of real world use cases, since I literally ran the Makefile I have to generate the vagrant image, which installed things, and then ssh'd in, checked out the branch, cloned the contrib repo from above, added the modules to my `.zpreztorc`, changed my prompt to `belak`, logged out and logged in again to verify that it worked as intended. Sounds exhaustive, but isn't in any way shape or form (I would have automated the latter, but didn't want to muck up the way the scripts are working for the former). I tried testing on Sierra, but it's late and things were breaking with the VM provisioning, so if anyone has Sierra any validation on testing of it would be appreciated.

## Tested on:
- [x] Ubuntu 16.04
- [x] Ubuntu 14.04
- [x] Debian Stretch
- [x] Debian Jessie
- [x] Debian Wheezy
- [x] CentOS 7
- [x] CentOS 6
- [x] Mac OS - El Capitan

### Issues:
- CentOS 6 
  - Threw an error due to an old version of zsh (4.3.11). That's a them problem, not an us problem.

## Untested
- [ ] Redhat
- [ ] Suse
- [ ] Gentoo
- [ ] Puppy
- [ ] All the other nix distros I cannot think of.
- [ ] Mac OS - Sierra
- [ ] Mac OS - High Sierra

belak:
I've been using this on my work computer (Sierra), my personal laptop (High Sierra), and an Archlinux VM for quite a while. 

With CentOS 6, I believe it's been fixed on the master branch (https://github.com/sorin-ionescu/prezto/pull/1491).

Thanks for doing all this testing! I really appreciate it.

johnpneumann:
Not a problem man. It's why we're both here. :) As an addendum to my review, I think we need some well defined docs around this: how it works, what to expect, avenues of support, etc. I'm fine with this being merged without that given that we don't have a contrib repo yet anyways, however, before the contrib repo goes live I'd like some docs around it. The question then is, how do we want it documented? In a readme? In the main readme? Wiki? Readthedocs? 

I personally lean towards the latter two, however, I don't want to mix docs into here, so I'm not sure if there's a good way around that with readthedocs. I like rtd because the layout is familiar for most folks and it's easy to navigate and document for. That being said, I don't want to add an entire Sphinx install with docs here at all. It just becomes bloat along the way. So if a wiki or a readme would suffice then lets do that instead.

hlecuanda:
Hey guys, I'm new here but I've using prezto since 2012 and for some time I got deep in the zsh internals, and played around with a port of python as a binary module (https://bitbucket.org/ZyX_I/zpython/commits/all) I see the maintainer ditched the work in progress and scaled back to implementing bindings instead of an interpreter... :( Anyway I'm cleaning up some code I wrote for "very vim zsh"  keybindings.. I Have some ideas about the contrib design, but instead of just popping up with unsolicited advice, I want to bring some code to back up a proposal... ;)

belak:
Also, for future testing, I just found https://github.com/zsh-users/zsh-docker which might be good for testing different versions in the future.

belak:
Unless something unexpected comes up (or someone points out a problem with the code), I plan on merging this some time this weekend.

Once it's merged, I'll add some basic docs and try to get an official contrib repo started.

belak:
Alright, it's time to merge this! Thanks to everyone for the comments and ideas. I'll be writing up some documentation later today so we can move forward with adding a contrib repo.

apjanke:
Hooray! This was a long time coming, and is going to make managing Prezto installs a lot easier.

🎉

apjanke:
Hey there! I just stumbled upon this today as I was in the process of trying out migrating from Oh My Zsh to Prezto again. This is great; it'll make Prezto a lot more manageable and extensible. The lack of a contrib mechanism (and the reliance on divergent forks) is the main reason I never stuck with Prezto the last couple times I tried it.

I'd like to suggest something: instead of having a single central contrib repo, how about adopting a convention or mechanism that lets users load in multiple named contrib repos from multiple authors, like [Mac Homebrew's Tap system](https://docs.brew.sh/Taps)? Like letting `contrib` hold various `contrib/$USER/$TOPIC` contrib repos that contain modules.

The issue with a single contrib repo is that the maintainer for it is the bottleneck for getting everybody's code out. And there's no good mechanism for sharing or installing works-in-progress. I think this bottleneck problem is why Oh My Zsh development basically died out. They ended up with a backlog of hundreds of PRs and Issues, most of them for small plugins that had no need to go through a central maintainer.

While I was playing with Prezto, I put together my own convention for doing this; written up [here](https://github.com/apjanke/zprezto-personal/blob/693df8ad36c875e30ca014c769837a580b12abcc/_doc/Prezto-Contrib-Conventions.md). Maybe you could adopt something like that? There'd be no need for additional tooling; just a convention about how to structure repos and where under `contrib/` to install them (that is: two levels down).

The `belak/prezto-contrib` repo could then be just one of many (though an official "blessed" one), conventionally located at `contrib/belak/contrib`.

belak:
Hey @apjanke. Thanks for the well thought out comment and docs.

It is already possible to load from multiple locations, but it (currently) needs to be managed manually. `:prezto:load:pmodule-dirs` can be used to specify additional locations to load from. However while I see the value, I don't currently have any plans to support something like the tap system (though, as this is an experimental feature I'd probably be alright accepting a pull request to change module loading to work like that).

Currently, it's designed in such a way that you can only load a module if there are no conflicting modules and there isn't a way to specify which one you want to load - this would need to be resolved if the tap system was adopted.

You're definitely right that a single repo with a single maintainer is a major bottleneck - the original base16 repo and oh-my-zsh (though OMZ has been doing a bit better lately) are good examples of this.

One of the reasons I've held off on this is because it adds complexity and prezto is already quickly approaching the point where development is a bit overwhelming because we need to deal with the interactions between plugins. Support for tons of external modules will also increase work on this repo because issues are going to be filed relating to external modules.

I've actually been trying to take a step back with this repo as much as possible to avoid getting burnt out... I only use a very small portion and started porting those into a set of [zsh-utils](https://github.com/belak/zsh-utils) which has a bit more of an explanation of my thoughts on the matter.

Those thoughts ended up being a bit more scattered than I would have liked, but the general idea is that I don't know what will happen with prezto long term, I think plugin managers are better than re-implementing another plugin manager inside prezto, but your proposal seems very useful and I'd be happy to see a PR which implements it.

apjanke:
That makes sense.

However, the thing about making space for "tap" style installations is that it doesn't require any code changes: all it needs is for you to have users install your main `prezto-contrib` repo at `~/.zprezto/contrib/belak/contrib` instead instead of at `~/.zprezto/contrib`, and treat it as a plain user-managed repo instead of a Git submodule. Or add the `contrib/belak/` dirs to the main `prezto` repo (with its `.gitignore` configured to ignore all the other subdirs under `contrib/` and `contrib/belak/`); then it could still be a submodule.

Any "tap management" functions added to Prezto would then be just a convenience, not a necessity.

I kind of prefer being able to manage my whole Prezto setup under a single directory instead of fiddling with the `:prezto:load:pmodule-dirs` to manage it. But I admit I haven't really tried that out. I'll give it a spin and see how it works. Maybe this is unnecessary after all.

What do you think?

PR #1460, opened 2017-09-04
[prompt_sorin] Use ✘ and show return value for non-0 rtrns

samcv:
If you do not want it to show the return code, you can set:
`zstyle ':prezto:module:prompt' show-return-val false`

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -129,10 +129,15 @@ function prompt_sorin_setup {
   zstyle ':prezto:module:git:info:keys' format \
     'status' '%b %p %c:%s%A%B%S%a%d%m%r%U%u'
 
-
+  # Set up non-zero return value display
+  local show_return="✘ "
+  # Default is to show the return value
+  if zstyle -T ':prezto:module:prompt' show-return-val; then
+    show_return+='%? '
+  fi
   # Set python-info format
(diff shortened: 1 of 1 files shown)


modules/prompt/functions/prompt_sorin_setup
```diff
@@ -129,10 +129,14 @@ function prompt_sorin_setup {
   zstyle ':prezto:module:git:info:keys' format \
     'status' '%b %p %c:%s%A%B%S%a%d%m%r%U%u'
 
-
+  # Default is to show return value
+  local show_return="✘ "
+  if zstyle -T ':prezto:module:prompt' show-return-val; then
+    show_return='✘ $? '
```
belak: I believe this would be better as `%?` which relies on zsh tracking it for the prompt, rather than a shell variable.

samcv: Fixed!

belak:
If we're keeping the existing behavior as a default, I think we'd need to use `-t`... I'm not sure what the default should be... other than that it looks good.

samcv:
I had the default to show the return value. See screenshot attached for how it looks with this change:
screenshot_20170904_150845


This is how it used to look:
screenshot_20170904_150916

belak:
This works for me. Better to avoid doublewide characters anyway since quite a few terminals don't handle them properly (as seen in the bottom screenshot).

Thanks!

PR #1461, opened 2017-09-06
Updating submodules

diraol:
Updating `completition`, prompt `async` and `syntax-highlighting`. All modules where checkout to their latest tag.


belak:
`fasd` should not be changed to the latest tag, as it's missing a security fix. Additionally, it isn't updated very often, so leaving it shouldn't be a huge deal. Other than that, thanks for doing this!

diraol:
@belak I didn't know about the `fasd` security fix. I've removed it from the PR. =)

belak:
Merged! Thanks again!

PR #1465, opened 2017-09-08
Fix typo in zprezto-update function

cdlm:
Just a tiny one-character typo, but seeing it on every update got on my nerves :)


indrajitr:
nerves soothed :)

PR #1467, opened 2017-09-09
Bind more unbound keys in viins and vicmd to nothing

samcv:
Unbound keys in viins or vicmd mode have undefined results. Bind
two more of these to nothing to avoid this.


belak:
Looks good to me, though I still don't understand why this happens...

PR #1476, opened 2017-10-01
Pwd length

Zoybean:
## Proposed Changes

  - Ensure prompt-pwd always expands '~' when pwd-length is set to 'full', so the path format is consistent between directories.
  - Add pwd-length option 'tail' to only show the last directory in the path, for minimalist prompts.

### DIFF
--- a/modules/prompt/functions/prompt-pwd
+++ b/modules/prompt/functions/prompt-pwd
@@ -12,11 +12,11 @@ setopt localoptions extendedglob
 local current_pwd="${PWD/#$HOME/~}"
 local ret_directory
 
-if [[ "$current_pwd" == (#m)[/~] ]]; then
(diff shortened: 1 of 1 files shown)


modules/prompt/functions/prompt-pwd
```diff
@@ -12,13 +12,15 @@ setopt localoptions extendedglob
 local current_pwd="${PWD/#$HOME/~}"
 local ret_directory
 
-if [[ "$current_pwd" == (#m)[/~] ]]; then
+if zstyle -m ':prezto:module:prompt' pwd-length 'full'; then
+  ret_directory=${PWD}
+elif [[ "$current_pwd" == (#m)[/~] ]]; then
```
belak: Shouldn't this be the last branch before the `else` to avoid using this for other explicit options as well?

PR #1479, opened 2017-10-04
Use SSH_ASKPASS variable.

flying-sheep:
fixes #1478

PR #1480, opened 2017-10-04
ssh: update SSH_ASKPASS workaround to more closely match what ssh-add needs

### LINKED ISSUE #1478: SSH module still asks for the passphrase
#1320 resurfaced, as #1408 immediately broke what i fixed in #1334.

the `</dev/null` was there for a reason (explained in #1320)

belak:
Fixes #1478 (and possibly supersedes #1479)

Note that I haven't been able to test this since I'm currently on a Mac. If anyone can confirm this fix works, I'd be happy to merge it in.

Also note that I don't think checking the OSTYPE is needed, as in other environments, DISPLAY wouldn't be set, and that mirrors what ssh-add is looking for.

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -39,11 +39,20 @@ fi
 # Load identities.
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
-  # Check for Linux system and ssh-askpass presence
-  if [[ "$OSTYPE" == linux* ]] && [[ ! -a /usr/lib/ssh/x11-ssh-askpass ]]; then
-    ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}"  2> /dev/null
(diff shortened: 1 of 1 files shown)


flying-sheep:
> and possibly supersedes #1479

definitely, thank you!

belak:
Can you confirm that this fix works for you? I'm not as familiar with this module since I don't personally use it.

flying-sheep:
yes, it does! thanks again

belak:
Alright, it's been merged. Thanks for filing the issue!

crivotz:
Hi, just for history with #1334 shell no longer asks for the passphrase you must install _ssh-askpass_
In #1408 I tried to fix it for those who do not use ssh-askpass

Now I tested with and without _ssh-askpass_ in both case no _X11_ windows is open and works fine.

flying-sheep:
NP, i should have checked and reviewed your PR, then it could have fixed it for all 😄

leshow:
I think I have gotten bitten by this. I log in to a bare terminal and start with `startx` on arch. Previously, after login I would be asked my ssh password, and then it was saved and I never got asked again.

I recently updated and my passphrase is no longer saved, I get prompted in the terminal each time

PR #1482, opened 2017-10-09
Clarify in README that brewu updates AND upgrades packages / formula

kkirsche:
As homebrew includes both an update and upgrade command, it is beneficial to clarify that this performs both, rather than just the update.

## Proposed Changes

  - Update homebrew module README to clarify that both update and upgrade is run by the `brewu` alias.


belak:
Seems like a good clarification. Thanks!

kkirsche:
Of course, thanks for the quick feedback — keep up the great work on the project!

PR #1486, opened 2017-10-14
[tmux] Avoid run tmux when the shell is executed on emacs

ryukinix:
This avoid weird bugs to try running zsh through `C-x M-x`
and bug happens because obviously tmux is not designed to be
executed inside emacs.


belak:
For context, it looks like INSIDE_EMACS is set when using `M-x shell`.

Thanks for your PR!

PR #1490, opened 2017-10-20
ssh: always redirect input from /dev/null when adding keys

### LINKED ISSUE #1489: ssh pass doesn't get saved, continues to prompt
### Description

Recently pulled from master and updated, ssh-add no longer appears to run on my ssh ids
### Expected behavior

my defined ssh-ids should be added
### Actual behavior

I get prompted for every ssh action
### Steps to Reproduce
If I revert ssh/init.zsh to this:
```
  if (( ${#_ssh_identities} > 0 )); then
	ssh-add "$_ssh_dir/${^_ssh_identities[@]}"  < /dev/null 2> /dev/null
  else
	ssh-add  < /dev/null 2> /dev/null
  fi
```

everything works as expected, or at least as it did before (I login to tty, I put my username/pass, and then am prompted for my ssh password)

I noticed this change: https://github.com/flying-sheep/prezto/commit/5a72d7c31f72c99c984964e65ec6d2471098d996

Am I supposed to define `SSH_ASKPASS` somewhere? I can't use x11-ssh-askpass because I don't have x11 loaded at this point in time (as I saw in the change before this one). I log in, and I run `startx` if I want to initiate an xsession. I'd like for my ssh-agent to behave regardless of environment.

### Versions

  - Prezto commit: master branch
  - ZSH version: 5.4.2
  - OS information: Linux 4.13.7-1-ARCH

belak:
Should fix #1489

## Proposed Changes

  - Always redirect input from /dev/null

## Notes

From my limited testing, this doesn't break anything on OSX with or without identities or keys. I'm not 100% sure how this will interact with Linux if SSH_ASKPASS isn't set, but pulling together all the issues I've seen, this *should* fix the issue.

CC @leshow

Also CC @crivotz because this essentially reverts the change you made... can you provide any more context as to why the input redirection was removed in your PR?

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -39,21 +39,10 @@ fi
 # Load identities.
 if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   zstyle -a ':prezto:module:ssh:load' identities '_ssh_identities'
-  # ssh-add has strange requirements for running SSH_ASKPASS, so we duplicate
-  # them here. Essentially, if the other requirements are met, we redirect stdin
-  # from /dev/null in order to meet the final requirement.
-  #
-  # From ssh-add(1):
-  # If ssh-add needs a passphrase, it will read the passphrase from the current
-  # terminal if it was run from a terminal. If ssh-add does not have a terminal
-  # associated with it but DISPLAY and SSH_ASKPASS are set, it will execute the
-  # program specified by SSH_ASKPASS and open an X11 window to read the
-  # passphrase.
-  if [[ -n "$DISPLAY" && -x "$SSH_ASKPASS" ]]; then
-    ssh-add "${_ssh_identities:+$_ssh_dir/${^_ssh_identities[@]}}" < /dev/null 2> /dev/null
-  else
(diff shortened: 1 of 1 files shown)


crivotz:
I had the problem that after the #1334 by @flying-sheep, I was asked for passphrase for each hosts without considering my ssh-agent because he use ssh-askpass thats it's not installed by-default.
In #1408 I introduced a check that If the machine was Linux and had `ssh-askpass` I used the @flying-sheep method, otherwise I use the old implementation without `ssh-askpass`

In the last modification #1480 I have not found any problems but it is also true that in my case X starts first.

belak:
Do you still run into that problem with this branch?

crivotz:
Yes, I have the same problem.
If you have installed ssh-askpass all works fine but without it you are prompted every time

flying-sheep:
please abandon this PR. the current way to do it is correct, see https://github.com/sorin-ionescu/prezto/issues/1489#issuecomment-338150109

belak:
Thanks to everyone for weighing in. I'm going to close this because it's not the actual fix.

PR #1491, opened 2017-10-20
Lower requirements down to 4.3.11 again

ParthKolekar:
## Proposed Changes

  - ZSH Syntax Highlighter added a dependency for zsh version 4.3.17 as minimum due to a feature that they added. That broke `prezto` for everyone under `4.3.17`. They have now re-added compatibility for zsh `4.3.11`, so we can support it again. I request this because `4.3.11` is the maximum zsh release found in epel for CentOS6. 
 
I tested this for my environment and it works there peachy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ Installation
 ------------
 
 Prezto will work with any recent release of Zsh, but the minimum required
-version is 4.3.17.
+version is 4.3.11.
 
   1. Launch Zsh:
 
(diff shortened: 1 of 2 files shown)


ParthKolekar:
Relevant Issue

- https://github.com/zsh-users/zsh-syntax-highlighting/issues/368

chauncey-garrett:
Seems fine to me to allow v4.3.11 as long as the syntax highlighting module is up to date and there aren’t other modules that require it. 

Glad they’ve fixed the regression.

jeffwidman:
> That being said, I don't know when we'll end support for old versions of zsh. I really don't like supporting them indefinitely...

👍

PR #1498, opened 2017-10-26
Update init.zsh

ParthKolekar:
It seems that my older git does not support '@' yet.

## Proposed Changes

  - I run git version `1.8.3.1`, and it does not seem to like '@', so I seek to use a slightly verbose solution that works for me.


belak:
This makes sense. Thanks for your contribution!

PR #1500, opened 2017-11-01
Set key binding to expand all aliases including global

akarzim:
control-space expands all aliases, including global.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -226,6 +226,14 @@ function prepend-sudo {
 }
 zle -N prepend-sudo
 
+# Expand aliases
+function glob-alias {
+  zle _expand_alias
+  zle expand-word
+  zle magic-space
+}
+zle -N glob-alias
+
 # Reset to default key bindings.
 bindkey -d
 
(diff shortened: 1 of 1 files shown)


modules/editor/init.zsh
```diff
@@ -368,6 +376,9 @@ for keymap in 'emacs' 'viins'; do
 
   # Insert 'sudo ' at the beginning of the line.
   bindkey -M "$keymap" "$key_info[Control]X$key_info[Control]S" prepend-sudo
+
+  # control-space expands all aliases, including global
+  bindkey -M "$keymap" "$key_info[Control] " globalias
```
belak: Should this be glob-alias, similar to above?

akarzim: My bad. I fixed it.

akarzim:
Which built in expansion?

The reason behind `^<Space>` is to expand on demand, not each time I press the space bar.

belak:
I'm still not exactly sure what this does, but I haven't had much time to look at it yet either. If you could give an example of where you use this, that would help me understand what this does.

akarzim:
Here is an example.

```sh
alias rbb=bundle
alias rbbi='bundle install --path vendor/bundle'
alias -g G='| grep'
```

globalias zsh

PR #1502, opened 2017-11-02
[editor] Add 'gg' and 'G' to vicmd mode

samcv:
G Moves to the start of the last line if no numeric argument. Otherwise
goes to the line number. 9G goes to the 9th line. 1G to the 1st line.

By default "gg" is bound to beginning-of-buffer-or-history for viins, which
causes inconsistent things to happen (moving to the first history entry).
Bind move-to-buffer-beginning so it acts more like vim

Also: Add additional key combinations

Add ControlUp, ControlDown, ControlHome, ControlEnd, ControlInsert,
ControlDelete and add additional ones for ControlPageUp and
ControlPageDown.


hlecuanda:
I have a `zsh` script that gets loaded from my `.zlogin` after all the prezto stuff that adds a lot of the missing or obvious `vi` maps (vanilla vim, no plugins) both for `vicmd` mode and `viedit` mode. 
Being a bit obsessive at the time, I even mapped `logout` to `:q` and `zzz` and have `vicmd` mode as the default map, because that's how vi works haha. I was tempted to reimplement vim as a `zle` widget, but work got in the way of fun xD

If you don't mind waiting a couple of days I'll wrap those keybindings in a nice prezto module , taking out some nonstandard quirks I've picked up over the years, add some tests and I'll contribute it to prezto ..

hlecuanda:
@belak  I've triaged this changes, being a long time user of `vi` mode in zsh. 

The standard behavior zle's widgets, oddly, feels more natural as the outcome depends on what is on your buffer. On an empty buffer, its almost expected to just go and get history. this new widget forgoes that behavior and could even feel as a regression. I guess its a matter of habit or getting used to it. but other widgets with a vertical motion are bound by default to overflow to history (cfr: `up-line-or-history` `down-line-or-history` etc)  just my $0,02

belak:
I'm not sure what to do about this... because gg and G in vim are definitely start and end of the buffer in vim. There's ^ and $ for start and end of line which I assume work in zsh as well (I think I use vim-rsi which binds C-a and C-e for this as well). That being said, there isn't the same concept of history in vim - I'd like to see a compromise somewhere. However, since I don't use vi-mode in zsh, it should probably be someone else making this decision.

samcv:
@belak this could be a configuration option then?

belak:
Personally, I think the change makes sense, now that I understand it... I'm not sure about the history argument, but that can always be changed later.

What about enabling this change by default and providing an option which disables any overriding default vi-mode keybinds? I think that's the method I'd go for.

samcv:
@belak or would it be better to only do the history functions when nothing is in the buffer?

PR #1503, opened 2017-11-02
Prunes local tracking branches that do not exist on remote anymore

akarzim:
Here is a usage example : 

```
❯ gfp
From git.example.com:example/repo
 - [deleted]         (none)     -> origin/feature/branch
Deleted branch feature/branch (was 4eaac1c).
```

inspired by : [SO — how to prune local tracking branches that do not exist on remote anymore](https://stackoverflow.com/questions/13064613/how-to-prune-local-tracking-branches-that-do-not-exist-on-remote-anymore)

### DIFF
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -114,6 +114,7 @@ zstyle ':prezto:module:git:alias' skip 'yes'
   - `gfcr` clones a repository into a new directory including all submodules.
   - `gfm` fetches from and merges with another repository or local branch.
   - `gfp` prunes local tracking branches that do not exist on remote anymore.
+  - `gfP` forcefully prunes local tracking branches that do not exist on remote anymore.
   - `gfr` fetches from and rebases on another repository or local branch.
 
 ### Flow
--- a/modules/git/README.md
+++ b/modules/git/README.md
@@ -113,6 +113,7 @@ zstyle ':prezto:module:git:alias' skip 'yes'
   - `gfc` clones a repository into a new directory.
   - `gfcr` clones a repository into a new directory including all submodules.
   - `gfm` fetches from and merges with another repository or local branch.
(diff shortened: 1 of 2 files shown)


belak:
So, this runs `git fetch --prune` then cleans up branches which don't exist any more. Is there any way to do the same operation without the fetch?

I like the idea, but this feels like a weird combination of a branch operation and fetching, which makes it a little confusing as to whether it should go under `gfp` or `gbp`.

diraol:
You are right to say that this seems a little bit confusing (fetch x branch). But the idea is to _fetch_ the remote server to see the existing branches and then remove the local branches that have no remote counterpart. But I do not have a strong opinion if we should go with `gfp` or `gbp`.

adambyrtek:
I'm just a user, but I'm not sure whether this is generic enough to end up in a framework. This looks more like a custom shell script than a generic Git alias for everybody.

jeffwidman:
I like the goal here, it's useful for everyone who contributes to an active multi-user git repo, so from that standpoint, I personally don't mind adding to `prezto`. 

However, the last time I researched this, I backed away from adding it to my personal aliases because it was too dangerous. There are a lot of subtle edge cases, and to implement a generic command required making broad assumptions about all scenarios I was likely to encounter. I wasn't confident I'd remember to verify that all those assumptions held true when I ran the command against a repo.

Maybe things have changed or there's a lowest common denominator safety bar that could be used. Although in that case, I wonder if playing to the lowest common denominator takes away a lot of the power of this command, because it'd be so cautious at leaving a lot of branches.

johnpneumann:
Maybe for particularly onerous commands like this one we throw a warning prompt? I'm not super excited about that possibility though.

jeffwidman:
I'm going to close this as none of the maintainers seem super enthusiastic about it, and potentially dangerous as a broad/generic use case.

PR #1513, opened 2017-11-28
Source correct module's `init.zsh`

duemir:
Now that modules can be located in multiple placed (#1458) 
the appropriate `init.zsh` should be sourced. 

## Proposed Changes

  - Source `init.zsh` from the `$module_location` and not `$ZPREZTODIR/modules/$pmodule`

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -123,8 +123,8 @@ function pmodload {
         done
       }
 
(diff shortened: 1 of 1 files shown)


belak:
Oops. Good catch. Thanks for your contribution!

PR #1517, opened 2017-12-07
Update zsh-autosuggestions to v0.4.2

thetrompf:
This fixes issues with autosuggestions when using an older version of zsh
see: https://github.com/zsh-users/zsh-autosuggestions/issues/296 for more info


belak:
Looks good to me. Thanks!

Issue #1081, opened 2016-02-06
~RVM_PROJECT_PATH is shown in the path


shineycode:
I have the latest & greatest (commit 7227c4f0bef5f8ae787c65150d7a7403394fff48) and after installing and configuring zprezto to use 'minimal' for prompt, my command line reports ~RVM_PROJECT_PATH - [branch-name●] instead of the current working directory. It doesn't matter which prompt I change to, I see the same result.

I'm running Mac OS 10.11.3

sdomula:
I have the same effect when using test-kitchen. After some time the prompt changes to:
~_kitchen_instances_cache_dir (branch-name).

I am also running Mac OS 10.11.3

sorin-ionescu:
Add `unsetopt AUTO_NAME_DIRS` to zshrc.

PR #1526, opened 2018-01-14
Don't set auto_name_dirs because it messes up prompts

salmanulfarzy:
This was ported from Oh-My-Zsh and since have been disabled in OMZ.

Explained in more detail https://github.com/rvm/rvm/issues/3091#issuecomment-60083194

Related: #998 , #1081


belak:
This may upset people because it changes a default, but I'm not sure it made much sense as a default anyway. It's super easy for anyone to add back if they want it.

Thanks for your contribution!

PR #1533, opened 2018-01-25
Set INTERACTIVE_COMMENTS by default

maximbaz:
This allows using `#` to comment lines in the interactive shell:

image

Seems to me a pretty reasonable default, that's why making a PR.

Here's a more realistic example of when this can be useful:

- Start typing a command
- Realize you need to run another command first
- Prepend current command with `#` and hit Enter
- Run another command
- Up, Up in history to get the commented command
- Remove the `#` and keep typing the first command

I am in a little doubt of whether this is the correct module to set this option.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -14,8 +14,8 @@ fi
 # Options
 #
 
-# Beep on error in line editor.
-setopt BEEP
+setopt BEEP                     # Beep on error in line editor.
+setopt INTERACTIVE_COMMENTS     # Enable comments in interactive shell.
 
 #
 # Variables


belak:
Yeah, I'm a big fan of this option as well. Thanks for submitting this!

belak:
I moved the setting to the `environment` module because it seems to make more sense there (it claims to be for general settings). Should be good now. :)

maximbaz:
Awesome, thanks!

hlecuanda:
you may also want to add a keybinding for the ZLE widget  `pound-insert` (emacs) or `vi-pound-insert` (vi) 

This widget will toggle the `#` comment on the current line by prepending and  and accept the line (possibly commiting to history too) in one operation. When a multiple line command is sitting on the ZLE buffer and several lines are commented, `pound-insert` will toggle all `#` comments from such command in one keystroke. 

There is no standard keybinding for this awesome feature. I chose to use the same binding as the `vim-commentary` plugin I use (bound to toggle comments), but i doubt it is in any way universal:

`bindkey -M vicmd ",c" vi-pound-insert`
**keystroke sequence**: type type type *realise you want to comment* **[ esc ][ , ][ c ]** 

`bindkey -M emacs "M-C" pound-insert` *(i don't use emacs mode, so i'm not sure if that is even a sane mapping)* 
**keystroke sequence**:: type type type *realise you want to comment* **[ meta ][ c ]** 

(meta is usually bound to [ esc ] by default so they're almost equivalent, except for the comma "lead" i'm used to on vim. )

belak:
The keybind on emacs is generally "M-;" which I haven't been able to get working properly in iTerm2.

Additionally, when I bind `pound-insert` to something, it seems to add the `#` and then `<ffffffff>`. This is a good idea, but my initial experiments were fairly buggy. I'd gladly accept pull requests which add this feature (probably in the `editor` module).

maximbaz:
Nice idea indeed. I will personally probably use <kbd>Ctrl+/</kbd> because that's the same binding that I have in other editors (`bindkey -M emacs "^_" pound-insert`), however I noticed that in emacs mode it automatically submits the line as soon as comment `#` is toggled... If you know a way to prevent autosubmit, please share how to achieve this 🙂

belak:
@maximbaz I think the easiest way might be to reimplement `pound-insert` as a custom zle widget. I've got a PoC that I'll commit fairly soon.

I have no idea if it'll work in vi mode, so help testing would be appreciated.

maximbaz:
It works really well! As for vi mode, I believe the standard `vi-pound-insert` will not autosubmit, so perhaps you can keep using it, but I'm not sure - @hlecuanda uses vi mode, let's kindly ask him 🙂

hlecuanda:
yup, i can confirm the stock `vi-pound-insert` works as documented. I've been using it since late adolescence, when floppies came in 5 1/5" and 3 1/4" flavors.. n.n

I remember seeing on the old (and spam-ridden) zsh wiki there used to be a workaround for the Meta- issue, it had something to do with an optional keys[] array that gets loaded when using zsh/termcap or zsh/terminfo modules. It had something to do with the actual terminfo database containing a bug.

I do believe zprezto loads one of these modules, depending on whether it finds itself on a  BSD (termcap) or Linux (terminfo) system. The fix was trivial, by "monkeypatching" the array with the correct value for your emulator or using a terminal emulator that correctly implements the capabilties from the  term{cap,info}, database, so YMMV depending on your emulator.

iTerm2 is pretty good at emulating xterm, but xterm meticulously follows the documented capabilities, while urxvt and gnome-term, even xterm on tmux (!) and xterm on tmux via mosh (!!) deviate from the documented capabilities. Each layer supposedly "fixes" the previous layer, making for fun debugging expeditions on keyboard event translation ....

since i frequently jump around terminals, I ended up writing a fairly specific (to my needs) sequence to monkeypatch the array and have both F-keys, arrow-keys and spanish language symbols correctly interpreted... i'll dig into that ancient code (which i'm sure must be embarrassig haha), but some latent keyboard and emulation issues can be ironed out using zsh/term{cap,info} =)

PR #1544, opened 2018-02-07
Add zdebuglog for basic debug logging and profiling

belak:
## Proposed Changes

  - Add a zdebuglog function which only prints if `:prezto:debug` is enabled.
  - Add some basic startup logging to zmodload

## Reason

This came about so I could work on debugging slow startup times of some modules. There may be a better way - I'm open to improvements or suggestions.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -17,6 +17,19 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
 fi
 unset min_zsh_version
 
+typeset -F SECONDS
+
+function zdebuglog {
+  if ! zstyle -t ":prezto" debug; then
+    echo no debug
+    return
+  fi
+
+  local format="$1"
+  shift
+  printf "[%f] $format\n" $SECONDS "$@"
+}
+
 # zprezto convenience updater
 # The function is surrounded by ( ) instead of { } so it starts in a subshell
 # and won't affect the environment of the calling shell
@@ -93,7 +106,9 @@ function pmodload {
 
   # Load Prezto modules.
   for pmodule in "$pmodules[@]"; do
+    zdebuglog "Started loading %q" $pmodule
     if zstyle -t ":prezto:module:$pmodule" loaded 'yes' 'no'; then
+      zdebuglog "Module %q already loaded" $pmodule
       continue
     else
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
@@ -131,6 +146,7 @@ function pmodload {
 
       if (( $? == 0 )); then
         zstyle ":prezto:module:$pmodule" loaded 'yes'
+        zdebuglog "Module %q loaded" $pmodule
       else
         # Remove the $fpath entry.
         fpath[(r)${pmodule_location}/functions]=()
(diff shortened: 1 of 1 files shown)


init.zsh
```diff
@@ -17,6 +17,19 @@ if ! autoload -Uz is-at-least || ! is-at-least "$min_zsh_version"; then
 fi
 unset min_zsh_version
 
+typeset -F SECONDS
+
+function zdebuglog {
+  if ! zstyle -t ":prezto" debug; then
```
hlecuanda: I <3 `zstyle` we need more of those (python humor) 

Nonetheless there is an `zsh` idiomatic way of achieving this. (perhaps the way it works makes other debugging features harder to implement? depends on your implementation details).

 My suggestion is to use  a `TRAPNAL` function or define a `trap`-list with the `trap` builtin. ether the `TRAPNAL` function or the _builtin_ trap are set to be triggered upon `SIGDEBUG` emission, which happens before command execution by default. if you `unsetopt DEBUG_BEFORE_COMMAND` then `SIGDEBUG` is emitted **after** command execution. It's like a hook  function, but **awesome-er**

the difference between a `TRAPNAL` function and a `trap` builtin-defned list  is that functions are executed with their own context independent context while `trap` lists are "_injected_" to the calling command's context, so 
```zsh
        TRAPDEBUG { print $LINENO } #always prints 0
        trap 'print $LINENO' DEBUG #always prints the line it was called from 
```
also, `TRAPDEBUG` functions have an internal parameter `ZSH_DEBUG_CMD` automagically  available inside them, and it is set to the string representation of the ***next*** command to be executed _vis a vis_ the one being debugged after all. This allows for interesting stuff happening in logs, for example via plain `syslog` and a more interesting `zstyle` set to an array of commands we are interested in profiling instead of just a boolean (overkill perhaps) it could look something like this:
```
zstyle -a ':prezto:funcs_to_debug' pmodload zmodload
TRAPDEBUG {
  zstyle -g funcs_to_debug
  if $ZSH_DEBUG_CMD in $funcs_to_debug; then
    logger.info($ZSH_DEBUG_CMD,$(date %S),"more_details")
}
```
You can go crazy there and do even counterintuitive stuff, such as skipping the execution o the next command when certain conditions are met, besides membership to the `zstyle` array. It' perhaps an elegant solution to instrumentate THE WHOLE SHELL in just a few lines of code.

init.zsh
```diff
@@ -93,7 +106,9 @@ function pmodload {
 
   # Load Prezto modules.
   for pmodule in "$pmodules[@]"; do
+    zdebuglog "Started loading %q" $pmodule
```
hlecuanda: See my detailed comment above about the possibility of avoiding this and instead do automatic debug  instrumentation for _any_ command, scripted or interactive via `TRAPSIG` functions :smiley:

init.zsh
belak: In my initial experiments with TRAPDEBUG (and TRAP 'do_stuff' DEBUG, you don't get the function calls in ZSH_DEBUG_CMD, you get the function body, so while this would be much more flexible, it also doesn't quite do what I want.

PR #1545, opened 2018-02-12
Don't double-source chruby

ixti:
If chruby was installed using default way, most likely chruby and
auto-switching will be already sourced (at least on most Linuxes).


ixti:
Does this has any chances to be merged?

belak:
Yes, sorry for the delay. Looks good to me!

ixti:
HUGE! Thank you!

PR #1546, opened 2018-02-14
Allow users to easily define LS_COLORS and LSCOLORS

### LINKED ISSUE #1539: Directory color is too dark on Mac (allow users to define LS_COLORS)
Prezto is [hardwired](https://github.com/sorin-ionescu/prezto/blob/master/modules/utility/init.zsh#L93) to use dark blue for directory color both for `ls` and for completion on Mac, which is barely visible on black terminal. 

I can override `LSCOLORS` (for `ls`) by executing this after I load Prezto:
```
export LSCOLORS="ExGxBxDxCxEgEdxbxgxcxd" # ls colors like in Linux
```
It makes directory color light blue, and in general I like Linux colors better, e.g. green executables.

Unfortunately, I cannot do the same for `LS_COLORS` (for completion): redefining them before or after I load Prezto has no effect, probably because they are already copied to `zstyle ':completion:*' list-colors`.

You probably will not change your default colors to Linux, but maybe you could use the existing value of `LS_COLORS` if it was defined and not override it?

Update:

Found a way to override them:
```
export LSCOLORS='ExGxBxDxCxEgEdxbxgxcxd' # ls colors like in Linux
# used https://geoff.greer.fm/lscolors/ to convert them to LS_COLORS:
export LS_COLORS='di=1;34:ln=1;36:so=1;31:pi=1;33:ex=1;32:bd=1;34;46:cd=1;34;43:su=0;41:sg=0;46:tw=0;42:ow=0;43' 
zstyle ':completion:*:default' list-colors "$LS_COLORS"
```

belak:
Fixes #1539 

## Proposed Changes

  - Only set LSCOLORS and LS_COLORS if not already set

I'm not sure if there's a better solution, but this should allow users to define LS_COLORS and LSCOLORS while still falling back to our settings.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -76,10 +76,13 @@ if is-callable 'dircolors'; then
   alias ls='ls --group-directories-first'
 
   if zstyle -t ':prezto:module:utility:ls' color; then
-    if [[ -s "$HOME/.dir_colors" ]]; then
-      eval "$(dircolors --sh "$HOME/.dir_colors")"
-    else
-      eval "$(dircolors --sh)"
+    # Call dircolors to define colors if they're missing
+    if [[ -z "$LS_COLORS" ]]; then
+      if [[ -s "$HOME/.dir_colors" ]]; then
+        eval "$(dircolors --sh "$HOME/.dir_colors")"
+      else
+        eval "$(dircolors --sh)"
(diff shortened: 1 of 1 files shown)


belak:
Sorry, it took so long to merge this. I forgot it was still open.

PR #1549, opened 2018-02-23
Fix  .zprezto/modules/python/init.zsh:85: command not found: add-zsh-…

bryndin:
Fixes ".zprezto/modules/python/init.zsh:85: command not found: add-zsh-hook"  error that started happening recently upon zsh start

```
Last login: Thu Feb xx xx:xx:xx on ttys000
/Users/xxxxx/.zprezto/modules/python/init.zsh:85: command not found: add-zsh-hook
```

happens on
  zsh 5.3 (x86_64-apple-darwin17.0)
  zsh 5.4.2 (x86_64-apple-darwin17.3.0)

with the latest prezto


belak:
Thanks for the fix! Looks good to me.

PR #1551, opened 2018-02-24
git performance improvement if aliased

diraol:
This PR replaces #638 from @teemow, fixing its conflicts against the current master branch.

I am still not sure if there is any reason to not merge the original proposal (despite the merge conflicts), as asked by @belak [here](https://github.com/sorin-ionescu/prezto/pull/638#issuecomment-291265036), but if there is not, here is my contribution. =)

Copying below the origin PR description:
If git is aliased to hub the performance of the git module was very poor. The prompt was very laggy.

I replaced all occurrences of git with command git and used it for a while now. Performance is back to normal and all git info is displayed nicely.

### DIFF
--- a/modules/git/functions/_git-hub-browse
+++ b/modules/git/functions/_git-hub-browse
@@ -8,7 +8,7 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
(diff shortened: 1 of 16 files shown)


belak:
I see no reason to delay in merging this. Thanks for resubmitting it!

PR #1561, opened 2018-03-12
Fix shell prompts on Termux. (#1560)

### LINKED ISSUE #1560: Prompt fails to load in Termux
### Description

Note: I am just making this bug to close it with my PR.

I am surprised this hasn't been reported before.

On the Android terminal emulator Termux, the default configuration breaks the shell prompt completely.

### Expected behavior

You see the pretty prompt on login like you do on desktop.

### Actual behavior

Instead of showing the prompt, zsh shows an error message and leaves the user with a fallback dollar prompt.

```
$ zsh --login
pmodload: no such module: prompt
$ # useless white dollar prompt
```

### Steps to Reproduce

1. Install Termux
2. `pkg install zsh`
3. Install prezto the normal way
4. `zsh --login` or `chsh -s zsh` and open a new window

### Versions

  - Prezto commit: 
    b3c27bb, but has existed for as long as I remember
  - ZSH version: 
    zsh 5.4.2 (arm-unknown-linux-androideabi)
  - OS information: 
    LG G3 d851
    Qualcomm Snapdragon 801 (armv7, 32-bit)
    Android 7.1.2 via Resurrection Remix 5.8.5
    Termux 0.60

easyaspi314:

Fixes #1560

Basically, on Termux, completion sets NO_CASE_GLOB, but prompt can only be found if CASE_GLOB is set. This forces the glob to use CASE_GLOB. 

This is really more of a workaround for termux/termux-packages#1894, and once we find the bigger issue, this workaround should be removed.

## Proposed Changes

  - Temporarily enable CASE_GLOB when searching for plugins, as globbing acts weird on Termux.
  - This allows the prompt to load.
  - This works on macOS, but I haven't tested desktop Linux or Windows.
  - To do: More testing.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -103,8 +103,7 @@ function pmodload {
   for pmodule in "$pmodules[@]"; do
     if zstyle -t ":prezto:module:$pmodule" loaded 'yes' 'no'; then
       continue
-    else 
- 
+    else
       setopt CASE_GLOB
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
       setopt ${cg}_CASE_GLOB
--- a/init.zsh
+++ b/init.zsh
@@ -76,8 +76,16 @@ function pmodload {
   local -a locations
   local pmodule
   local pmodule_location
-  local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
 
+  # For some reason, Termux fails to locate the "prompt" module when NO_CASE_GLOB
+  # is set, but simply setting CASE_GLOB messes up the prompt itself.
+  # This temporarily enables it for the two globs, then immediately resets it.
+  # Adapted from https://unix.stackexchange.com/a/42031
+  local case_glob_prefix=NO
+  [[ -o CASE_GLOB ]] && case_glob_prefix= 
+  setopt CASE_GLOB
+  local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
+  setopt ${case_glob_prefix}_CASE_GLOB
   # Load in any additional directories and warn if they don't exist
   zstyle -a ':prezto:load' pmodule-dirs 'user_pmodule_dirs'
   for user_dir in "$user_pmodule_dirs[@]"; do
@@ -95,8 +103,11 @@ function pmodload {
   for pmodule in "$pmodules[@]"; do
     if zstyle -t ":prezto:module:$pmodule" loaded 'yes' 'no'; then
       continue
-    else
+    else 
+ 
+      setopt CASE_GLOB
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
+      setopt ${cg}_CASE_GLOB
       if (( ${#locations} > 1 )); then
         print "$0: conflicting module locations: $locations"
         continue
--- a/init.zsh
+++ b/init.zsh
@@ -106,7 +106,7 @@ function pmodload {
     else
       setopt CASE_GLOB
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
-      setopt ${cg}_CASE_GLOB
+      setopt ${case_glob_prefix}_CASE_GLOB
       if (( ${#locations} > 1 )); then
         print "$0: conflicting module locations: $locations"
         continue


init.zsh
```diff
@@ -76,8 +76,16 @@ function pmodload {
   local -a locations
   local pmodule
   local pmodule_location
-  local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
 
+  # For some reason, Termux fails to locate the "prompt" module when NO_CASE_GLOB
+  # is set, but simply setting CASE_GLOB messes up the prompt itself.
+  # This temporarily enables it for the two globs, then immediately resets it.
+  # Adapted from https://unix.stackexchange.com/a/42031
+  local case_glob_prefix=NO
```
belak: Should this be `NO_`?

init.zsh
```diff
@@ -76,8 +76,16 @@ function pmodload {
   local -a locations
   local pmodule
   local pmodule_location
-  local pfunction_glob='^([_.]*|prompt_*_setup|README*|*~)(-.N:t)'
 
+  # For some reason, Termux fails to locate the "prompt" module when NO_CASE_GLOB
```
belak: For fairly obvious reasons I'm not a huge fan of this fix - it's a dirty hack designed to work around a setting that probably shouldn't be set in the first place... but it also shouldn't cause issues either. I don't know why this glob would cause any problems anyway...

init.zsh
belak: Additionally, we may want to do this check after every module load so we don't accidentally clobber it if it's set in a module.

easyaspi314:
Well, I was looking at the source tree for zsh and I think I might have
actually found the real issue: I think zsh is assuming char is signed.

kursion:
Hi, any news about using prezto and android termux ?

Would be awesome to have prezto on my phone :D

uudashr:
Any updates?

emersonrp:
FWIW I found this via Google when seeing this same behavior in a fresh Centos VM;  applying this diff by hand fixed the problem.

alexanderadam:
Oh nice, I didn't know there is a bugfix for this.
Any chance to have this merged this year?

belak:
If my feedback is addressed, I don't see any reason it couldn't be merged.

alexanderadam:
@easyaspi314 are you planning to integrate the feedback on some point?

Boddlnagg:
I'm also hitting an issue that's similar to #1560 in a situation where my HOME directory is loaded from some network share via `/myshare/home/myusername/`, but then there's a second home directory at `/myshare/HOME/myusername/`, i.e. these two directory only differ in casing. When zprezto is installed in both of these directories, apparently it gets confused and doesn't know where to load modules from, although the actual home directory only ever points at *one* of those.

This PR seems to solve that issue, so I'd appreciate it getting merged.

easyaspi314:
This workaround is not enough. It is a bug with zsh itself. Feel free to use it temporarily though

alexanderadam:
> This workaround is not enough. It is a bug with zsh itself. Feel free to use it temporarily though

Can you please elaborate on that (since it always worked fine for me, so far)?

easyaspi314:
The actual bug occurs due to zsh not handling `EACCES` in its globbing function. This happens because Android doesn't allow you to access `/data/data` directly. 

I sent an email to the mailing list mentioning the actual bug, as a solution upstream would fix EVERY instance of this bug.

alexanderadam:
Ah nice. Do you have a link to a web version of the mailing list conversation?

easyaspi314:
I don't see it in the archives. I just looked it up, anyway.

If you want a temporary workaround to make zsh work, try this hacky shell script:

```shell
#!/data/data/com.termux/files/usr/bin/sh

apt update && apt install -y make clang termux-exec git
cd
git clone https://github.com/termux/termux-exec
cd termux-exec
cat >> termux-exec.c <<EOF
#include <dirent.h>
// Wrapper for opendir() which will fake the existence of "/".
DIR *opendir(const char *filename)
{
    DIR *ret;
    DIR *(*real_opendir)(const char *) = dlsym(RTLD_NEXT, "opendir");
    if (filename[0] == '/' && !strstr(filename, "com.termux") && access(filename, R_OK) != 0) {
        char *buf = malloc(strlen(filename) + strlen("/data/data/com.termux/files/home/termux-exec/fakeroot") + 1);
        strcpy(buf, "/data/data/com.termux/files/home/termux-exec/fakeroot");
        strcat(buf, filename);
        ret = real_opendir(buf);
        free(buf);
    } else {
        ret = real_opendir(filename);
    }
    return ret;
}
EOF
make install
mkdir -p fakeroot/data/data
cd fakeroot
# make some symlinks
for i in proc dev sdcard mnt storage system
do
    ln -s /$i $i
done
cd data/data
# link the data folder
ln -s /data/data/com.termux com.termux
echo "Done, now restart Termux."
```

alexanderadam:
Would it make sense that you make a patch for ZSH to the Termux package?

easyaspi314:
No, it needs a real fix. I am searching for one now.

easyaspi314:
I submitted a proper patch for upstream and a PR to backport it for termux-packages. 

Make sure to do

```
apt install --reinstall termux-exec
```
to undo the ugly hack.

msdos:
Backport PR for those interested: https://github.com/termux/termux-packages/pull/6246

easyaspi314:
The fix is up in the Termux repos, and should eventually be applied upstream.

keyb0ardninja:
Reading this thread gives me the impression that the issue was fixed in January 2021, but I'm still experiencing the exact issue today in September 2021. Am I missing something?

termux prezto error

iacchus:
@keyb0ardninja was exactly the same here, I changed for Oh-My-Zsh and worked like charm
(prompt theme is `powerlevel10k`.)

Here is the command for installing it: https://wiki.termux.com/wiki/ZSH

Screenshot_20211218-103829

keyb0ardninja:
@iacchus Yeah, I too ended up using Oh-My-Zsh since that's the one that worked.

ryukinix:
I fixed my zsh zpresto over termux based on that fix. Thank you, PR opener. @easyaspi314

Issue #1449, opened 2017-08-29
Rename "OS X" module to "macOS"


jeffwidman:
Apple renamed `OS X` to `macOS`:
1. We should start thinking about whether/how to migrate this module (a breaking change)
2. Rename OS X wherever it shows up in the docs: https://github.com/sorin-ionescu/prezto/search?utf8=%E2%9C%93&q=OS+X&type=

rjcoelho:
If you do that (a breaking change) then prezto should have a MIGRATIONS.md and be semantic versioned (in other words bump the major in a major.minor.patch format)

belak:
We could also symlink the module and/or have aliases when loading modules. I'm not sure what the best way would be to handle this because we don't really have a mechanism for dealing with breaking changes.

See also: #1354

PR #1572, opened 2018-04-16
Rename "Mac OS X" to "macOS" in comments

ashishgandhi:
This only changes references in text for human consumption. It leaves
out renaming paths because that can cause breaking changes.

Related issue https://github.com/sorin-ionescu/prezto/issues/1449.

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -122,7 +122,7 @@ Initializes OCaml package management.
 OSX
 ---
 
-Defines Mac OS X aliases and functions.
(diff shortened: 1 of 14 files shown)


modules/README.md
```diff
@@ -122,7 +122,7 @@ Initializes OCaml package management.
 OSX
```
ashishgandhi: This is not renamed as this is the name of the module.

belak:

PR #1579, opened 2018-05-01
Export VIRTUAL_ENV_DISABLE_PROMPT when enabling virutalenv

cmcginty:
Add `export` to ensure the define is available in the shell so pyenv
or virtualenvwrapper do not duplicate the virtualenv name on the
prompt.


belak:
I know this works with virtualenvwrapper already but this does look like an oversight. Thanks for the PR!

PR #1582, opened 2018-05-06
Update cache files when .zpreztorc file is modified

### LINKED ISSUE #1581: fasd tab completion returns no result
Tab completion with fasd is not working on my system and I notice there is a difference between the fasd function in prezto and the default one. Maybe something has changed in fasd recently.

https://github.com/sorin-ionescu/prezto/blob/948d9b3aa51f9bbf320649801565dc9164181e5c/modules/fasd/init.zsh#L40-L47

```
$ fasd --init posix-alias
alias a='fasd -a'
alias s='fasd -si'
alias sd='fasd -sid'
alias sf='fasd -sif'
alias d='fasd -d'
alias f='fasd -f'
# function to execute built-in cd
fasd_cd() {
  if [ $# -le 1 ]; then
    fasd "$@"
  else
    local _fasd_ret="$(fasd -e 'printf %s' "$@")"
    [ -z "$_fasd_ret" ] && return
    [ -d "$_fasd_ret" ] && cd "$_fasd_ret" || printf %s\n "$_fasd_ret"
  fi
}
alias z='fasd_cd -d'
alias zz='fasd_cd -d -i'
```

cmcginty:
Fixes #1581


belak:
Looks good to me. My one request would be that the `||` be moved to the start of the next line, as that's how we seem to do it in other modules.

PR #1583, opened 2018-05-11
Remove zlogout's Apu quote

cbowns:
## Proposed Changes

  - Remove the Apu quote on logout. It's unnecessary, and I personally find it in poor taste. For more context about Apu and the examination and critique of his role on The Simpsons, read about and watch https://en.wikipedia.org/wiki/The_Problem_with_Apu

### DIFF
--- a/runcoms/zlogout
+++ b/runcoms/zlogout
@@ -9,12 +9,12 @@
 [[ -o INTERACTIVE && -t 2 ]] && {
 
 # To print a message on logout:
-# uncomment the commented-out lines below, an
-# insert your content below the cat:
+# uncomment all of the commented-out lines below, and
+# insert your content at <your message here>
 
 # cat <<-EOF
 #
-# add your message here
+# <your message here>
 #
 # EOF
 #
--- a/runcoms/zlogout
+++ b/runcoms/zlogout
@@ -8,14 +8,12 @@
 # Execute code only if STDERR is bound to a TTY.
 [[ -o INTERACTIVE && -t 2 ]] && {
 
-# To print a message on logout:
-# uncomment all of the commented-out lines below, and
-# insert your content at <your message here>
+SAYINGS=(
+    "So long and thanks for all the fish.\n  -- Douglas Adams"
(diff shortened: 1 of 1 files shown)


runcoms/zlogout
```diff
@@ -8,11 +8,14 @@
 # Execute code only if STDERR is bound to a TTY.
 [[ -o INTERACTIVE && -t 2 ]] && {
 
-# Print the message.
-cat <<-EOF
+# To print a message on logout:
+# uncomment the commented-out lines below, an
```
belak: Nit-picking: `uncomment the commented-out lines below, an`

I believe you meant `and`

cbowns: 👍

runcoms/zlogout
```diff
@@ -8,11 +8,14 @@
...
-EOF
-
-} >&2
+# cat <<-EOF
+#
+# <your message here>
+#
+# EOF
+#
+# } >&2
```
belak: Just realized that I don't think this was supposed to be commented out.

cbowns:
@nsp is there any easy way to have .logout rotate between them?

runcoms/zlogout
cbowns: whoops, yeah, lemme fix.

belak:
Thanks for your contribution! Merged!

nomasprime:
Personally find 'The Problem With Apu' argument pretty contentious. I don't have a single Indian friend that has a problem with the character, the majority seem to think the whole issue's ridiculous.

belak:
Thanks for your opinion @nomasprime. While there are probably plenty of people who are alright with the original quote, it's still a valid concern. If you have a problem with it being removed, you're welcome to open an issue and start a conversation there. Alternatively, you could add it back in your own fork. However, reviving a pull request that has been closed for over a year and a half is probably not the best place for a productive discussion to happen.

I personally think the new quotes are an improvement on multiple levels.

1. Whether you find the argument contentious or not, they're more inclusive.
2. There are now multiple quotes, which makes it a bit more interesting (on both a technical and practical level).
3. I happen to be a big Douglas Adams fan

PR #1586, opened 2018-05-18
Add module alias-tips

N4M3Z:
Introduces support for external module [djui/alias-tips](https://github.com/djui/alias-tips).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "modules/prompt/external/powerlevel9k"]
 	path = modules/prompt/external/powerlevel9k
 	url = https://github.com/bhilburn/powerlevel9k.git
+[submodule "modules/alias-tips/external"]
+	path = modules/alias-tips/external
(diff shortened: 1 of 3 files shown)


belak:
Hello, unfortunately we're trying to avoid any new modules in prezto-core, especially those with external dependencies. If you're willing to submit it elsewhere, it would probably be a better fit for the [prezto-contrib](https://github.com/belak/prezto-contrib) repo.

N4M3Z:
Hi @belak. Your point makes sense, I'll make a pull request to the contrib repository. I have about 5 more modules to contribute. I left Prezto for Zplug about about a year ago exactly because I was missing specific modules. The contrib folder is a good solution. Is the prezto team currently planning to move to a different way of organizing the project and modules?

belak:
This was submitted to contrib in https://github.com/belak/prezto-contrib/pull/10

belak:
This was submitted to the contrib repo, as mentioned above. Thanks for doing that by the way!

PR #1594, opened 2018-06-14
Fix path to global ssh known hosts file

### LINKED ISSUE #1593: Global ssh known hosts file not considered for completion
### Description

The host names used for `ssh` completion are only read from the user's `~/.ssh/known_hosts` file, not from the global known hosts file in `/etc/ssh/ssh_known_hosts`.

### Expected behavior

Completion with names from the global known hosts file.

### Actual behavior

Completion does not include names from the global known hosts file.

### Steps to Reproduce

1. Write new host key to `/etc/ssh/ssh_known_hosts`:
```
$ echo 'foo.example.com ssh-ed25519 AAAAC3NzaC1l123' > /etc/ssh/ssh_known_hosts
```
2. Attempt completion in a new shell, type `ssh <TAB>`
3. The name `foo.example.com` is not part of the completion

### Versions

  - Prezto commit: d4332e2faf420f73ada88397bee95dc4f1da6b97
  - ZSH version: `zsh 5.3.1 (x86_64-debian-linux-gnu)`
  - OS information: Debian stretch (stable)

fd0:
Fixes #1593 

## Proposed Changes
  - Corrects the path to the global known hosts file, `/etc/ssh/ssh_known_hosts`

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -107,7 +107,7 @@ zstyle ':completion::*:(-command-|export):*' fake-parameters ${${${_comps[(I)-va
 zstyle -a ':prezto:module:completion:*:hosts' etc-host-ignores '_etc_host_ignores'
 
 zstyle -e ':completion:*:hosts' hosts 'reply=(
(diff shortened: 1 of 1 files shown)


belak:
This looks right. Thanks for the fix!

fd0:
Thanks!

Issue #1595, opened 2018-06-19
brew cask search is deprecated


jedahan:
### Description
when using the homebrew module, `casks` runs `brew cask search`, but according to homebrew as of 10 days ago, `Warning: Calling \`brew cask search\` is deprecated!`.

### Expected behavior

I'm not sure if we should remove the `casks` alias, or have it do the same thing as `brews`. I am leaning towards the former.

### Steps to Reproduce

1. run `casks whatever`
2. observe warning and no results shown

### Versions

  - Prezto commit: 0d6e7de0312366322bf7e7e1dbb41697b5bcd3b1 
  - ZSH version: zsh 5.5.1 
  - OS information: Darwin 17.6.0

singularitti:
I have made a pull-request (#1597) to redirect `casks` to `brews`.

hairuo:
I met the similar problem looks like this:
$ brew cask search wiznote
Warning: Calling `brew cask search` is deprecated!
Use `brew search` instead.
/usr/local/Homebrew/Library/Homebrew/cask/lib/hbc/cli/abstract_command.rb:33:in `run'

==> Casks
wiznote ✔                                       wiznote ✔                                       wiznote ✔

belak:
I'm currently happy with the fix in #1623. Eventually we'll probably want to remove the deprecated aliases entirely, but that doesn't need to happen right now. Thanks for the feedback everyone!

PR #1597, opened 2018-06-21
Redirect `casks` to `brew search`

singularitti:
## Fixes
In most recent [brew](https://github.com/Homebrew/brew), calling `brew cask search` is deprecated.
It is suggested to use `brew search` instead.
See pull-request [#4316](https://github.com/Homebrew/brew/pull/4316) for details.

## Proposed Changes

  - Redirect `brew cask search` to `brew search` as suggested by Homebrew


singularitti:
A possible way to solve #1595?

dbernheisel:
I'm conflicted if this is the right way to go. I understand there's muscle-memory here, and this will rid the deprecation warning, but if Homebrew won't have "brew cask search" anymore then should we have an alias for it?

Or should it be replaced with something similar, but not functionally the same, like `brew cask search --casks`? (The difference here I think is that it's an offline search vs online search).

```
❯ brew search --help
                                                                                                                    
brew search, -S:
    Display all locally available formulae (including tapped ones).
    No online search is performed.

brew search --casks
    Display all locally available casks (including tapped ones).
    No online search is performed.

brew search [--desc] (text|/text/):
    Perform a substring search of cask tokens and formula names for text. If text
    is surrounded with slashes, then it is interpreted as a regular expression.
    The search for text is extended online to official taps.

    If --desc is passed, search formulae with a description matching text and
    casks with a name matching text.

brew search (--debian|--fedora|--fink|--macports|--opensuse|--ubuntu) text:
    Search for text in the given package manager's list.
```

singularitti:
@dbernheisel That's a good idea. I was thinking about when users are typing `casks`, they are expecting only casks are listed out. But currently, Homebrew will use `brew search`, resulting in `casks` will print some `Formulae`. That's not intuitive, I would rather remember it and use `brews` next time.

dbernheisel:
for what it's worth, the only thing I can come up with that is somewhat equivalent to an online search that's only casks is this:

```sh
$ brews chrome | sed -ne '/==> Casks/,$p'
==> Casks
chrome-devtools
chrome-remote-desktop-host
chromedriver
dmm-player-for-chrome
epichrome
google-chrome
google-chrome-beta
google-chrome-canary
google-chrome-dev
mkchromecast
```

But this loses all the formatting that brew gives. For example, I have chromedriver installed so I _should_ see it bolded with a checkmark

image

Maybe a `sed` or `awk` expert can help.

singularitti:
Homebrew officially has a command line tool for doing similar things: [homebrew-aliases](https://github.com/Homebrew/homebrew-aliases).

indrajitr:
We need to revisit the whole of `brew` and `cask` situation. Multiple `cask` subcommands have been unified into `brew` lately.

singularitti:
> We need to revisit the whole of brew and cask situation. Multiple cask subcommands have been unified into brew lately.

That's the point.

belak:
The fix in #1623 seems to do this and make it clear that these commands may not exist in the future. I'm going to close this for now since that PR does pretty much the same thing. Thanks for submitting this!

PR #1603, opened 2018-07-14
Improve startup time by skipping pyenv rehash

sloria:
After profiling startup time, I found that "pyenv rehash" is
by far the slowest piece. This change skips rehashing on init.

See https://github.com/pyenv/pyenv/issues/784#issuecomment-404850327

Also pass the shell explicitly, as suggested in
https://github.com/pyenv/pyenv/issues/784#issuecomment-404944318,
which provides a modest improvement.

In total, this improves startup time from 1s to 0.3s on my machine.
Using the following command:
`python -m timeit "__import__('subprocess').Popen(['zsh', '-i', '-c', 'echo']).communicate()"`

Result before this change:
10 loops, best of 3: 1 sec per loop

Result after this change:
10 loops, best of 3: 334 msec per loop

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -11,12 +11,12 @@
 if [[ -s "$HOME/.pyenv/bin/pyenv" ]]; then
   path=("$HOME/.pyenv/bin" $path)
   export PYENV_ROOT=$(pyenv root)
-  eval "$(pyenv init -)"
+  eval "$(pyenv init - --no-rehash zsh)"
 
 # Load package manager installed pyenv into the shell session.
 elif (( $+commands[pyenv] )); then
   export PYENV_ROOT=$(pyenv root)
-  eval "$(pyenv init -)"
+  eval "$(pyenv init - --no-rehash zsh)"
 
 # Prepend PEP 370 per user site packages directory, which defaults to
 # ~/Library/Python on macOS and ~/.local elsewhere, to PATH. The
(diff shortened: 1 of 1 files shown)


sloria:
Marking as WIP because this breaks display of the current virtualenv in the prompt.

sloria:
Ah, without running `pyenv rehash`, `$VIRTUAL_ENV` does not get set and therefore the virtualenv info can't be populated here:

https://github.com/sorin-ionescu/prezto/blob/86a2e0ff4f1e4d36c13175a587cd8082c8ebf5ee/modules/python/functions/python-info#L28

Furthermore, running `pyenv rehash` results in an error:

```console
$ pyenv rehash
basename: illegal option -- -
usage: basename string [suffix]
       basename [-a] [-s suffix] string [...]
```

@blueyed Any suggestions? Is there a way for `$VIRTUAL_ENV` to get set without running `rehash`?

modules/python/init.zsh
```diff
@@ -11,12 +11,12 @@
...
   path=("$HOME/.pyenv/bin" $path)
   export PYENV_ROOT=$(pyenv root)
-  eval "$(pyenv init -)"
+  eval "$(pyenv init - zsh --no-rehash)"
 
 # Load package manager installed pyenv into the shell session.
 elif (( $+commands[pyenv] )); then
   export PYENV_ROOT=$(pyenv root)
-  eval "$(pyenv init -)"
+  eval "$(pyenv init - zsh --no-rehash)"
```
blueyed: Using `zsh` here already makes sense by itself - and could go into a separate PR / commit.

blueyed:
> without running pyenv rehash, $VIRTUAL_ENV does not get set

Are you sure?
There is nothing in pyenv/pyenv-virtualenv that would set it, but activating a virtualenv does so - AFAIK.

blueyed:
Usage is wrong, should be `pyenv init - --no-rehash zsh`, otherwise you get:

```
% pyenv init - zsh --no-rehash
…
export PYENV_SHELL=--no-rehash
```

sloria:
@blueyed Ah, good catch. Doing `pyenv init - --no-rehash zsh` fixed the problem.

This is working now, and my startup time is now blazing fast. ⚡️ ⚡️ ⚡️

johnpneumann:
lgtm. @belak - thoughts?

belak:
Seems good to me as well!

indrajitr:
LGTM as well

indrajitr:
Since we are all aligned, applying this. Thanks @sloria, @blueyed!

PR #1604, opened 2018-07-18
Make node-info work without nvm/nodenv

citizen428:
## Proposed Changes

  - I'm using [n](https://github.com/tj/n) as my Node version manager and `node-info` didn't work, so I used a similar check as in `modules/node/init.zsh` to get it to work.


indrajitr:
Makes sense. Thanks for the PR @citizen428! 👍

Issue #1589, opened 2018-05-22
None of the themes show any of the fancy icons, shown in the README.md


davidv171:
### Description

This is what it currently looks like. Honestly, just regular command prompt with 3 colored arrows at the start.

http://i.imgur.com/uvvyujC.png

### Expected behavior

I expected it to be as pretty as the picture from the:

https://camo.githubusercontent.com/612c6be742478221e06c64b66b896085f6e4af63/687474703a2f2f692e696d6775722e636f6d2f6e7247563670672e706e67

### Steps to Reproduce

1. Open zsh
2. Be disappointed

### Versions

zsh 5.5.1 (x86_64-solus-linux-gnu)

I'm sorry if this requires some further customizing, but I mostly only have the readme to go off of and some sketchy forks, that only broke the entire thing.

salmanulfarzy:
Icons displayed on the right side on the `README` image indicates various `git` status. They'll be displayed only when required. It depend on the [`git` module](https://github.com/sorin-ionescu/prezto/tree/master/modules/git#theming). Make sure you've that module enabled on your configuration. 

Also, Please properly format your images on issue description.

jeffwidman:
Closing now that #1606 is merged. @davidv171 if this doesn't solve your problem, let us know and we can re-open.

PR #1606, opened 2018-07-21
Add explanation for extra git theme symbols in README.md

clounie:
Helps with #1589

## Proposed Changes

  - Add explanation for how to make symbols display in prompt themes. For new users this is not clear at all - I only figured it out by searching the repository issues.


indrajitr:
Thank you very much @clounie!

PR #1611, opened 2018-08-01
command-not-found: Restore idiomatic loading of homebrewed handler on MacOS

### LINKED ISSUE #1451: command-not-found module is super slow
While working on https://github.com/sorin-ionescu/prezto/pull/1450, I noticed that enabling the `command-not-found` added ~2 seconds to shell startup time for an interactive session. This is on macOS using Homebrew.

However, directly calling `brew command command-not-found-init` seems to take only ~1 second.

Looking at https://github.com/sorin-ionescu/prezto/commit/b0c85eddf6bee4d75ca9920c9c6180ab7ff1962a, I'm wondering if the section:

    && brew command command-not-found-init > /dev/null 2>&1;

is actually running the command just to check its existence? If so, then this is running a very slow script twice on every shell startup...  Very possible I'm misreading this, as I'm very much a n00b when it comes to ZSH scripting.

@indrajitr am I reading this code correctly? Is it possible to change this to only run the script once?

### LINKED ISSUE #1577: Improve Prezto "command-not-found" module load time
If I disable the "`command-not-found`" module in my `zpreztorc` file the init time drops from 1.2s to 0.2s. It seems like this module takes a lot of time to initialize.

1) Are there any options to reduce this time
2) Can some of the initialization be delayed until after the prompt is available?

For reference I'm testing with the command:

    $ time (zsh -ic exit)

indrajitr:
## Proposed Changes

As is the convention in prezto, we cache the command-not-found handler to avoid
incurring the performance penalty of loading ruby interpreter on every call.
This restores the 'homebrew way' of loading command-not-found handler.

This also reinstates support for custom taps or non-standard homebrew location.

Fixes #1577, Fixes #1451

### DIFF
--- a/modules/command-not-found/init.zsh
+++ b/modules/command-not-found/init.zsh
@@ -12,9 +12,28 @@ if [[ -s '/etc/zsh_command_not_found' ]]; then
 # Load command-not-found on Arch Linux-based distributions.
 elif [[ -s '/usr/share/doc/pkgfile/command-not-found.zsh' ]]; then
   source '/usr/share/doc/pkgfile/command-not-found.zsh'
-# Load command-not-found on macOS when homebrew tap is configured.
-elif [[ -s '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-command-not-found/handler.sh' ]]; then
-  source '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-command-not-found/handler.sh'
+# Load command-not-found on macOS when Homebrew tap is configured.
+# To avoid performance penalty, we do not use Homebrew's ruby based command
+# lookup mechanism (viz., `brew command command-not-found-init`) and instead
+# `find` it ourselves from `TAP_DIRECTORY` defined internally in Homebrew.
+elif (( $+commands[brew] )); then
+  cnf_command="$(brew --repository 2> /dev/null)"/Library/Taps/homebrew/homebrew-command-not-found/cmd/brew-command-not-found-init.rb
+  if [[ -s "$cnf_command" ]]; then
+    cache_file="${TMPDIR:-/tmp}/prezto-brew-command-not-found-cache.$UID.zsh"
+
+    if [[ "$cnf_command" -nt "$cache_file" \
+          || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
+          || ! -s "$cache_file" ]]; then
+      # brew command-not-found-init is slow; cache its output.
+      brew command-not-found-init >! "$cache_file" 2> /dev/null
+    fi
+
+    source "$cache_file"
+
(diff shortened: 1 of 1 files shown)


indrajitr:
@belak, @cmcginty, @jeffwidman: mind giving this a shot?

jeffwidman:
@indrajitr I may be dense, but I'm not understanding how this is better than the current solution in https://github.com/sorin-ionescu/prezto/commit/f8f93d4b91ba59a3323baf0a1ed91c6c33da72b8? The current solution is simple, straightforward, and sufficiently speedy. Versus the code seems significantly more complex with this solution...

indrajitr:
f8f93d4b91ba59a3323baf0a1ed91c6c33da72b8 makes two explicit assumptions:

1. Homebrew is installed in `/usr/local/Homebrew` which might not be the case
2. The user is using the default homebrew-command-not-found tap which might not be the case either

modules/command-not-found/init.zsh
```diff
@@ -13,8 +13,19 @@ if [[ -s '/etc/zsh_command_not_found' ]]; then
 elif [[ -s '/usr/share/doc/pkgfile/command-not-found.zsh' ]]; then
   source '/usr/share/doc/pkgfile/command-not-found.zsh'
 # Load command-not-found on macOS when homebrew tap is configured.
-elif [[ -s '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-command-not-found/handler.sh' ]]; then
-  source '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-command-not-found/handler.sh'
+elif (( $+commands[brew] )) \
+        && [[ -s "${cnf_command::=$(brew command command-not-found-init 2> /dev/null)}" ]]; then
```
belak: This is the main downside right here - this call is painfully slow, taking almost a full second on my computer... and there isn't really a way to turn it off or cache.

jeffwidman: Yeah. It's fine if it's slow when I "miss" on a command... but not fine on every terminal instantiation...

indrajitr: Wow, ~1s just for `brew command` is rough and rather strange.
Running `time (brew command command-not-found-init)` gives me around 0.4s - 0.5s on a very slow non-retina MBP.

indrajitr:
@belak: Feel free to give this another shot. This should eliminate the ruby overhead.

indrajitr:
Now that `brew --repository` is a simple shell script call (without Ruby overhead), we don't have the performance impact anymore.

This still would support custom Homebrew installation in a non-default location, although we stick to default homebrew-command-not-found tap (for now).

belak:
This is awesome @indrajitr. There may be other places we should revisit calls to `brew` because of this.

PR #1616, opened 2018-08-13
Correct alphabetical order in zprestorc

newtonne:
I had to, sorry.


indrajitr:
Thank you @newtonne! Small things like these do matter :)

PR #1619, opened 2018-08-23
Awan

Awan:
This is a bash script to install prezto . It is a part of my dotfiles in which I have zprezto as submodule and once I clone it recursively and run this script, it install all other files with prezto. I'm gonna remove other enteries so it would become only prezto related.

### DIFF
--- a/install
+++ b/install
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+# This script will add all required setup files in theie place.
+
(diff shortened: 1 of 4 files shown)


belak:
Thanks for your contribution. Do you still plan on cleaning this up? It looks like there's quite a bit of personal changes in here.

PR #1620, opened 2018-08-23
Also check "commondir" when checking for a stash

hotpxl:
In `git-info`, we check if there is git stash by looking for `.git/refs/stash` file.
But in git worktrees, there is a file named `commondir` under `$(git-dir)`, and it's usually `../..`, and `refs/stash` is not in `$(git-dir)` but in `$(git-dir)/../..` (the `../..` part comes from the content of `commondir`).

This patch checks there as well.

### DIFF
--- a/modules/git/functions/git-info
+++ b/modules/git/functions/git-info
@@ -212,10 +212,16 @@ function git-info {
 
   # Format stashed.
   zstyle -s ':prezto:module:git:info:stashed' format 'stashed_format'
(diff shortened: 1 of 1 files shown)


hotpxl:
ok. i feel it's still important to check the file for performance reasons.

let me know if this'll do

hotpxl:
Giving this a little push..

belak:
Looks good to me! Thanks for submitting this! And sorry about the delay in reviewing.

PR #1622, opened 2018-09-15
homebrew: Load standard Homebrew environment variables into shell session

indrajitr:
## Proposed Changes

Load standard Homebrew environment variables into shell session.
Since `brew shellenv` is relatively new, there is extra guard. This can be removed a few months later.


belak:
LGTM! 👍

PR #1626, opened 2018-09-17
Add a basic fzf module

ashishgandhi:
`fzf` is a very good fuzzy history completer among other things. Add a basic module to support it. Sets up a place to add more features from it in the future if we decide to.

### DIFF
--- a/modules/fzf/README.md
+++ b/modules/fzf/README.md
@@ -1,0 +1,13 @@
+fzf
+===
+
+Sets up fzf key bindings.
+
+Authors
+-------
+
+*The authors of this module should be contacted via the [issue tracker][1].*
+
+  - [Ashish Gandhi](https://github.com/ashishgandhi)
+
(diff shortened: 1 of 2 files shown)


shwanton:
👍

belak:
Thanks for your contribution. Unfortunately, we're currently avoiding new modules in the prezto core. However, I'd be happy to merge something like this in the contrib repo. https://github.com/belak/prezto-contrib

belak:
Hi! Thanks for submitting this! I like the idea but we're also not really accepting new modules or prompts in the main repo. If you're still interested in getting this in, please feel free to submit it to the contrib repo (https://github.com/belak/prezto-contrib) and I'd be happy to merge it as a starting point for an fzf module.

ashishgandhi:
I switch to managing my own Zsh now no longer use `prezto`. I am happy if someone else wants to run with this. It's not like super novel anything.

Stay away from COVID-19.

PR #1633, opened 2018-10-13
Ability to set a custom PROMPT_EDGE 

twashing:
Using variable PROMPT_EDGE to set the desired prompt character

ex: 
```
export PROMPT_EDGE="λ"
```

### DIFF
--- a/modules/prompt/functions/prompt_steeef_setup
+++ b/modules/prompt/functions/prompt_steeef_setup
@@ -85,9 +85,10 @@ function prompt_steeef_setup {
   zstyle ':prezto:module:python:info:virtualenv' format '(%v)'
 
   # Define prompts.
+  PROMPT_EDGE=${PROMPT_EDGE:=\$}
   PROMPT="
 ${_prompt_steeef_colors[3]}%n%f at ${_prompt_steeef_colors[2]}%m%f in ${_prompt_steeef_colors[5]}%~%f "'${vcs_info_msg_0_}'"
-"'$python_info[virtualenv]'"$ "
+"'$python_info[virtualenv]'"${PROMPT_EDGE} "
   RPROMPT=''
 }
 


modules/prompt/functions/prompt_steeef_setup
```diff
@@ -85,9 +85,10 @@ function prompt_steeef_setup {
   zstyle ':prezto:module:python:info:virtualenv' format '(%v)'
 
   # Define prompts.
+  PROMPT_EDGE=${PROMPT_EDGE:=\$}
```
indrajitr: Please consider using `zstyle` based configuration.

modules/prompt/functions/prompt_steeef_setup
```diff
@@ -85,9 +85,10 @@ function prompt_steeef_setup {
...
   PROMPT="
 ${_prompt_steeef_colors[3]}%n%f at ${_prompt_steeef_colors[2]}%m%f in ${_prompt_steeef_colors[5]}%~%f "'${vcs_info_msg_0_}'"
-"'$python_info[virtualenv]'"$ "
+"'$python_info[virtualenv]'"${PROMPT_EDGE} "
```
indrajitr: Instead of using `${PROMPT_EDGE}`, consider using the standard `${editor_info[keymap]}` so that the configuration can be exposed via `zstyle`.

twashing: Hey @indrajitr , 

Thanks for the feedback. I'm new to zsh and prezto. So I don't know how the `zstyle ':prezto:module:...` and `${editor_info[keymap]}` mechanisms work. 

I'm combing through [theming](https://github.com/sorin-ionescu/prezto/tree/master/modules/editor#theming) and [examples](https://github.com/sorin-ionescu/prezto/blob/master/runcoms/zpreztorc). And if I'm understanding the examples correctly, this PR isn't needed. If it is needed, I'm not clear on where the `PROMPT_EDGE` would go?

johnpneumann:
Thanks for following up on this @twashing. :) See the PR from @indrajitr

indrajitr:
Fixed in #1634

PR #1638, opened 2018-11-06
Also lookup on ~/.ssh/config.d/* files when doing ssh host complete

rjcoelho:
Because I use
Include ~/.ssh/config.d/*
on my ~/.ssh/config

Anyone knows how to put this into a SSH_CONFIG env var ?

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -109,7 +109,7 @@ zstyle -a ':prezto:module:completion:*:hosts' etc-host-ignores '_etc_host_ignore
 zstyle -e ':completion:*:hosts' hosts 'reply=(
   ${=${=${=${${(f)"$(cat {/etc/ssh/ssh_,~/.ssh/}known_hosts(|2)(N) 2> /dev/null)"}%%[#| ]*}//\]:[0-9]*/ }//,/ }//\[/ }
   ${=${(f)"$(cat /etc/hosts(|)(N) <<(ypcat hosts 2> /dev/null))"}%%(\#${_etc_host_ignores:+|${(j:|:)~_etc_host_ignores}})*}
-  ${=${${${${(@M)${(f)"$(cat ~/.ssh/config 2> /dev/null)"}:#Host *}#Host }:#*\**}:#*\?*}}
+  ${=${${${${(@M)${(f)"$(cat ~/.ssh/config ~/.ssh/config.d/* 2> /dev/null)"}:#Host *}#Host }:#*\**}:#*\?*}}
 )'
 
 # Don't complete uninteresting users...


indrajitr:
We probably have to scan `Include` in `~/.ssh/config` instead of blindly assuming `~/.ssh/config.d/*`. What do you think @rjcoelho?

rjcoelho:
> We probably have to scan `Include` in `~/.ssh/config` instead of blindly assuming `~/.ssh/config.d/*`. What do you think @rjcoelho?

It's going to slow down the startup time, plus it doesn't really change on every login. Its something done once on "setup". How about an env var SSH_CONFIG=~/.ssh/config.d/* ?

indrajitr:
> > We probably have to scan `Include` in `~/.ssh/config` instead of blindly assuming `~/.ssh/config.d/*`. What do you think @rjcoelho?
> 
> It's going to slow down the startup time, plus it doesn't really change on every login. Its something done once on "setup". How about an env var SSH_CONFIG=~/.ssh/config.d/* ?

Hmm, let me think about this a bit more. I'm hesitant to add yet another magic environment variable.

If at all, `zstyle` based configuration would be preferred instead of an environment variable.

rjcoelho:
> > > We probably have to scan `Include` in `~/.ssh/config` instead of blindly assuming `~/.ssh/config.d/*`. What do you think @rjcoelho?
> > 
> > 
> > It's going to slow down the startup time, plus it doesn't really change on every login. Its something done once on "setup". How about an env var SSH_CONFIG=~/.ssh/config.d/* ?
> 
> Hmm, let me think about this a bit more. I'm hesitant to add yet another magic environment variable.
> 
> If at all, `zstyle` based configuration would be preferred instead of an environment variable.

Yes I agree, since AFAIK zstyle is the preferred way of configuring in zsh. Anyone knows how to do it using zstyle ?

diraol:
Hey folks, not sure if this has happened or will happen somewhere in the future, but I just want to add to the discussion that while using `Include` I usually have multiple config files separated across many different paths (I do it for gitconfig too), one for each "context".

So, for each independent project I'm part of I usually have a `~/.<project>` path with configs for it (gitconfig, ssh, aliases, env vars, etc).

So, if we are going with zstyle config, it would be nice to have the ability to setup multiple paths.

srijanshetty:
Having `Include config.d/*` as the first line on my `.ssh/config` automatically completes all the hosts present in `config.d/*`.

srijanshetty:
> Hey folks, not sure if this has happened or will happen somewhere in the future, but I just want to add to the discussion that while using `Include` I usually have multiple config files separated across many different paths (I do it for gitconfig too), one for each "context".
> 
> So, for each independent project I'm part of I usually have a `~/.<project>` path with configs for it (gitconfig, ssh, aliases, env vars, etc).
> 
> So, if we are going with zstyle config, it would be nice to have the ability to setup multiple paths.

I have a similar setup but I symlink all files to `.ssh/config.d/` and autocomplete works automatically.

srijanshetty:
Closing this as zsh supports Include directive natively [now](https://github.com/zsh-users/zsh/commit/5ded0ad96740b62ea0214387ee260b515726a05e).

PR #1640, opened 2018-11-08
Update submodules

diraol:
Just updating submodules versions ;)

Updating to the latest released tags for modules:
completetion
syntax-highlighting

And for agnoster prompt


indrajitr:
Thanks again @diraol, for doing this!

PR #1642, opened 2018-11-16
add store passphrases in your keychain. only macOS

und3fined:
update for macOS, never ask passphaeses after save to keychain
---
https://apple.stackexchange.com/a/250572


johnpneumann:
I like the idea, but this needs to be optional behavior, not the default. Can you provide a configuration option for this. If unset, it defaults to not using the keychain. If set, then it does whatever the value is set to. (As an [example option](https://github.com/sorin-ionescu/prezto/blob/master/runcoms/zpreztorc#L152) of [how it's handled](https://github.com/sorin-ionescu/prezto/blob/master/modules/editor/init.zsh#L412-L414))

und3fined:
yeah, i'll add a option for this module

johnpneumann:
Awesome. Thank you!

PR #1648, opened 2018-11-30
Add alias `z` provided by `fasd`, which makes `z` users happy without ditching `fasd`.

### LINKED ISSUE #656: Readd z

tevino:
Fixes #656 (hopefully)

## Proposed Changes

  - Add an alias `z` for non-interactive directory changing.


tevino:
@sorin-ionescu PTAL

PR #1650, opened 2018-12-09
python: Make cached completion file mangling more reliable

indrajitr:
## Proposed Changes

While mangling cached completion file, we cannot just assume that
`$pip_command` would resolve to `pip` -- it might be `pip2` or `pip3`
depending on the relative position in zsh `$commands` array.  Thus replace
the whole of 'pip*' with 'pip pip2 pip3' for compctl assignment.

Also, use more apropriate cache filename for pip completion.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -150,7 +150,7 @@ fi
 
 # Load PIP completion.
 if (( $#commands[(i)pip(|[23])] )); then
-  cache_file="${TMPDIR:-/tmp}/prezto-python-cache.$UID.zsh"
+  cache_file="${TMPDIR:-/tmp}/prezto-pip-cache.$UID.zsh"
 
   # Detect and use one available from among 'pip', 'pip2', 'pip3' variants
   pip_command="$commands[(i)pip(|[23])]"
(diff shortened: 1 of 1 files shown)


modules/python/init.zsh
```diff
@@ -160,10 +160,11 @@ if (( $#commands[(i)pip(|[23])] )); then
         || ! -s "$cache_file" ]]; then
     # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
     $pip_command completion --zsh \
-      | sed -e "s|compctl -K [-_[:alnum:]]* pip|& pip2 pip3|" >! "$cache_file" 2> /dev/null
+      | sed -e "s|\(compctl -K [-_[:alnum:]]*\) pip|\1 pip pip2 pip3|" >! "$cache_file" 2> /dev/null
```
belak: What does this regex even do?

indrajitr: Replaces the line `compctl -K _pip_completion pip` with `compctl -K _pip_completion pip pip2 pip3` in the output buffer of `pip completion --zsh`.

This snippet will probably make it a bit more obvious:

```shell
~❯❯❯ diff <(pip completion --zsh) <(pip completion --zsh | sed -e "s|\(compctl -K [-_[:alnum:]]*\) pip|\1 pip pip2 pip3|")
11c11
< compctl -K _pip_completion pip
---
> compctl -K _pip_completion pip pip2 pip3
```

PR #1651, opened 2018-12-09
add "show/hide hidden files" functions

vanyakosmos:
## Proposed Changes

Add `showfiles` and `hidefiles` functions to osx module.


mattmc3:
I know this is an old PR, but I'd love to see it merged. OMZ has something similar: https://github.com/ohmyzsh/ohmyzsh/blob/fc40b53e6460560ed9b256deb87f2165f8d48f1f/plugins/macos/macos.plugin.zsh#L10-L11

PR #1653, opened 2018-12-11
rsync: use --inplace for rsync base cmd by default

indrajitr:
## Proposed Changes

  - Use `--inplace` flag by default. Updating detination file 'inplace' is almo st alway what we want. This is also a recommended option on COW filesystems (like BTRFS). @Eriner suggested this [few years ago](https://github.com/Eriner/prezto/commit/2c3e9209726652fe32e6c82e1fd167517f13476f) and I have been using it since quite successfully.
  - Update link to Bombich rsync page again

### DIFF
--- a/modules/rsync/init.zsh
+++ b/modules/rsync/init.zsh
@@ -22,7 +22,7 @@ if grep -q 'xattrs' <(rsync --help 2>&1); then
 fi
 
 # macOS and HFS+ Enhancements
-# http://help.bombich.com/kb/overview/credits#opensource
+# https://bombich.com/kb/ccc5/credits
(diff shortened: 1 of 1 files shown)


belak:
I'm not 100% sure this should be the default... doesn't `--inplace` change from using a tempfile and moving it into place to overwriting the file? This means if you have a file that's changed and rsync fails part way through you could be left with a broken file. It also means that processes accessing that file might be accessing a file that's being changed. I'm open to suggestions, but I'd prefer to err on the side of safety.

indrajitr:
Agreed very much. Another issue I see is that the flag might not be available in some old `rsync` implementation.

Let me think about it a bit. Will probably close this if something better doesn't come to my mind.

indrajitr:
@belak, @johnpneumann: You guys are right, this isn't worth being a default.

PR #1655, opened 2018-12-13
node: Make nvm lookup mechanism more efficient in homebrewed environment

indrajitr:
## Proposed Changes

In homebrewed environment, avoid using `brew --prefix nvm` which is ruby based and is super slow. Instead, rely on homebrew standard behavior wherein all installed packages are available in canonical path `$(brew --prefix)/opt/<package>` (for nvm it would obviously be `$(brew --prefix)/opt/nvm`).

NB: `$(brew --prefix)` (without additional argument) is a simple shell shortcut and doesn't have the same performance impact.


indrajitr:
Also see #1611 for related discussion.

PR #1126, opened 2016-04-14
Add Signed commits

IhabSoliman:
Add extra commit alias to sign the commits with GPG key

PR #1657, opened 2018-12-16
Remove duplicated information from git module README

diraol:
This was added in #1126 while updating the README.


belak:
Good catch. Thanks!

PR #1663, opened 2018-12-22
Add aliases for "brew services"

citizen428:
## Proposed Changes

Just adding some simple aliases for managing `brew services`.

### DIFF
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -14,7 +14,7 @@ fi
 # Variables
 #
 
-# Load standard Homebrew shellenv into the shell session. 
+# Load standard Homebrew shellenv into the shell session.
 # Load 'HOMEBREW_' prefixed variables only. Avoid loading 'PATH' related
 # variables as they are already handled in standard zsh configuration.
 if (( $+commands[brew] )); then
@@ -34,6 +34,12 @@ alias brewo='brew outdated'
 alias brews='brew search'
 alias brewu='brew upgrade'
 alias brewx='brew uninstall'
+
+# Homebrew Services
+alias bservl='brew services list'
+alias bservr='brew services restart'
+alias bservs='brew services start'
+alias bservx='brew services stop'
 
 # Homebrew Cask
 alias cask='brew cask'


modules/homebrew/init.zsh
```diff
@@ -35,6 +35,12 @@ alias brews='brew search'
 alias brewu='brew upgrade'
 alias brewx='brew uninstall'
 
+# Homebrew Services
+alias bservl='brew services list'
+alias bservr='brew services restart'
+alias bservs='brew services start'
+alias bservx='brew services stop'
+
```
indrajitr: While I like the idea, I am still on the fence regarding the choice of alias prefix.

I am not sure if we should keep using `brew<...>` or introduce yet another set `bserv<...>`.

citizen428: I'm happy to go with whatever, `brewsl`, `brewsr` etc. would work too.

indrajitr: Sure @citizen428! `brews` is already taken by `brew search`, so something different might be preferred.

Maybe `brewSl`, `brewSr` etc. might be considered. But then again, we are using lower `s` and upper `S` to distingush between completely unrelated things.

Which is why I am not exactly sure (and on the fence) what might be a better convention. Let's see what others have to say.

citizen428: It seems no one has to say anything, so maybe let's just close this?

belak: I agree with @indrajitr that the prefix is a little odd, but I'd still be fine including this if you're still interested in the change.

citizen428: I personally don't use Prezto anymore, so it's up to you really. But if you prefer another prefix I can update the PR as a parting gift 😃

jeffwidman: IMO let's keep `brew` as the prefix for all Homebrew related commands. If someone types it so frequently that they want something shorter, they can add it to their personal setup.

jeffwidman:
I'm going to close as I don't think this has much traction for now... If someone is interested in pushing across the finish line, I think we'd be open to it, with the caveat I mentioned here: https://github.com/sorin-ionescu/prezto/pull/1663#discussion_r376077667

PR #1668, opened 2019-01-03
Fix typo: gupl --> gulp

jeffwidman:
I _think_ this is a typo, but not sure as I don't actually use `gulp`... just spotted while reviewing the diff of latest changes.


belak:
Definitely a typo. Good catch.

PR #1669, opened 2019-01-04
history: remove mutually exclusive option

hotpxl:
According to Zsh documentation http://zsh.sourceforge.net/Doc/Release/Options.html#History
INC_APPEND_HISTORY, INC_APPEND_HISTORY, and SHARE_HISTORY are mutually exclusive.
I didn't see any bug specifically, but just to be safe
I figured we want SHARE_HISTORY since it has the most features (and is what we actually use)

### DIFF
--- a/modules/history/README.md
+++ b/modules/history/README.md
@@ -8,8 +8,9 @@ Options
 
   - `BANG_HIST` treats the **!** character specially during expansion.
   - `EXTENDED_HISTORY` writes the history file in the *:start:elapsed;command* format.
-  - `INC_APPEND_HISTORY` writes to the history file immediately, not when the shell exits.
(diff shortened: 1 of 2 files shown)


jeffwidman:
Great catch!

hotpxl:
I believe this actually results in a different behavior :(
Can't really trust the documentation. I'm still testing and see why/what is happening

AvverbioPronome:
I ended up investigating this, since it collates live all my shells, making me press arrowup like an idiot that doesn't know ctrl-r. I just kept the plugin enabled and added two lines to my `.zshrc`:

```
unsetopt SHARE_HISTORY
setopt INC_APPEND_HISTORY
```

This way history is still shared across time, but multiple shells opened at the same time won't interfere with each other.

PR #1670, opened 2019-01-04
Add option to disable pip cache generation

hotpxl:
On my computer, this process takes seconds. And as I ssh onto a lot of "fresh" hosts, it's very annoying as the initial logon is very slow.

Adding this option so user could disable. But by default it's on, as before

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -22,6 +22,12 @@ zstyle ':prezto:module:python' conda-init 'on'
 
 Caution: using conda and virtualenvwrapper at the same time may cause conflicts.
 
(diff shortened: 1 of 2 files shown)


indrajitr:
Let's hold off on this one, please. I'm working on a different strategy to avert eagerly invoking cache generation across prezto.

hotpxl:
Sure, it would be really nice to have it do the caching on first pip invocation, and not always.

But could we have some form of workaround in the mean time?

hotpxl:
Hi, has there been any progress on this?

jeffwidman:
nudge @indrajitr

PR #1385, opened 2017-07-18
Migrate sorin prompt to zsh-async

belak:
Fixes #1384 and is a final tweak for the security issue relating to "pw3nage".

## Proposed Changes

  - Add zsh-async as an external module
  - Move all git-info checking in the sorin prompt to zsh-async
  - Ensure that pw3nage does not trigger with this method

PR #1671, opened 2019-01-06
sorin-prompt: Remove old prompt tempfile and pid variable

fd0:
## Proposed Changes

  - Remove the tempfile and PID variable for the sorin prompt, it's not needed any more. Probably a leftover from #1385 

CC @belak


indrajitr:
Good catch, thanks @fd0!

PR #1673, opened 2019-01-07
Update syntax-highlighting submodule

diraol:
Hi folks!

I'm sending this update in a separated PR because I think a decision needs to be made in regard to this submodule update policy.

Prezto usually take submodules TAGs into account, which means that if there are tags, we should use them.
But for this specific module (`syntax-highlighting`), the developers seems not to be tagging it anymore, since August/2017, and the development seems to be taking its courses normally.

So, my question is:
  - Do we keep tracking the TAGs (more than 1 1/2 year behind the master branch);  or
  - Do we start tracking the master branch?


indrajitr:
Thank you @diraol!

Issue #1164, opened 2016-07-04
gpg-agent: a gpg-agent is already running


ephexeve:
I have enabled gpg in my `.zpreztorc`, and whenever I start a new terminal, I get the following message: `gpg-agent: a gpg-agent is already running - not starting a new on`

I am currently using a Smartcard with a PGP Authentication key, which should be loaded when running `ssh-add -L` when having GPG module is enabled. This works, however, after some time, the ssh key disappears and I need to kill ssh-agent, gpg-agent to work again. 

```
$ ps aux | grep gpg
ephexeve    20570  0.0  0.0  91828  1320 ?        Ss   14:36   0:00 gpg-agent --daemon
$ ps aux | grep ssh
$ echo $SSH_AGENT_PID
19199
$ echo $SSH_AUTH_SOCK
/tmp/ssh-agent.sock
$ ssh-add -L
ssh-rsa <long output>
# closes terminal and opens a new one...
$ ssh-add -L
The agent has no identities
$ echo $SSH_AGENT_PID
21855
$ echo $SSH_AUTH_SOCK
/tmp/ssh-agent.sock
$ ps aux | grep ssh
ephexeve    21855  0.0  0.0  11164   316 ?        Ss   14:37   0:00 ssh-agent
 $ ps aux | grep gpg
ephexeve    20570  0.0  0.0 165560  1376 ?        Ss   14:36   0:00 gpg-agent --daemon
```
My `gpg-agent.conf`:

```
pinentry-program /usr/bin/pinentry-gnome3
default-cache-ttl 18000
enable-ssh-support
```

Running on Archlinux Kernel  4.6.3-1-ARCH gpg version 2.1.13

jasonmhite:
Same issue here, I'm willing to bet it has to do with this issue on the GPG tracker...

https://bugs.gnupg.org/gnupg/issue2258

ephexeve:
I fixed it by adding the following to my `.zshenv`
(Still get the annoying `gpg-agent: a gpg-agent is already running - not starting a new one` though.

```
# Set SSH to use gpg-agent
unset SSH_AGENT_PID
if [ "${gnupg_SSH_AUTH_SOCK_by:-0}" -ne $$ ]; then
  export SSH_AUTH_SOCK="/run/user/$UID/gnupg/S.gpg-agent.ssh"
fi
```

Change the `SSH_AUTH_SOCK` path accordingly.

belak:
[...]

PR #1682, opened 2019-01-26
Fix gpg module condition 

### LINKED ISSUE #1681: GPG module output agent already running on NixOS
### Description
1. On NixOS, gpg don't export `GPG_AGENT_INFO` variable and `S.gpg-agent` file is under `/run/user/<uid>/gnupg/S.gpg-agent`, not `~/.gnupg/`.

2. grep cannot match correctly. 
[prezto-repo about gpg init](https://github.com/sorin-ionescu/prezto/blob/master/modules/gpg/init.zsh#L23)
Because of first point, the second if-condition about grep, cannot match gpg-agent process, but the process exists.

Output `echo ${${${(s.:.)GPG_AGENT_INFO}[2]}:--1} gpg-agent`: `-1 gpg-agent`
Output  `ps -U "$LOGNAME" -o pid,ucomm | grep " gpg-agent` : `29385 gpg-agent`

One way to solve is to use `gpgconf` command I think.

By the way, possibly, there are other system having the same problem. I think we can find another way to decide whether gpg-agent is run or not, or try to fix the second grep regex expression to let it fix all system. 

### Expected behavior
Output: `gpg-agent: a gpg-agent is already running - not starting a new one`

### Actual behavior
No output if it is started.

### Steps to Reproduce
NixOS only.  Reproduce may be done by installing gpg with `nix`.  

### Versions

  - Prezto commit:master
  - OS information: NixOS 18.09

Vonfry:
Fixes #1681 

## Proposed Changes

  - gpg module: if-condition ( grep expression and find socket file) in modules/gpg/init.zsh between line  16 to 25.

### DIFF
--- a/modules/gpg/init.zsh
+++ b/modules/gpg/init.zsh
@@ -11,9 +11,9 @@ if (( ! $+commands[gpg-agent] )); then
 fi
 
 # Set the default paths to gpg-agent files.
-_gpg_agent_conf="${GNUPGHOME:-$HOME/.gnupg}/gpg-agent.conf"
+_gpg_agent_conf="$(gpgconf --list-dir homedir)/gpg-agent.conf"
 _gpg_agent_env="${TMPDIR:-/tmp}/gpg-agent.env.$UID"
-_gpg_agent_socket="$(gpgconf --list-dir | grep "agent-socket:.*" | sed "s/agent-socket://")"
+_gpg_agent_socket="$(gpgconf --list-dir agent-socket)"
 
 # Load environment variables from previous run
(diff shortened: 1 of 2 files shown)


Vonfry:
Add more: remove re-export `SSH_AUTH_SOCK` in ssh module. Becasue some tools such as pam ssh auth will check socket path. A link to this socket is useful but re-export the variable will break some tools.

edubenetskiy:
This pull request seems to fix #1164 for me.

Vonfry:
close pr for fixing this in another branch  not master

PR #1687, opened 2019-02-10
Add the ability to set locale environment variables

N4M3Z:
Allow setting locale environment variables `$LC_ALL` and `$LANG` from prezto runcom.

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -62,3 +62,14 @@ if zstyle -t ':prezto:environment:termcap' color; then
   export LESS_TERMCAP_ue=$'\E[0m'          # Ends underline.
   export LESS_TERMCAP_us=$'\E[01;32m'      # Begins underline.
 fi
+
+#
+# Locale
+#
+
+zstyle -s ':prezto:environment' language 'LANGUAGE'
+if [[ ! -z $LANGUAGE ]]
+then
+  export LC_ALL="$LANGUAGE"
+  export LANG="$LANGUAGE"
+fi
--- a/runcoms/zpreztorc
+++ b/runcoms/zpreztorc
@@ -66,6 +66,13 @@ zstyle ':prezto:module:editor' key-bindings 'emacs'
 #zstyle ':prezto:module:editor' ps-context 'yes'
 
 #
+# Environment
+#
+
+# You may need to manually set your language environment
+zstyle ':prezto:environment' language 'en_US.UTF-8'
+
+#
 # Git
 #
 


N4M3Z:
Hi Prezto team. Is there a particular reason why this trivial contribution has not been merged yet?

belak:
In general LC_ALL should not be set because it overrides everything. Setting LANG should be enough... And if that's the case, it seems like that could just go in the runcoms directly

N4M3Z:
As it has been multiple months I don't remember exactly now, but I had to explicitly set `LC_ALL` as well to avoid `Python locale error: unsupported locale setting` on my setup. We can remove it from the condition and perhaps add a zstyle to set `LC_*` separately, probably somewhat like this:
```
zstyle ':prezto:environment:language' all 'en_US.UTF-8'
zstyle -s ':prezto: environment:language' all 'LC_ALL'
```
What do you think?

belak:
Yeah, the language settings can be confusing.

The `locale` command can be used to debug how the system interprets the environment variables (though some languages may not respect all these). My guess is that the language for your system was set to something that didn't exist. There's a ton of background [here](https://unix.stackexchange.com/a/87763) if you're interested in details.

Also, I played around with it a bit - python's getlocale isn't as useful as I'd like because it doesn't respect the LANG variable and just directly checks LC_ALL (by default). It looks like [getdefaultlocale](https://docs.python.org/2/library/locale.html#locale.getdefaultlocale) is more useful (if you're writing your own programs) as it will properly respect the locale variables.

There's also some background info in those docs which state "`According to POSIX, a program which has not called setlocale(LC_ALL, '') runs using the portable 'C' locale. Calling setlocale(LC_ALL, '') lets it use the default locale as defined by the LANG variable. Since we do not want to interfere with the current locale setting we thus emulate the behavior in the way described above.`

Thanks for submitting this, but I'd still prefer to leave this out (at least the portion relating to LC_ALL because it overrides other language settings)... and this seems like it would be better solved by simply setting these in the runcoms directly if you need them.

jeffwidman:
@belak's comments make sense to me, shall we close this?

@N4M3Z what do you think? If you are aware of more facts re: language settings, I and the other maintainers would be happy to hear them and reconsider, aside from @belak I suspect not many of us know very much about these...

N4M3Z:
I don't understand what's the problem with having `LANG ` and `LC_ALL` __**optionally**__ setable by the runcom using the zstyle convention. We can simply remove the default:
```
zstyle ':prezto:environment' language 'en_US.UTF-8'
```
and then everything being optional, I don't see the harm.

belak:
Setting `LC_ALL` isn't a great practice to get into because it overrides a ton of locale variables, even the default environment in which programs start. Many people use this because it allows you to force the locale to a specific setting, but it's more of a blunt instrument. This can be what's intended, but as many people assume a project like prezto will support best practices, I'd rather not provide a built-in way to override it. 

Also, any user can set these variables in their zshrc or zprofile. Setting it using zstyle adds additional code (and this is additional code which needs to be maintained) without making the system more flexible or convenient.

I'm open to other suggestions if there's a strong reason that can be backed up with documentation and/or reasoning, or if another maintainer disagrees with me here, but my current opinion is that while this definitely can solve some issues, this generally fits better when set manually by the user.

N4M3Z:
As you wish, I'll just use my fork.

belak:
Out of curiosity, why do you view this as better than adding the export calls directly in zpreztorc or your zshrc?

N4M3Z:
It's just a preference. In my current setup I configure everything through zstyle, I've expanded the options to suit my needs where prezto hasn't implemented it.

johnpneumann:
FWIW, you don't need a fork to have this accessible. I'm with @belak on this.

PR #1689, opened 2019-02-16
archive: enhance parallel operations

maximbaz:
This adds a few improvements to the way `archive` module works:

1. Support `pixz` as the parallel version of `xz` when creating an archive
1. Support `lbzip2` as a better alternative to `pbzip2` (`lbzip2` works twice as fast in my tests)
1. Make `unarchive` helper also use parallel alternatives, and make it behave exactly the opposite to `archive` helper in terms of how commands are executed.

### DIFF
--- a/modules/archive/functions/archive
+++ b/modules/archive/functions/archive
@@ -8,7 +8,7 @@
 
 # function archive {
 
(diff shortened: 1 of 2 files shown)


belak:

PR #1695, opened 2019-03-13
Add Powerlevel10k theme

romkatv:
[Powerlevel10k](https://github.com/romkatv/powerlevel10k) is a 10-100 faster reimplementation of the popular Powerlevel9k theme. It renders the same prompt when given the same configuration options but does it faster. If an existing user of powerlevel9k replaces `zstyle :prezto:module:prompt theme powerlevel9k` in their `.zpreztorc` with `zstyle :prezto:module:prompt theme powerlevel10k`, their prompt will get blazingly fast while still looking the same as before.

Obligatory screenshot:

Powerlevel10k

This PR was tested in a pristine Ubuntu environment with prezto:

```zsh
docker run -e LANG=C.UTF-8 -e LC_ALL=C.UTF-8 -e TERM=$TERM -it --rm ubuntu bash -uexc '
  apt update && apt install -y zsh git
  git clone --recursive -b powerlevel10k https://github.com/romkatv/prezto.git ~/.zprezto
  zsh -ec "
    setopt EXTENDED_GLOB
    for rcfile in ~/.zprezto/runcoms/^README.md(.N); do
      ln -s \$rcfile ~/.\${rcfile:t}
    done
  "
  git clone https://github.com/belak/prezto-contrib ~/.zprezto/contrib
  cd ~/.zprezto/contrib
  git submodule init
  git submodule update
  echo "zstyle :prezto:module:prompt theme powerlevel10k" >> ~/.zpreztorc
  echo "
    # Your prompt configuration goes here.
    POWERLEVEL9K_LEFT_PROMPT_ELEMENTS=(root_indicator dir_writable dir vcs)
    POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS=(status command_execution_time background_jobs time)
    POWERLEVEL9K_PROMPT_ON_NEWLINE=true
  " >> ~/.zshrc
  cd ~/.zprezto  # more fun here because it is a git repo
  zsh -i
'
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "modules/prompt/external/powerlevel9k"]
 	path = modules/prompt/external/powerlevel9k
 	url = https://github.com/bhilburn/powerlevel9k.git
+[submodule "modules/prompt/external/powerlevel10k"]
+	path = modules/prompt/external/powerlevel10k
+	url = https://github.com/romkatv/powerlevel10k.git
--- a/modules/prompt/functions/prompt_powerlevel10k_setup
+++ b/modules/prompt/functions/prompt_powerlevel10k_setup
@@ -1,0 +1,1 @@
+../external/powerlevel10k/powerlevel10k.zsh-theme


belak:
Thanks for your submission! I've been watching the development of this and was curious if someone would submit it eventually. Unfortunately, we're not accepting new themes in prezto core, but I'd be happy to merge this into https://github.com/belak/prezto-contrib.

romkatv:
> Unfortunately, we're not accepting new themes in prezto core...

Oh, it's a shame. Your users get an offer to use powerlevel9k while there is a strictly better option. FWIW, if you repoint powerlevel9k submodule URL to powerlevel10k, all user configs will still work.

> I'd be happy to merge this into https://github.com/belak/prezto-contrib.

I'll take what I can. What do I need to do? Send another PR against that repo?

belak:
Yeah, one of the things I've been strongly considering is moving *all* external themes to contrib... then people who want to use just the core can easily do so.

Yep, just submit a PR against that repo. There are already a number of themes in the contrib-themes module. There should be instructions in the repo root for getting contrib working with prezto.

Teknitix:
@belak: It would be wonderful to add Powerlevel10k to the core -- especially since it's highly optimized, is a drop-in replacement for Powerlevel9k, and is becoming very popular: https://www.reddit.com/r/zsh/comments/ay6ny7/powerlevel10k_goes_instant_50_times_faster_than/

belak:
I'm probably going to catch flak for this... and it's a slippery slope adding *any* new themes to this repo... however, one of my original worries is that this was going to be a temporary project. That doesn't seem to be the case. You've shown that you're up for continuing to maintain this, so I'm willing to merge this in.

Thanks for your patience and for submitting this in the first place.

belak:
To anyone else submitting a theme who stumbles across this issue: there would need to be a *very* good reason for inclusion. The maintainers of pl10k have given multiple valid reasons - this is a faster reimplementation of another theme already included with a fair amount of support behind it. I just want to be clear that this is an exception to the rule and not the new norm.

PR #1697, opened 2019-03-18
Disable node-info output when value is system.

mkcode:
This makes the behavior consistent with rbenv and pyenv.


indrajitr:
Thanks @mkcode!

PR #1703, opened 2019-04-04
Allow tmux session name to be passed as env var

alerque:
Note this is a replacement for the obsolete PR #952.

I finally got this actually figured out and rebased on master. The priority is now:

1. A manual environment variable (great for keyboard shortcut launches, remote ssh connections, etc.)
2. The value of the zstyle setting
3. A default string of "prezto"
----
For normal usage this change should be a NOOP. However it enables the
use case where the user might want to predefine what session gets
connected to when the shell is launched. Running `zsh` will still net
you a connection to the last used tmux session or a new one named
"zprezto" if none exists. However `env tmux_session=myname zsh` will
connect you specifically to a session called 'myname' including creating
it if it doesn't exist.

This is particularly useful in window managers where a key binding might
open a terminal with a predefined session or even show a list of current
sessions and open a new terminal connected to that session.

The default session name may still be set using zstyle as documented.


jeffwidman:
👍 on the general idea here. 

I had a hard time stepping through the code... have you tested all these scenarios?

Also, your PR has a great explanation... maybe copy some of that to the `tmux` module `readme`?

alerque:
Yes I've tested all the scenarios mentioned. I've been using this since 2015 and use all four scenarios on a daily basis. I had to re-code the logic a little bit to keep up with prezto changes, but the basic idea is the same.  The code isn't that complicated to step through:

1. If the var (trimmed) is empty, then try to set it from zstyle. If it's still empty set it to the default string.
2. Use the var.
3. Explicitly use the var rather than letting tmux guess in case the requested session isn't the default.

I'm on the road right now but will try to add docs at some point.

jeffwidman:
nudge re: adding docs as previously discussed

alerque:
Thanks @jeffwidman, you caught me at a computer this time (since the issues isn't assigned to me it wasn't really staying on my radar!). I've added some documentation to the README and stuffed it in the same commit.

jeffwidman:
Thank you!

So stepping through this, I'm a little confused. You said:

> For normal usage this change should be a NOOP... Running zsh will still net
you a connection to the last used tmux session or a new one named
"zprezto" if none exists.

However, reading the code and the comments, it looks like the fallback behavior _does_ change from the last session session used to instead using the "prezto" session name.

If someone doesn't set this var, and last used a different session name, will this change the behavior for them to instead start attaching to the "prezto" name?

I don't have an opinion yet whether that is a good or bad thing--it might be an improvement it might not, but I first want to make sure I understand how this works.

alerque:
@jeffwidman Thanks for your patience. My logic did in fact slip a cog somewhere, this needs to be a 4 stage fallback not a 3 stage fallback.

1. env variable
2. last used
3. user set zstyle default
4. hard coded 'prezto' default

Да?

PR #1715, opened 2019-06-17
update powerlevel10k submodule to the latest commit

romkatv:
The pressing reason for the update is that Pure Power doesn't work with the Powerlevel10k from prezto.


belak:
Looks good to me, thanks!

romkatv:
Thanks for the super speedy merge!

PR #1718, opened 2019-07-04
Ignore tmux autostart in vscode terminal

robsonpeixoto:
The vscode terminal are not setting $VSCODE_PID anymore, maybe because of a bug: https://github.com/microsoft/vscode/issues/63558

But there are a PR adding the variable `TERM_PROGRAM` on terminal: https://github.com/microsoft/vscode/pull/30346.


belak:
This seems reasonable. Thanks for looking in to this!

PR #1721, opened 2019-07-12
Resolves 1641 - Checks whether the prompt uses editor-info (keymap) a…

### LINKED ISSUE #1641: Entering vi insert mode eats one line above multi-line prompt
### Description
Using the vi key-bindings, the line above the multi-line prompt (e.g. _pure_) gets eaten up when entering from normal to insert mode.

### Expected behavior
Entering insert mode should not change the lines above the current prompt.

### Actual behavior
Entering the vi insert mode using any of the buttons "i", "I" or "R" removes the last line above the prompt and moves the prompt upwards.

In normal mode:
image
Entered insert mode by pressing "i":
image

However, it does not happen when entering insert mode by pressing "a" or "o".

### Steps to Reproduce

1. In `zpreztorc`: Change from emacs to vi keybindings and select a multi-line prompt (e.g. _pure_)
2. Open a new terminal and enter a few commands so that the terminal fills up.
3. Go to vi normal mode by pressing ESC and change back to insert mode by either pressing "i", "I" or "R".
4. Note that the last line above the multi-line prompt got eaten up.

### Versions

  - Prezto commit: [e6eea17](
https://github.com/sorin-ionescu/prezto/commit/e6eea17b12d18775cb26df7f0eddcf4d9b5cc59d)
  - ZSH version: 5.6.2, also tested with 5.3
  - OS information: Mac OS Mojave, iTerm2

### Possible Solution ?
I found [in these lines](https://github.com/sorin-ionescu/prezto/blob/master/modules/editor/init.zsh#L178) that there is a special treatment for the functions "vi-insert", "vi-insert-bol" and "vi-replace" and that by commenting out the 3 function definitions, I can get the expected behaviour.
However, I unfortunately understand too less of these things to anticipate the consequences of commenting out these lines. I hope some of you can shed some light on it.

johnpneumann:
Checks whether the prompt uses editor-info (keymap) and an update to the prompt readme to ensure requirements for adding a prompt is clear

Fixes #1641 

## Proposed Changes

  - Check whether or not a prompt has the primary keymap format set before using the editor-info
  - Updates the readme to call out that you need to set a primary keymap to allow for us to actually manage your prompt

## Additional Information

I timed whether using `zstyle -t` with a custom flag over `zstyle -m` with a generic pattern was faster and found that while `zstyle -t` was faster by around double, we were dealing nanoseconds, so it really didn't matter. 

This change allows us to enable prompt developers to utilize our way of doing things, without cluttering up everything else with additional options that need to be set (like I did with the ps-context - it was necessary, but I still hate it). When you're creating a new prompt, you're generally going to create a primary keymap format, so utilizing what we already have makes the most sense.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -92,27 +92,29 @@ function bindkey-all {
 # array.
 function editor-info {
   # Clean up previous $editor_info.
-  unset editor_info
-  typeset -gA editor_info
+  if zstyle -m ':prezto:module:editor:info:keymap:primary' format '*'; then
+    unset editor_info
+    typeset -gA editor_info
 
-  if [[ "$KEYMAP" == 'vicmd' ]]; then
-    zstyle -s ':prezto:module:editor:info:keymap:alternate' format 'REPLY'
-    editor_info[keymap]="$REPLY"
-  else
-    zstyle -s ':prezto:module:editor:info:keymap:primary' format 'REPLY'
-    editor_info[keymap]="$REPLY"
-
-    if [[ "$ZLE_STATE" == *overwrite* ]]; then
-      zstyle -s ':prezto:module:editor:info:keymap:primary:overwrite' format 'REPLY'
-      editor_info[overwrite]="$REPLY"
+    if [[ "$KEYMAP" == 'vicmd' ]]; then
+      zstyle -s ':prezto:module:editor:info:keymap:alternate' format 'REPLY'
+      editor_info[keymap]="$REPLY"
     else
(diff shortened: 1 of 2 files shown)


johnpneumann:
@sorin-ionescu - I tagged you specifically, because I felt like this change is something that is central to prezto and would appreciate your blessing on the change. I know prezto isn't a priority for you at this point, so if we don't hear back, we'll handle it as needed.

belak:
Does this require a change in any prompts as well?

johnpneumann:
@belak - I'm pulling this for now. It's only partially there.

PR #1727, opened 2019-07-24
update powerlevel10k submodule to the latest commit

romkatv:
The last update has picked up a nasty bug in gitstatus that disables all user aliases. This commit fixes it.


belak:
Thanks for keeping this up to date!

PR #1731, opened 2019-08-02
prompt: update powerlevel10k submodule to the latest commit

romkatv:
A dozen bug fixes. Most important:

- `rbenv` and a few other segments didn't work if `IFS` was set to something unusual.
- `vcs` segment couldn't properly apply subsegment style if color overrides used mnemonic names.
- the check for `.p10k.zsh` already being sourced was too strict.


belak:
LGTM, thanks for keeping this up to date!

romkatv:
Thanks for merging my PRs so quickly 🤗

PR #1732, opened 2019-08-07
Nixos gpg module fix

### LINKED ISSUE #1681: GPG module output agent already running on NixOS

Vonfry:
Fixes #1681

Proposed Changes
- gpg module: if-condition ( grep expression and find socket file) in modules/gpg/init.zsh between line 16 to 25.
- remove re-export SSH_AUTH_SOCK in ssh module. Becasue some tools such as pam ssh auth will check socket file permission. A link to this socket is useful but re-export the variable will break some tools.

This pr is the same as #1682 but switching into a new branch(not master).

### DIFF
--- a/modules/gpg/init.zsh
+++ b/modules/gpg/init.zsh
@@ -11,9 +11,9 @@ if (( ! $+commands[gpg-agent] )); then
 fi
 
 # Set the default paths to gpg-agent files.
-_gpg_agent_conf="${GNUPGHOME:-$HOME/.gnupg}/gpg-agent.conf"
+_gpg_agent_conf="$(gpgconf --list-dir homedir)/gpg-agent.conf"
 _gpg_agent_env="${TMPDIR:-/tmp}/gpg-agent.env.$UID"
-_gpg_agent_socket="$(gpgconf --list-dir | grep "agent-socket:.*" | sed "s/agent-socket://")"
+_gpg_agent_socket="$(gpgconf --list-dir agent-socket)"
 
 # Load environment variables from previous run
 source "$_gpg_agent_env" 2> /dev/null
--- a/modules/gpg/init.zsh
+++ b/modules/gpg/init.zsh
@@ -20,7 +20,7 @@ source "$_gpg_agent_env" 2> /dev/null
 
 
 # Start gpg-agent if not started.
-if [[ -z "$GPG_AGENT_INFO" && ! -S ${_gpg_agent_socket}" ]]; then
+if [[ -z "$GPG_AGENT_INFO" && ! -S "${_gpg_agent_socket}" ]]; then
   # Start gpg-agent if not started.
   if ! ps -U "$LOGNAME" -o pid,ucomm | grep -q -- "^[0-9]* gpg-agent$"; then
     eval "$(gpg-agent --daemon | tee "$_gpg_agent_env")"
@@ -49,7 +49,7 @@ if grep '^enable-ssh-support' "$_gpg_agent_conf" &> /dev/null; then
 fi
 
 # Clean up.
-unset _gpg_agent_{conf,env}
+unset _gpg_agent_{conf,env,socket}
 
 # Disable GUI prompts inside SSH.
 if [[ -n "$SSH_CONNECTION" ]]; then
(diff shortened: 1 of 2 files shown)


belak:
Can you give me an example of tools that are broken? Isn't `SSH_AUTH_SOCK` needed for things like `ssh-add`?

Vonfry:
@belak On my system, prezto will print `gpg-agent: a gpg-agent is already running - not starting a new one.`  without this pr, which I don't see on other system.

`SSH_AUTH_SOCK` is used by many tools, but it is export by ssh automatically. In the old prezto ssh module, this variable is exported with a fixed path like `${TMPDIR:-/tmp}/ssh-agent.sock.$UID` which doesn't exist on some system, such as NixOS (this system uses /var/run/$UID/xxx). I removed the repeated assignment in prezto ssh module.

andrei-pavel:
image

Potential alternative fix

Vonfry:
@andrei-pavel In this pr, I use gpgconf to get the correct path instead of manual coding path,so I think we should not add a new condition on the pr

mvcouwen:
Any progress on this PR? I seem to have the same issue when ssh-ing into a ubuntu server.

1wilkens:
I described the problem and potential solutions in #1164

Vonfry:
@1wilkens gpgconf always returns the path whether gpg-agent starts or not. When not starting, it returns the path which agent creates by default.

1wilkens:
But in that case it would return `${GPG_HOME}/.gnupg/S.gpg-agent` wouldn't it? The docs seem to suggest that this is the default and systemd overrides that.
So in our case that would be wrong. Not pointing fingers here, just not sure, what the "correct" solution is..

Vonfry:
systemd use `sockets.target` to install sockets, I'm not sure what will happended if we open a zsh with prezto and then start gpg systemd service. Due to some reason, I haven't a linux machine to test this now. I will try to find how to solve this after September.

Vonfry:
@1wilkens IMO, the one who uses prezto to startup a gpg-agent, may not want to use systemd to start it again. If systemd has started gpg-agent, gpg module in prezto will check it, and won't start again. 

In the first scenario, gpgconf returns the default path which is used by prezto to create sockets.
In the secound scenario, gpgconf can find the socket created by systemd instead of a default value.

Vonfry:
@belak Could this pr be merged? I edit the description and make it compatible on some system.

Proposed Changes:

- gpg module: Use `gpgconf` to get gpg-agent path instead of using environment variables and a fixed path by default, which is more compatible. For example: gpg has been started by systemd, and the sockets is set by sockets.target which is different from the default path ~/.gnugp/S.gpg-agent and no environment variable is set for current shell, but gpgconf can find the correct socket file and using it to check whether gpg-agent is already run. Otherwise, prezto will prompt `gpg-agent: a gpg-agent is already running - not starting a new one` at startup and it don't pick up the correct gpg-agent socket.
- ssh module: remove re-export SSH_AUTH_SOCK in ssh module. Becasue some tools such as pam ssh auth will check socket file's permission. A link to this socket is useful but re-export the variable will break some tools.

tigerkzr:
I found the fix on another users fork but I did not see a pull request for it and they have one syntax error.  What's the protocol for this? I'm new to this kind of thing, but I would like to get the fix into the system.

tigerkzr (changes requested):
Missing double quote on line 23. should be "${-gpg_agent_socket}"

diraol:
@indrajitr or @belak can you folks take a look at this PR to see if it make sense, please? It would be nice to finally _not_ have this message in every terminal 😃

PR #1734, opened 2019-08-18
Fix async using 100% CPU

### LINKED ISSUE #1701: sorin prompt + git = 100% cpu _zsh sub-process
### Description

with the sorin prompt enabled, there are certain git repos I navigate to which cause a -zsh process to start and peg a core at 100%. It eventually stops after a minute or so. RPROMPT never populates w/ git-info when this happens.

### Expected behavior

Browse git repo without CPU pegging, and git-info displaying properly

### Actual behavior

-zsh process starts, pegs CPU at 100%, eventually stops. RPROMPT never updates w/ current branch & status.

### Steps to Reproduce

1. Fresh install
2. Enable 'git' module
3. Browse to git repo

This does not happen for me in every git repo... I'm not sure how to debug this.

### Versions

  - Prezto commit: 1f4601e44c989b90dc7314b151891fa60a101251
  - ZSH version: zsh 5.1.1 (x86_64-ubuntu-linux-gnu)
  - OS information: Ubuntu 16.04
  - Git version: git version 2.19.1

pedrosland:
Fixes #1701

After getting tired of my git state not updating most of the time and realising that it was also eating my CPU, I pulled out GDB. It indicated that async was in the loop consuming CPU which makes sense. After checking the async module, I found a wonderful commit message in the latest version (mafredri/zsh-async@43de5e0):

```
Prevent infinite loop in case of broken zpty file descriptor (#31)

In some cases, the fd created by zpty can become corrupted: if the other
end was disconnected, if the fd was closed and so on.
If we ignore this condition, then we enter a infinite loop.

Instead, when the fd becomes invalid, unregister the corresponding
worker and notify the callback.
```

Since I updated the submodule earlier today, I haven't had any troubles with the git status.

## Proposed Changes

  - Update async submodule to v1.7.2 to fix issue where the the git status in Sorin prompt doesn't update and uses 100% CPU on one core.


kylemanna:
I've been experiencing this as well on both of my daily use machines.  I'm going to give this a try, thanks!

belak:
Awesome! Thanks for looking into this! I'm so glad someone who was seeing this issue looked into it!

PR #1739, opened 2019-09-10
history: Permit use of previously set HISTFILE

shaneog:
## Proposed Changes

  - Honor the use of the HISTFILE environment variable, if it's already set.


belak:
This seems reasonable. Thanks for submitting this!

PR #1740, opened 2019-09-10
fix directory module docs: it doesn't really set AUTO_NAME_DIRS

romkatv:
Module `directory` used to set `AUTO_NAME_DIRS`. This was changed in https://github.com/romkatv/prezto/commit/73e94b84bbe9bebfe612438507a53885485938cb but `README.md` hasn't been updated. This looks like an accidental omission that this PR fixes.


belak:
Thanks for submitting this! Stuff like this can be easy to miss.

PR #1741, opened 2019-09-12
add bullet train

### LINKED ISSUE #1282: Add additional external prompt themes
In particular, https://github.com/bhilburn/powerlevel9k and https://github.com/caiogondim/bullet-train.zsh both have over 1000 stars and a reasonably large number of stars/followers.

powerlevel9k claims to have prezto support, but it seems like most people use other frameworks so if prezto isn't installed to ~/.prezto it's broken. Including it in prezto should make it easier for people to get started with it and improve support on both fronts over time.

jan-xyz:
I'm using this theme already with a submodule in my dotfiles and saw that others also would like to have it so I thought it made sense to add it here as well :)

Fixes #1282

## Proposed Changes

  - add bullet_train theme

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "modules/prompt/external/powerlevel10k"]
 	path = modules/prompt/external/powerlevel10k
 	url = https://github.com/romkatv/powerlevel10k.git
+[submodule "modules/prompt/external/bullet-train"]
+	path = modules/prompt/external/bullet-train
+	url = https://github.com/caiogondim/bullet-train.zsh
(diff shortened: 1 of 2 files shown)


belak:
Hi! Thanks for submitting this! Unfortunately, we aren't accepting new theme submissions in the main repo. There is a contrib repo (https://github.com/belak/prezto-contrib) which is fairly easy to plug into.

In terms of git handling, you may want to take a look at the [theming section](https://github.com/sorin-ionescu/prezto/tree/master/modules/git#theming) of the git module as well as the [simple](https://github.com/sorin-ionescu/prezto/blob/master/modules/prompt/functions/prompt_minimal_setup) and [belak](https://github.com/belak/prezto-contrib/blob/master/contrib-prompt/functions/prompt_belak_setup) prompts as examples.

Because we're not currently accepting new themes, please feel free to re-submit this to the contrib repo, but I'm going to close this for now.

Issue #1713, opened 2019-06-06
zprezto-update does not init a new submodule


lodagro:
### Description
Running zprezto-update will not init a new submodule.

### Expected behavior
Also init new submodule(s)

### Actual behavior
ac65e8e added the powerlevel10k sumodule. However the ```zprezto-update``` function did not init this submodule on my existing repo. On shell startup this results in the message: ```Couldn't read file ~/.zprezto/modules/prompt/functions/prompt_powerlevel10k_setup containing theme powerlevel10k.```

A manual ```git submodule update --init ~/.zprezto/modules/prompt/external/powerlevel10k``` fixed things.

### Versions
  - Prezto commit: 1d58546
  - ZSH version: 5.4.1
  - OS information: RHEL6.6

lodagro:
Fixed by 1bc0da5

PR #1748, opened 2019-10-15
Init submodules when using zprezto-update (#1713)

lodagro:


belak:
Yeah, this makes sense. Especially since we had an issue with this when we added powerlevel10k. Thanks!

PR #1750, opened 2019-10-15
Allow mapping Ctrl+S and Ctrl+Q shortcuts

maximbaz:
Hi @belak, please treat this as a question and a proposal 🙂 

While looking at [this PR](https://github.com/jhawthorn/fzy/pull/124) I learned that `Ctrl+S` and `Ctrl+Q` are by default now allowed to be used by other apps.

Why don't we allow this by default? Is there a drawback in doing so?

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -38,7 +38,8 @@ setopt INTERACTIVE_COMMENTS # Enable comments in interactive shell.
 setopt RC_QUOTES            # Allow 'Henry''s Garage' instead of 'Henry'\''s Garage'.
 unsetopt MAIL_WARNING       # Don't print a warning message if a mail file has been accessed.
 
-stty -ixon                  # Allow mapping Ctrl+S and Ctrl+Q shortcuts
+# Allow mapping Ctrl+S and Ctrl+Q shortcuts
+(( $+commands[stty] )) && stty -ixon
 
 #
 # Jobs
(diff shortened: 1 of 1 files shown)


belak:
I've personally always had this in my zshrc - https://github.com/belak/dotfiles/blob/90535a7f889d7fdc074c0a84673c26c87b877451/zshrc#L76

This is a good question because at least in my mine ^s and ^q are unexpected behavior by default...

The only possible drawback is that I'm not sure how this would interact on Windows, but we can probably make a fix if/when that issue pops up.

I'd be happy to merge this (or some derivative of this) if other maintainers also think it's a good idea.

CC @jeffwidman @johnpneumann

belak:
Maybe wrap this in `if (( $+commands[stty] ))` so we can be sure this won't break systems where stty doesn't exist, then I think this would be good to go

maximbaz:
Sorry for late reply, agree, done 👍

johnpneumann:
I'm ok with this. Anyone who wants the default functionality would likely know how to flip it back.

belak:
Looks good. Thanks!

jeffwidman:
Sorry, I was mostly offline on paternity leave when you pinged me @belak, just catching up now. 

No objections from me regarding this change.

belak:
Oh wow, congratulations @jeffwidman!

No worries, we got it taken care of.

PR #1751, opened 2019-10-17
Update homebrew module documentation and add alias `brewL`

lambdanerd:
## Proposed Changes

 - Remove the documentation for the obsolete `brewC` alias
 - Add alias `brewL` to list leaves

### DIFF
--- a/modules/homebrew/README.md
+++ b/modules/homebrew/README.md
@@ -18,7 +18,6 @@ Aliases
 ### Homebrew
 
   - `brewc` cleans outdated brews and their cached archives.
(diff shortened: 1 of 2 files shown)


belak:
Thanks for doing this! I meant to update the docs after the PR that removed brewC but I forgot!

I'm almost tempted to reverse the `brew list` and `brew leaves` aliases since I know I care about leaves more often than not... but it would be a breaking change and not everyone thinks like that.

lambdanerd:
I too use leaves almost exclusively, but as the saying goes, "don't fix what ain't broken".

belak:
Sounds good. Thanks for submitting this!

PR #1752, opened 2019-10-19
fix(alias): ls alias lx is an illegal option on darwin bsd style ls

casaper:
### Fixes illegal option `-X` on MacOS

The option `-X` gets MacOS' `ls` (and probably also other BSD-ish OS) to complain about an illegal option:

```sh
user@host ~ $ ls -lhXB
ls: illegal option -- X
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

### Proposed Changes

  - only define the alias `lx` with option `-X` on machines that are not Darwin

I would also exclude other BSD like OS but since I do not have such a host at hand I cannot verify if that is even really needed for them.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -118,12 +118,16 @@ alias ll='ls -lh'        # Lists human readable sizes.
 alias lr='ll -R'         # Lists human readable sizes, recursively.
 alias la='ll -A'         # Lists human readable sizes, hidden files.
 alias lm='la | "$PAGER"' # Lists human readable sizes, hidden files through pager.
-alias lx='ll -XB'        # Lists sorted by extension (GNU only).
 alias lk='ll -Sr'        # Lists sorted by size, largest last.
 alias lt='ll -tr'        # Lists sorted by date, most recent last.
 alias lc='lt -c'         # Lists sorted by date, most recent last, shows change time.
 alias lu='lt -u'         # Lists sorted by date, most recent last, shows access time.
 alias sl='ls'            # I often screw this up.
+
+# Only make alias if ls supports -XB (darwins BSD style ls doesn't)
+if ls -XB &> /dev/null; then
+  alias lx='ll -XB'        # Lists sorted by extension (GNU only).
+fi
 
 # Grep
 if zstyle -t ':prezto:module:utility:grep' color; then


wadkar:
👍 
Makes sense. Also the recent PR #1758 will make it work with people who've done a `brew install coreutils`.

casaper:
@wadkar 

> Makes sense. Also the recent PR #1786 will make it work with people who've done a `brew install coreutils`.

Who could I request a review from then? I mean in order to get it merged?

modules/utility/init.zsh
```diff
@@ -118,7 +118,9 @@ alias ll='ls -lh'        # Lists human readable sizes.
 alias lr='ll -R'         # Lists human readable sizes, recursively.
 alias la='ll -A'         # Lists human readable sizes, hidden files.
 alias lm='la | "$PAGER"' # Lists human readable sizes, hidden files through pager.
-alias lx='ll -XB'        # Lists sorted by extension (GNU only).
+if [[ "$OSTYPE" != darwin* ]]; then
```
srijanshetty: Can we move this to the top? Makes the code more maintainable in the long run. (this can be considered as a nitpick :) )

srijanshetty:
Added a comment

modules/utility/init.zsh
casaper: can do. But I just considered to keep changes at a minimum. 

But regarding readability I also had something in mind, and that is why I did #1793 with helpers. Presuming that helper would exist, would that calm your nit'ish feeling a little? :wink:

srijanshetty: I don't have a view on #1793, so I was waiting for the other contributors decide. Probably should add @belak to the reviewers list :smile: for #1793

casaper: #1793 is highly optional. I was just mentioning it, because I interpreted your comment as a readability one.

Will add belak then... thx for the hint 😄

srijanshetty: My comment is totality a readability one :smile:

belak:
I'm not sure the best way to do this, but this alias should be enabled only if ls is coreutils based. I believe this will also be broken on BSD.

It's good to know though. I'd be happy to merge this with a slightly better check.

casaper:
@belak 

> It's good to know though. I'd be happy to merge this with a slightly better check.

What would you think if I checked for ls running with code 0 with `-XB` instead, and only then add the alias?

This way it would really only depend on ls's capabilities, and not on the OS at all.

```sh
if ls -XB &> /dev/null; then
  alias  lx='ll -XB'        # Lists sorted by extension (GNU only).
fi
```

Haven't tried if that would work well everywhere. The idea just came up when you asked for a better check...

modules/utility/init.zsh
casaper: @srijanshetty 

I changed this code now. Now I check for the capability of ls instead of the OS running. 

Just the fact that you're running on darwin doesn't prove your having a BSD style `ls`.
Since I'm aware of its existence I myself have the gnu coreutils installed here, so for myself this fix is not needed any more. :wink:

modules/utility/init.zsh
```diff
@@ -118,13 +118,17 @@ alias ll='ls -lh'        # Lists human readable sizes.
...
 alias lm='la | "$PAGER"' # Lists human readable sizes, hidden files through pager.
-alias lx='ll -XB'        # Lists sorted by extension (GNU only).
 alias lk='ll -Sr'        # Lists sorted by size, largest last.
 alias lt='ll -tr'        # Lists sorted by date, most recent last.
 alias lc='lt -c'         # Lists sorted by date, most recent last, shows change time.
 alias lu='lt -u'         # Lists sorted by date, most recent last, shows access time.
 alias sl='ls'            # I often screw this up.
 
+# Only make alias if ls supports -XB (darwins BSD style ls doesn't)
+if ls -XB &> /dev/null; then
```
indrajitr, line 128: Wondering if you would consider `if is-callable 'dircolors'` instead.
`ls -XB` might be expensive depending on the directory on which it is executed.

These things keep on adding to the startup time (unless we memoize it on first call).

belak: Maybe we could use `ls /dev/null` since that should only be a single file and should be fast.

indrajitr: Yes, `ls -XB /dev/null` would be a great idea! That said, would be nice to be consistently using *one* way to detect BSD/GNU versions. Turns out we use `is-callable 'dircolors'` to detect this already.

indrajitr: Or just detect the options via `--help` like so:
```shell
if grep -q '\-X' <(ls --help 2>&1); then
  # ...
fi
```
We do similar thing in `rsync` module.

casaper: @indrajitr I wasn't aware of the performance penalty my idea could introduce, and your concern makes quite a lot of sense.

On the part of finding a better solution:  

At first glance the most straight forward solution @belak has suggested `ls -XB /dev/null` seems feasible, but in the end the `callable 'dircolors'` appears to be the most efficient and elegant alternative. 

But regarding the test with dircolors, I have one question: Is there an unlikeley case I might not see, where dircolors is not callable and gnu-ls is present or the other way around?

PR #1753, opened 2019-10-25
Container devel

hlecuanda:
## Proposed Changes

  - Provides a tool for rapid testing and development without disrupting your own einvironment
  - Most CI tools are based don docker containers so this could help get on the CI and QA bandwagon
  - I'm not exactly sure where to place this, and i'm sure  is a lot of room for improvement. the Makefile and Dockerfile are heavily commented, plus i added a [container-README.md ](https://github.com/hlecuanda/prezto/blob/container-devel/container-README.md) that walks through the features already on the container if you wish to try it out

here's a quick screencast too https://asciinema.org/a/277054

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ RUN apk --no-cache update && apk upgrade && \
     apk add util-linux pciutils usbutils coreutils binutils\
     findutils grep man man-pages mdocml-apropos less less-doc \
     make grep zsh zsh-vcs zsh-zftp zsh-calendar zsh-doc git \
-    vim git-zsh-completion tmux tmux-zsh-completion \
+    vim git-zsh-completion tmux tmux-zsh-completion tree \
     docker-zsh-completion
 RUN addgroup prezto && adduser -D prezto -G prezto -S /bin/zsh
 WORKDIR /home/prezto
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,11 @@
 FROM alpine as prezto-devel
 LABEL maintainer="h@hlo.mx"
 RUN apk --no-cache update && apk upgrade && \
-    apk add make grep zsh git
+    apk add util-linux pciutils usbutils coreutils binutils\
+    findutils grep man man-pages mdocml-apropos less less-doc \
+    make grep zsh zsh-vcs zsh-zftp zsh-calendar zsh-doc git \
+    vim git-zsh-completion tmux tmux-zsh-completion \
+    docker-zsh-completion
 RUN addgroup prezto && adduser -D prezto -G prezto -S /bin/zsh
 WORKDIR /home/prezto
 USER prezto
@@ -11,7 +15,8 @@ RUN cp src/Makefile .
 RUN make .clone
 RUN make src .homercs
 # ENTRYPOINT ["/bin/zsh", "-"]
-CMD ["/bin/zsh","-l","-o","verbose","-o","xtrace","-o","sourcetrace"]
+# CMD ["/bin/zsh","-l","-o","verbose","-o","xtrace","-o","sourcetrace"]
+CMD ["/bin/zsh","-l"]
 
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM alpine as prezto-devel
+LABEL maintainer="h@hlo.mx"
+RUN apk --no-cache update && apk upgrade && \
+    apk add make grep zsh git
+RUN addgroup prezto && adduser -D prezto -G prezto -S /bin/zsh
+WORKDIR /home/prezto
+USER prezto
+RUN cd /home/prezto && mkdir src
+COPY . src
+RUN cp src/Makefile .
+RUN make .clone
+RUN make src .homercs
+# ENTRYPOINT ["/bin/zsh", "-"]
+CMD ["/bin/zsh","-l","-o","verbose","-o","xtrace","-o","sourcetrace"]
+
+
+
+
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 *.zwc.old
 modules/*/cache.zsh
 contrib
+.container
--- a/Makefile
+++ b/Makefile
@@ -96,8 +96,9 @@ run: .container/image
 # Removes the current container and image, so we can build a new one
 # from scratch. If you want a real clean slate, then "make realclean"
 clean:
-	docker container rm $(TAG):
-	docker image rm $(TAG):
+	docker container rm $(TAG)
+	docker image rm $(TAG)
+	rm -fv .container/*
 
 # Deep cleaning, will remove dangling (intermediate build) images
 # and containers for which a final image or container cannot be found
@@ -105,6 +106,7 @@ clean:
 # containers. Run every once in a while.
 realclean:
 	$(MAKE) clean
+	docker rmi $(TAG)
 	docker container prune
 	docker image prune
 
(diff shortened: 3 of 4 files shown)


belak:
Sorry, I've been super busy - I really like the idea behind this, especially making some issues easier to reproduce. Thanks for taking the time to do this!

hlecuanda:
My pleasure! and I'm sorry too, year end is coming up, and it's hectic around here this time of the year. 
I'm sure there is a lot of room for improvement for this container, and i'd be happy to work on any feedback to make it more useful for the prezto project. 

Do you think this is the right place for this? or is this better suited for a separate branch, or an external contrubution? maybe a plugin? please let me know what ypu think

Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM alpine as prezto-devel
+LABEL maintainer="h@hlo.mx"
+RUN apk --no-cache update && apk upgrade && \
```
casaper: Instead of --no-cache the common recommendation is this at the end

```
  && rm -rf /var/cache/apk/*
```

And Why do you do `apk upgrade` ? Since you imply `alpine:latest` in `FROM`.
Because If you didn't do that, it would probably shrink the image size significantly, without you having any flaws I can see.

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    apk add util-linux pciutils usbutils coreutils binutils\
+    findutils grep man man-pages mdocml-apropos less less-doc \
+    make grep zsh zsh-vcs zsh-zftp zsh-calendar zsh-doc git \
+    vim git-zsh-completion tmux tmux-zsh-completion tree \
+    docker-zsh-completion
+RUN addgroup prezto && adduser -D prezto -G prezto -S /bin/zsh
+WORKDIR /home/prezto
+USER prezto
+RUN cd /home/prezto && mkdir src
+COPY . src
```
casaper: If you used `ADD` instead of `COPY` it would speed up things, because then the files would not really be copied into the container.  
Or do you need it like that so that your local sources won't be changed with the later make runs?

wadkar: Can you please tell me how `ADD` will speed things up? According to docker's [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy) individual `COPY` statements are preferred.

casaper: ADD adds the directory as a docker volume.  
COPY really creates copies of the files into the docker container.

At least that was how I understood it so far...

Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+RUN cp src/Makefile .
+RUN make .clone
+RUN make src .homercs
```
casaper: What I don't understand is why you run make inside the docker file.

Without fully understanding your makefile and your shell script, I just try to tell you how I would approach that what I assume you want to achieve. If that assumptions are wrong or if I'm anyhow wrong, no big deal. I only try to help.

My Dockerfile would be like this:

```Dockerfile
FROM alpine

RUN apk --no-cache add \
  util-linux \
  pciutils \
  usbutils \
  coreutils \
  binutils \
  findutils \
  grep \
  man \
  man-pages \
  mdocml-apropos \
  less \
  less-doc \
  make \
  grep \
  zsh \
  zsh-vcs \
  zsh-zftp \
  zsh-calendar \
  zsh-doc \
  git \
  vim \
  git-zsh-completion \
  tmux \
  tmux-zsh-completion \
  tree \
  docker-zsh-completion \
  && addgroup prezto \
  && adduser -D prezto -G prezto -S /bin/zsh

WORKDIR /home/prezto
USER prezto

CMD ["/bin/zsh"]
```

And then I would add a `docker-compose.yml` file and run the stuff through that:

```yml
version: "3.7"

services:
  prezto:
    build: .
    volumes:
      - ./:/home/prezto/.zprezto
      - ./my-test-stuff:/home/prezto/test
      - ./runcoms/zshenv:/home/prezto/.zshenv
      - ./runcoms/zprofile:/home/prezto/.zprofile
      - ./runcoms/zshrc:/home/prezto/.zshrc
      - ./runcoms/zpreztorc:/home/prezto/.zpreztorc
      - ./runcoms/zlogin:/home/prezto/.zlogin
      - ./runcoms/zlogout:/home/prezto/.zlogout
```

Now you can set that up with:

```sh
docker-compose build  # builds your docker file
docker-compose run --rm prezto /bin/zsh
Couldn't read file /home/prezto/.zprezto/modules/prompt/functions/prompt_agnoster_setup containing theme agnoster.
Couldn't read file /home/prezto/.zprezto/modules/prompt/functions/prompt_powerlevel10k_setup containing theme powerlevel10k.
Couldn't read file /home/prezto/.zprezto/modules/prompt/functions/prompt_powerlevel9k_setup containing theme powerlevel9k.
Couldn't read file /home/prezto/.zprezto/modules/prompt/functions/prompt_powerline_setup containing theme powerline.
Couldn't read file /home/prezto/.zprezto/modules/prompt/functions/prompt_pure_setup containing theme pure.
prompt_sorin_setup:7: async: function definition file not found
prompt_sorin_async_tasks:4: command not found: async_start_worker
prompt_sorin_async_tasks:5: command not found: async_register_callback
prompt_sorin_async_tasks:10: command not found: async_flush_jobs
prompt_sorin_async_tasks:13: command not found: async_job
~ ❯❯❯
```

*I did exactly this here, and it is partial C&P from my shell. So I'm sure that it works.*

This has several benefits:

1. your image will not have to check out the prezto repository, because that you allready have in your dev dir
1. `/home/prezto` has all the dotfiles like `.zshenv` allready in there, directly from your dev machines runcoms
1. You can use the same image, it's smaller, and you only need to create new containers

You can have some test scripts inside the prezto project like this:

```sh
 docker-compose run --rm prezto /home/prezto/test/some-script.sh
I'm echoed by  some-script.sh doing the stuff you want to do...
```

This way you can then run technically anything in that container and get that container flushed right after your done.  
And you're working with your current WIP directories code.

I just recon it is not really the way to run your tests within the `Dockerfile`'s `RUN` statements.

Dockerfile
wadkar: I don't think it works like that. Neither statements add a directory as a volume. Docker volumes are different directives manipulated by VOLUME statement in Dockerfile or `-v` flag at runtime.

Both `ADD` and `COPY` statements copy the source from the build context (first argument) to the destination in docker image (second argument).

Specifically, as the best practices link explains, the ADD statement has additional magic to it - it will _automagically_ uncompress a compressed file, or if the first argument is a URL it will download it.

`COPY . src` is good first pass for a Dockerfile. Though ideally, one would prefer to have separate statements for each directory, so that a change in one file in one directory will not invalidate the entire cache. But then we run into the problem of how we will order individual COPY statements. So, I am okay with `COPY . src` for now.

wadkar:
Looks good after reading through the diff @hlecuanda , let me follow the README and `make` the container over the weekend. Does that sound good to you?

Dockerfile
casaper: Well however, it doesn't really matter to me actually if ADD or COPY is used in the end here. 

My main point is not running stuff in the image building process, but rather run it on the finished image... 
Its just quicker, and sllicker. If you prefer to build the image newly each time. you run something, or change code and then test it, and fail and change it again... up to you. :wink:

wadkar:
So I played with thee Dockerfile over the weekend and it was fun. I think it'll help us develop/test prezto without messing with our day-to-day config changes.

I am more inclined towards following @casaper 's suggestion to mount the locally cloned prezto repository instead of using Makefiles to install it. I was on a limited bandwidth over the weekend, and my developer experience was severely hampered every time I needed to build the image with some modifications to the Dockerfile of my own.

Dockerfile
wadkar: Yeap, this makes a lot more sense than building prezto inside the image.

PR #1756, opened 2019-11-03
Add module nvm-auto-use

lambdanerd:
`nvm` Auto Use
==============

This module integrates [`nvm` auto use](https://github.com/nvm-sh/nvm#zsh) shell integration into Prezto. It will automatically call `nvm use` in a directory with a .nvmrc file. This recipe is **not** supported by the `nvm` development team, but they are accepting pull requests for more examples.

Screen Shot 2019-11-03 at 4 04 37 PM
Screen Shot 2019-11-03 at 4 05 30 PM

### DIFF
--- a/modules/nvm-auto-use/README.md
+++ b/modules/nvm-auto-use/README.md
@@ -1,0 +1,16 @@
+`nvm` Auto Use
+==============
+
+This module integrates [`nvm` auto use][1] shell integration into Prezto. It will automatically call `nvm use` in a directory with a .nvmrc file. This recipe is **not** supported by the `nvm` development team, but they are accepting pull requests for more examples.
+
+This module must be loaded **after** the [node](../node) module.
+
+Author
+------
+
+*The author of this module should be contacted via the [issue tracker][2].*
+
+- [Jared Manning](https://github.com/lambdanerd)
+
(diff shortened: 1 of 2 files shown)


rjcoelho:
What's wrong with using direnv instead ?

lambdanerd:
@rjcoelho I’m not familiar with direnv. This is specifically for nvm.

rjcoelho:
https://direnv.net/ is a tool to source .envrc file (a shell script) when you change into a directory. Is better then chpwd because it undoes any changes when you cd out of the directory.

lambdanerd:
@rjcoelho Okay, but that's a different tool altogether. Besides, the last `elif` takes care of that anyways. Plus, this isn't my own custom code, it's provided by `nvm`. I don't think it's constructive to come into a pull request and suggest that it's invalid because an existing tool provides a similar function.

belak:
Thanks for submitting this! Unfortunately, we're trying to keep core fairly lean and avoiding new modules. If you want to re-submit this to the [contrib repo](https://github.com/belak/prezto-contrib) I'd be happy to accept it there.

lambdanerd:
Will do, thanks @belak!

PR #1757, opened 2019-11-05
prompt: update powerlevel10k submodule to the latest commit

romkatv:
- New feature: Instant Prompt.
  - Zsh starts instantly even if zshrc loads dozens of plugins.
  - Must be explicitly enabled via `p10k configure`.
- New prompt segment: azure.
- New styles: Pure and Rainbow.
- 6 times faster loading (independent of instant prompt).


belak:
Sorry about the delay. Thanks!

PR #1771, opened 2019-12-20
apply `stty -ixon` only to tty

romkatv:
Currently, prezto may call `stty -ixon` when stdin and/or stdout are not connected to a tty. This results in an error:

```text
stty: 'standard input': Inappropriate ioctl for device
```

This PR fixes it by explicitly redirecting `stty` to `$TTY`. `TTY` is a special parameter set by zsh. It points to the tty (if there is one) even if stdin and/or stdout are redirected.

I should also note that the comment above `stty -ixon` is a bit misleading:

```zsh
# Allow mapping Ctrl+S and Ctrl+Q shortcuts
```

`stty -ixon` disables Ctrl+S and Ctrl+Q terminal shortcuts which stop and resume terminal output respectively. It does allow you to bind these keys to zle widgets as a side effect. This is not a requirement, though. These same keys can be bound if `flow_control` option is unset. This option disables the default Ctrl+S and Ctrl+Q behavior only within zle without affecting everything else that you might run in the terminal.

My preferred setup is to **not* call `stty -ixon` (so that I can pause/resume console output when something is spewing out diagnostic messages very fast) and to unset `flow_control` (because freezing zle is never desired and being able to bind Ctrl+S and Ctrl+Q is nice).

Note that prezto unsets `flow_control` in `modules/completion/init.zsh`.

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -39,7 +39,7 @@ setopt RC_QUOTES            # Allow 'Henry''s Garage' instead of 'Henry'\''s Gar
 unsetopt MAIL_WARNING       # Don't print a warning message if a mail file has been accessed.
 
 # Allow mapping Ctrl+S and Ctrl+Q shortcuts
-(( $+commands[stty] )) && stty -ixon
+[[ -n ${TTY:-} && $+commands[stty] == 1 ]] && stty -ixon <$TTY >$TTY
 
 #
 # Jobs


magnetophon:
This fixes the error for me too, thanks!

belak:
Thanks for submitting this! Good catch.

jcassee:
Thanks, @romkatv

qosmio:
```shell
[01/17/20 05:42:39]
pi@rpi4-1 ~
➤ sudo su - hass
/srv/hass/.zprezto/modules/environment/init.zsh:42: permission denied: /dev/pts/2
```
With this latest change, I'm getting the error above. This seems to only affect when 'sudo su -' into accounts. I reverted the change back to 
```shell
(( $+commands[stty] )) && stty -ixon
``` 
for the time being

romkatv:
Oh man, I should've just deleted that line. It was added by mistake.

Could you try this?

```zsh
[[ -r ${TTY:-} && -w ${TTY:-} && $+commands[stty] == 1 ]] && stty -ixon <$TTY >$TTY
```

qosmio:
Yes! Sorry for the long delay. Confirming that fixes the issue.

romkatv:
Thanks for the confirmation, @qosmio.

PR #1775, opened 2020-01-03
don't raise an error in zlogin when stderr is not a TTY

romkatv:
The existing code in runcoms/zlogin results in `$?` being equal to `1` when starting a login shell if stderr is not a TTY. For example:

```zsh
zsh -l 2>/dev/null
```

When using a theme that displays error/success status of the last command, the first prompt will show an error.

This commit fixes it so that error code is zero after sourcing zlogin (unless something unexpected and bad happens).

### DIFF
--- a/runcoms/zlogin
+++ b/runcoms/zlogin
@@ -15,7 +15,7 @@
 } &!
 
 # Execute code only if STDERR is bound to a TTY.
-[[ -o INTERACTIVE && -t 2 ]] && {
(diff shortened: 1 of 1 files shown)


belak:
LGTM! Thanks for submitting this!

hlecuanda:
OMG that error status on login had been driving me nuts for months!! Thx!

romkatv:
> OMG that error status on login had been driving me nuts for months!! Thx!

Thanks to @jayree for reporting it 🥇

PR #1777, opened 2020-01-08
Allow module functions directory to be a symlink

jcassee:
I am converting the [nix-zsh-completions](https://github.com/spwhitt/nix-zsh-completions) plugin to a module. It comes with completions. It makes sense to link the `functions` directory into `external`, otherwise I would have to create individual symlinks. (There are about 20.)

## Proposed Changes

Replace the expansion that locates the functions directory from  `functions(/FN)` to `functions(-/FN)`, so it matche symlinks to directories.

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -109,7 +109,7 @@ function pmodload {
       pmodule_location=${locations[1]}
 
       # Add functions to $fpath.
-      fpath=(${pmodule_location}/functions(/FN) $fpath)
+      fpath=(${pmodule_location}/functions(-/FN) $fpath)
 
       function {
         local pfunction


jcassee:
Alternatively, many oh-my-zsh plugins have their functions in the root of the plugin directory. Would it be a good idea to support them natively? Not sure how to indicate to prezto that this is the case, though.

belak:
I'd like to avoid also having a module loader in here... otherwise prezto turns into a plugin manager in addition to a framework and there are a plethora of plugin managers out there for zsh already.

belak:
Merged. Thanks for submitting this!

PR #1780, opened 2020-01-09
Support pmodule-allow-overrides feature

### LINKED ISSUE #1779: Feature request - allow overriding built-in modules with the same name
Oh-my-zsh allows for a zsh_custom directory for custom plugins and plugin overrides. Prezto's equivalent, `contrib`, is actually pretty compatible with omz's $ZSH_CUSTOM folder. For example, prezto handles loading omz-style files (_name_.plugin.zsh) instead of init.zsh files. However, if you name a module the same as one that prezto already has, this code block prevents having name collisions:

https://github.com/sorin-ionescu/prezto/blob/7c94154dda2dfcf02515f1caf1e9dfe6d80b0350/init.zsh#L100-L102

I believe it would be good to be able to override modules prezto provides with the same name in your own custom contrib folder. This wouldn't have to be the default behavior - a zstyle config could be added to allow overrides with the same name to favor your contrib version. One added benefit beyond OMZ migrations would be that user's contribs wouldn't break when a new module with a name collision was added to prezto. I'm willing to submit a PR that implements this feature if it would be accepted.

mattmc3:

Fixes #1779

## Proposed Changes
 
 - Adds a new optional zstyle setting (defaulted to 'no' or not-existing): `zstyle ':prezto:load' pmodule-allow-overrides`
  - Adds support in init.zsh for overriding modules with the same name based on whether `zstyle ':prezto:load' pmodule-allow-overrides 'yes'` is set in .zpreztorc

### DIFF
--- a/init.zsh
+++ b/init.zsh
@@ -98,10 +98,7 @@ function pmodload {
     else
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
       if (( ${#locations} > 1 )); then
-        if zstyle -t ':prezto:load' pmodule-allow-overrides 'yes'; then
-          # Use the final location as an override when there are module name conflicts
-          locations=($locations[-1])
-        else
+        if ! zstyle -t ':prezto:load' pmodule-allow-overrides 'yes'; then
           print "$0: conflicting module locations: $locations"
           continue
         fi
@@ -111,7 +108,7 @@ function pmodload {
       fi
 
       # Grab the full path to this module
-      pmodule_location=${locations[1]}
+      pmodule_location=${locations[-1]}
 
       # Add functions to $fpath.
       fpath=(${pmodule_location}/functions(/FN) $fpath)
--- a/init.zsh
+++ b/init.zsh
@@ -98,8 +98,13 @@ function pmodload {
     else
       locations=(${pmodule_dirs:+${^pmodule_dirs}/$pmodule(-/FN)})
       if (( ${#locations} > 1 )); then
-        print "$0: conflicting module locations: $locations"
-        continue
+        if zstyle -t ':prezto:load' pmodule-allow-overrides 'yes'; then
+          # Use the final location as an override when there are module name conflicts
+          locations=($locations[-1])
+        else
+          print "$0: conflicting module locations: $locations"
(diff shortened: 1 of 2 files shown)


belak:
What would you think if it just loaded the first one no matter what and just warned if there were more than one. It would still be useful to have the setting to disable the warning text.

mattmc3:
@belak - I don't think that will work, or maybe I'm not understanding what you're saying. The way prezto works today is that the first one is always loaded already, and the first one is the ones that come with prezto. Say you added your own custom git module to $ZPREZTODIR/contrib/git, it will always warn you that there's a conflict because $ZPREZTODIR/modules/git exists. Basically, once prezto uses a module name, it's locked out for anyone else to use for theirs. This feature fixes that by letting you use the LAST one specified with no warning message. Using the first one is no good, and doesn't capture the override concept that OMZ uses to great effect.

belak:
The current code skips loading if there are conflicting module names.

Your approach in this PR seems good, though I think it can be simplified a little bit - down below, where we set pmodule_locations, that could be set to the last location rather than the first, which means we only need the error case above, rather than the two separate cases.

Thanks for submitting this! It was something I debated about when I wrote this code a long time ago - I can see why it makes more sense this way.

mattmc3:
@belak - Great suggestion. I've made that change. See if this is more to your liking.

Also, just so it exists in the discussion for posterity, there's one other consideration we should sync on. The intent of this functionality is that if you can override a built-in prezto module, but if you want to build off one instead of replacing it, you can easily still source prezto's from your own. For example, to extend the git module, the first line in `$ZPREZTODIR/contrib/git/init.zsh` could be `source $ZPREZTODIR/modules/git/init.zsh`. However, that doesn't account for functions in the original module, and you'll also need to do something like this as well:

```shell
fpath=("$ZPREZTODIR"/contrib/git/functions $fpath)
for _zfunc in "$ZPREZTODIR"/contrib/git/functions/*(.N); do
  autoload -Uz "$_zfunc"
done
unset _zfunc
```

Not sure if you would like this documented anywhere. Happy to do that if that too if you tell me where.

belak:
Thanks for submitting this! Looks good to me!

PR #1781, opened 2020-01-12
Update zsh-autosuggestions to v0.6.4

aldanor:
`zsh-autosuggestions` released a new version, adding a few major features like being able to ignore patterns in history, would be nice to get that bumped here as well.

Been using that locally with prezto for a while now, seems all fine.


belak:
I'd be happy to merge this, but it looks like you're pointing to the latest commit rather than 0.6.4. Once that's fixed, I'll merge it in.

aldanor:
@belak Indeed... fixed now.

belak:
It looks like I was mistaken and the v0.6.4 tag was actually at the tip. I pushed a change manually and included your account as the author. Thanks for submitting this!

PR #1785, opened 2020-01-18
attempt to unbreak the TTY check

romkatv:
Bug report from @qosmio: https://github.com/sorin-ionescu/prezto/pull/1771#issuecomment-575596417.

I'm unable to reproduce it and @qosmio hasn't yet confirmed or disconfirmed whether this fix works. It's a good idea to change the code this way in either case.

(As I mentioned in #1771, this `stty -ixon` call may have been added by mistake. It doesn't do what the comment suggest and is at least as likely to be harming users as helping them.)

### DIFF
--- a/modules/environment/init.zsh
+++ b/modules/environment/init.zsh
@@ -39,7 +39,7 @@ setopt RC_QUOTES            # Allow 'Henry''s Garage' instead of 'Henry'\''s Gar
 unsetopt MAIL_WARNING       # Don't print a warning message if a mail file has been accessed.
 
 # Allow mapping Ctrl+S and Ctrl+Q shortcuts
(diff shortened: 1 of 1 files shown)


belak:
Thanks for following up on this. Makes sense.

belak:
So, there seem to be conflicting reasons to allow ^Q/^S to be rebound. Some want it everywhere, some want it just in ZSH. It sounds like you want it to just be disabled in ZSH, while many people want it everywhere... Personally I prefer it disabled everywhere because it means that I can't accidentally freeze a console program but I can see the argument for the other side.

Maybe we should only use flow_control rather than stty as it still allows the user to call stty later. Or maybe it makes more sense to continue to use stty as a user who wants that back can always re-enable it. I'm really on the fence about this but would be option to suggestions.

PR #1786, opened 2020-01-20
utility: add support for g-prefixed coreutils ls

### LINKED ISSUE #1758: Feature request to make gls work on macos with coreutils
After installing coreutils via below command

$> brew install coreutils

'gls' and 'gdircolors' are available. So if I do below colors and directory 

$> git diff
diff --git a/modules/utility/init.zsh b/modules/utility/init.zsh
index a2e5c2b..4400cf0 100644
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -73,6 +73,12 @@ if zstyle -T ':prezto:module:utility' safe-ops; then
 fi

 # ls
+
+if is-callable 'gdircolors'; then
+  alias dircolors='gdircolors'
+  alias ls="${aliases[ls]:-gls}"
+fi
+
 if is-callable 'dircolors'; then
   # GNU Core Utilities

Can you check if coreutils has been installed ?

belak:
Fixes #1758

## Proposed Changes

  - Define g-prefixed ls aliases
  - Default to g-prefixed ls in aliases

Note that this does change default behavior a bit by changing which ls will be selected for the default alias.

### DIFF
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -72,28 +72,41 @@ if zstyle -T ':prezto:module:utility' safe-ops; then
   alias ln="${aliases[ln]:-ln} -i"
 fi
 
-# ls
-if is-callable 'dircolors'; then
-  # GNU Core Utilities
+function -define-coreutils-aliases {
+  # Prefix will either be g or empty. This is to account for GNU Coreutils being
+  # installed alongside BSD Coreutils
+  local prefix=$1
 
   if zstyle -T ':prezto:module:utility:ls' dirs-first; then
-    alias ls="${aliases[ls]:-ls} --group-directories-first"
+    alias ${prefix}ls="${aliases[${prefix}ls]:-${prefix}ls} --group-directories-first"
   fi
 
   if zstyle -t ':prezto:module:utility:ls' color; then
     # Call dircolors to define colors if they're missing
     if [[ -z "$LS_COLORS" ]]; then
       if [[ -s "$HOME/.dir_colors" ]]; then
-        eval "$(dircolors --sh "$HOME/.dir_colors")"
+        eval "$(${prefix}dircolors --sh "$HOME/.dir_colors")"
       else
-        eval "$(dircolors --sh)"
+        eval "$(${prefix}dircolors --sh)"
       fi
     fi
 
-    alias ls="${aliases[ls]:-ls} --color=auto"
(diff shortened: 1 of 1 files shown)


casaper:
@wadkar 

I myself have not installed gnu coreutils yet, because I wasn't aware of this being available on my mac. So you've lifted me to the idea here, thanks! 👍 

But seeing this brings me to the idea to look if it'd make sense to do something similar with darwin users having gnu- programmes installed.

I for example, have `gnu-sed`, `gnu-tar` and `gnu-getopt` because I really don't like their standard darwin counterparts.

belak:
This actually already exists in https://github.com/sorin-ionescu/prezto/tree/master/modules/gnu-utility.

On a similar note, maybe it would make more sense to only change to the g-prefixed tools if gnu-utility is enabled.

modules/utility/init.zsh
```diff
@@ -72,28 +72,41 @@ if zstyle -T ':prezto:module:utility' safe-ops; then
...
+
+# ls
+
+# GNU Core Utilities
+if is-callable 'gdircolors'; then
+  -define-coreutils-aliases g
+  alias ls="${aliases[gls]:-gls}"
+
+elif is-callable 'dircolors'; then
+  -define-coreutils-aliases
```
indrajitr, line 108: Shouldn't we consider honoring user configured `':prezto:module:gnu-utility'` zstyle instead of assuming `g`?
Granted Homebrew doesn't honor customizing `g` prefix for gnu coreutils [anymore](https://brew.sh/2019/02/02/homebrew-2.0.0/), but macport still does.

Besides, if a user loads `gnu-utility` module, both `ls` and `gls` would be candidate for alias modification. Shouldn't we support both?

indrajitr: FWIW, it might be tricky with some of the `g_` prefixed GNU utility commands/aliases -- particularly the ones used with `git` module (`gls` is aliased to `git log` for example).

belak: Good catches. I'm not 100% sure what the proper fix is on this... coreutils are a bit of a pain to detect and work with especially with the git aliases.

indrajitr: Took another (rather opinionated) attempt at addressing this here #1917. Criticism/comment appreciated :)

PR #1793, opened 2020-02-13
feat(helper): add os-type helper functions

casaper:
## Add `$OSTYPE` functions

Add the following functions checking for the `$OSTYPE`:

- is-bsd
- is-cygwin
- is-darwin
- is-linux

And apply them everywhere in prezto I found code doing that.

Everywhere but in [zprofile](runcoms/zprofile), because I presume zprofile will not have helpers available.  
Is that correct?

## Reasons

1. I personally find functions like this just more readable in code.
1. If a test for OSTYPE xy proves to not cover some edge case one day, there will be only one place to fix it.
1. I just use Prezto for quite some time, and if I want to start contributing, I ought to be with small helper utilities like these.

## Questions

1. I'm not sure if that kind of thing is wanted in prezto?
1. Is [helper](modules/helper/init.zsh) the right place for this?
1. Will [helper](modules/helper/init.zsh) be available to all those places I applied my helpers? (will it work there)

### DIFF
--- a/modules/haskell/init.zsh
+++ b/modules/haskell/init.zsh
@@ -11,7 +11,7 @@ if (( ! $+commands[ghc] )); then
 fi
 
 # Prepend Cabal per user directories to PATH.
-if [[ "$OSTYPE" == darwin* && -d $HOME/Library/Haskell ]]; then
+if is-darwin && [[  -d $HOME/Library/Haskell ]]; then
   path=($HOME/Library/Haskell/bin(/N) $path)
 else
   path=($HOME/.cabal/bin(/N) $path)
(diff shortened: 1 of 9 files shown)


casaper:
@johnpneumann && @jeffwidman 

I hope I did not ask the wrong guys for reviews here. If I'm doing things wrong for you, please don't hesitate to tell me where or to who else to go to! :wink:

belak:
This seems reasonable - is there any reason this was closed?

casaper:
> This seems reasonable - is there any reason this was closed?

I just have a habit of closing PR's if I don't see a probability that they're going to get processed in any reasonable time. This is because a lot of lagging not very important PR's will even rise the reason for this happening, as it is distracting from others.

But feel free to reopen it and if commonly agreed on merge it....

belak:
This looks good to me and seems to be a net positive in readability. Thanks!

PR #1797, opened 2020-03-02
use $TTY instead of $(tty); it's faster and more reliable

romkatv:
`tty` command infers the current TTY from file descriptor 0. `$TTY` gives current TTY directly. The latter works even when file descriptor 0 is redirected. It's also over 1000 times faster.

### DIFF
--- a/modules/gpg/init.zsh
+++ b/modules/gpg/init.zsh
@@ -26,7 +26,7 @@ if [[ -z "$GPG_AGENT_INFO" && ! -S "${GNUPGHOME:-$HOME/.gnupg}/S.gpg-agent" ]]; 
 fi
 
 # Inform gpg-agent of the current TTY for user prompts.
-export GPG_TTY="$(tty)"
+export GPG_TTY=$TTY
 
 # Integrate with the SSH module.
 if grep '^enable-ssh-support' "$_gpg_agent_conf" &> /dev/null; then


jeffwidman:
> tty command infers the current TTY from file descriptor 0. $TTY gives current TTY directly. The latter works even when file descriptor 0 is redirected. It's also over 1000 times faster.

Is any of this documented anywhere in the ZSH docs? I couldn't find it and it'd be nice to include a link in the commit description...

romkatv:
`TTY` is documented in `man zshparam`.

> #### PARAMETERS SET BY THE SHELL
> 
> `TTY`    The name of the tty associated with the shell, if any.

`tty` is documented in `man tty`.

> `tty` - print the file name of the terminal connected to standard input

romkatv:
Let me know if you want me to include commands `man tty` and `man zshparam` in the commit description.

jeffwidman:
This is good enough for me. Thanks!

jeffwidman:
Hah, just checked if there were any other instances of this to cleanup in teh code, and looks like we already use `$TTY` all over the place: https://github.com/sorin-ionescu/prezto/search?q=%24%28tty%29&unscoped_q=%24%28tty%29

romkatv:
Thanks for the quick merge, Jeff!

Issue #1493, opened 2017-10-23
[Cygwin] Prezto makes zsh hang at exit


vtavernier:
### Description

On Cygwin, when zsh is used with prezto, exiting the shell (via exit or Ctrl+D) hangs the shell, and the process has to be terminated. At least `sorin` and `pure` themes are affected, even using the default configuration.

### Expected behavior

When exiting the shell via exit or Ctrl+D, the shell process terminates normally.

### Actual behavior

When exiting the shell via exit or Ctrl+D, the zlogout message is displayed but the process gets stuck without exiting.

### Steps to Reproduce

1. Install cygwin with the zsh package (currently 5.3-1, but 5.1.1-1 is affected too)
2. Install prezto by following the instructions
3. Open a shell and try to exit it via exit or Ctrl+D. It should hang, and the terminal window should stay open.

### Git bisect result

This issue was not happening before I updated my prezto working copy. According to a git-bisect between the revisions before and after the update, this is the first revision affected by this issue:

```
9bdc1b35d51407515edb5e82dd3f8635fd771511 is the first bad commit
commit 9bdc1b35d51407515edb5e82dd3f8635fd771511
Author: Kaleb Elwert <belak@coded.io>
Date:   Mon Jul 24 11:55:02 2017 -0700

    Migrate sorin prompt to zsh-async (#1385)

    This includes some improvements by @indrajitr in addition to the main migration.

    The first step was to avoid PROMPT and RPROMPT modification when possible (which may help resolve some other issues as well relating to zsh crashes with the sorin prompt) then update the displayed git information in a separate variable rather than a command.

    We use zsh-async for creating and running background tasks. The sorin prompt uses it to update git info without blocking the prompt from displaying (because of how long it can take). In the future it may be worth moving more tasks and more prompts to using this.

    The move to zsh-async does make the git prompt slower in some circumstances (most noticeable in large repos), but this is a worthwhile tradeoff to avoid the cache file which had a number of potential security holes.

    We have also switched to adding zsh-async as an external submodule (rather than the version bundled with pure) which may cause some migration headaches, but it will be worth it in the long run.

:100644 100644 46e6a0e69d7da7a7610cb7b85816ee96d5556da7 365b059226a226575fa09bae58691e60406e908c M      .gitmodules
:040000 040000 1a718dd9075fc540c8d59fa584c157a1434b7678 a6ab330ba6ca36d39ec83eb468f84d4d326f0491 M      modules
```

### Versions

[...]

PR #1805, opened 2020-03-19
Switch to an alternative method of async git in the sorin prompt

### LINKED ISSUE #1795: async_job:zpty:12 no such pty command: prompt_sorin
Same issue as #1736 but on sorin prompt

  - Prezto commit: e9f281cd6f62983f44faaae0816866d13bc1474e
  - ZSH version: 5.8 (x86_64-pc-linux-gnu)
  - OS information: Arch Linux 5.5.3

belak:
Fixes #1795, maybe #1493

## Proposed Changes

- Update sorin prompt to use zle callbacks rather than zsh-async

This should fix a number of random, hard to debug issues we've had with zsh-async and may improve Windows support. It doesn't get updates super often either so it may be worth investigating removing it completely if we can.

The code here is based off a simplified version of the code in autosuggestions mentioned here: https://github.com/mafredri/zsh-async/issues/24

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -32,29 +32,25 @@
 # Load dependencies.
 pmodload 'helper'
 
-function prompt_sorin_async_callback {
-  case $1 in
-    prompt_sorin_async_git)
-      # We can safely split on ':' because it isn't allowed in ref names.
-      IFS=':' read _git_target _git_post_target <<<"$3"
+function prompt_sorin_async_git_callback {
+  # We can safely split on ':' because it isn't allowed in ref names.
+  IFS=':' read _git_target _git_post_target <<<"$3"
 
-      # The target actually contains 3 space separated possibilities, so we need to
-      # make sure we grab the first one.
-      _git_target=$(coalesce ${(@)${(z)_git_target}})
+  # The target actually contains 3 space separated possibilities, so we need to
+  # make sure we grab the first one.
+  _git_target=$(coalesce ${(@)${(z)_git_target}})
 
-      if [[ -z "$_git_target" ]]; then
-        # No git target detected, flush the git fragment and redisplay the prompt.
-        if [[ -n "$_prompt_sorin_git" ]]; then
-          _prompt_sorin_git=''
-          zle && zle reset-prompt
-        fi
-      else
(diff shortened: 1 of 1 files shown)


modules/prompt/functions/prompt_sorin_setup
```diff
@@ -32,29 +32,25 @@
...
-      if [[ -z "$_git_target" ]]; then
-        # No git target detected, flush the git fragment and redisplay the prompt.
-        if [[ -n "$_prompt_sorin_git" ]]; then
-          _prompt_sorin_git=''
-          zle && zle reset-prompt
-        fi
-      else
-        # Git target detected, update the git fragment and redisplay the prompt.
-        _prompt_sorin_git="${_git_target}${_git_post_target}"
-        zle && zle reset-prompt
```
jeffwidman: So are we already using both `zle` _and_ `zsh-async`? It does seem weird that we'd be using two async frameworks...

belak:
> Which doesn't get updates? `zsh-async` or `zle`

zsh-async. There are people saying reverting to 1.1 fixes the issue but that’s a really old release and there have been a ton of fixes since then.

Yeah, there are obviously some bugs... it was a quick implementation thrown together in an hour and it’s hard for me to debug this because I don’t have a windows machine available at the moment.

modules/prompt/functions/prompt_sorin_setup
belak: zle is a zsh built-in and does a ton of stuff related to line editing. For some reason there's also some functionality in there around managing file descriptors for sub-shells.

jeffwidman:
@belak Is this still relevant given that #1810 appears to fix the root issue? Or were there other problems with `zsh-async` such that you want to continue pursuing this?

belak:
As far as I know, the fix for zsh-async still doesn't fix the zsh hang on exit with windows because (by design) there's a process still running to handle git information. I'm not sure if this is worth pursuing or not.

jeffwidman:
Gotcha... yeah, Idk either... Microsoft/windows has been changing a lot to be more linux-friendly, I would probably vote to let this sit a bit and see if they provide easier hooks to do that. This would be years in the future, so if someone urgently wanted it and came up with a way, I wouldn't be opposed, but otherwise not something I'd pursue...

PR #1810, opened 2020-03-29
Add fixes for #1795 (prompt_sorin)

### LINKED ISSUE #1795: async_job:zpty:12 no such pty command: prompt_sorin

ackerleytng:
Fixes #1795 

A similar fix will work for #1723 

The issue here is that when the async worker dies for whatever reason, `prompt_sorin_async_tasks` doesn't know about it and calls `async_job` on a worker that is not existent.

Between zsh-async v1.1.0 and v1.7.0, the change that made this difference is that in the newest version v1.7.2, zsh-async (I believe, correctly) checks for an error message (`hup`, `nval` or `err`) and stops the worker. In v1.1.0, it did not stop the worker, so `_async_zle_watcher` ends up calling `async_process_results` many, many times in an infinite loop, causing zsh to take up 100% CPU.

## Proposed Changes

  - Reinitialize the worker on the last callback from `_async_zle_watcher`
  - Also always start the worker in `prompt_sorin_async_tasks` (Initializing the worker is handled idempotently by zsh-async)

### DIFF
--- a/modules/prompt/functions/prompt_sorin_setup
+++ b/modules/prompt/functions/prompt_sorin_setup
@@ -54,6 +54,13 @@ function prompt_sorin_async_callback {
         zle && zle reset-prompt
       fi
       ;;
+    "[async]")
+      # If something happend to the worker, prompt_sorin_async_callback will be
+      # called one last time. In this case, we look out for `hup` or `err` and
+      # restart the worker
+      [[ $5 = *'returned error hup'* ]] || [[ $5 = *'returned error err'* ]] \
+        && prompt_sorin_initialize_worker
+      ;;
   esac
 }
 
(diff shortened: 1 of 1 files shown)


ackerleytng:
The error can be reproduced reliably this way:

1. Start zsh
2. See that zsh has one child (probably the async manager)
3. Kill the child with kill -9
4. Press enter in the zsh where that child was killed

Now if you switch to v1.1.0, the message goes away, but the zsh instance whose child was killed goes into some kind of infinite loop and takes 100% CPU.

Got a hint here from straceing the zsh process with an infinite loop, and seeing lots of stuff related to WINCH and /dev/null.

I understand that #1805 might fix this issue, didn't try that yet. This proposal also explains the root cause!

belak:
Thanks for doing more research on this! I've been unsuccessfully been trying to nail this down for quite some time.

jeffwidman:
Nice job @ackerleytng

ackerleytng:
I'm so sorry I meant a similar fix will work for #1736, the prompt_pure version of the issue.

Eventually I actually ported code from the latest version of prompt_pure, so this error probably wouldn't come up for prompt_pure.

ackerleytng:
This is the commit from zsh-async that checks and stops the worker.

https://github.com/mafredri/zsh-async/commit/43de5e0738eaf366dfde4d0a204de655de16f18b

And the fix I ported was from this commit

https://github.com/sindresorhus/pure/commit/0e3e7770d60df4bbe3071692c9898ca33d526ae6

This was committed on January 13 last year, but we've been using an older version of prompt_pure in prezto until @synasius noticed this on Feb 6 https://github.com/sorin-ionescu/prezto/issues/1736#issuecomment-582785876

So it matches up!

ysoftman (commented):
👍

PR #1816, opened 2020-04-06
fix: regression in homebrew module

### LINKED ISSUE #1815: homebrew module regression
Commit f4ca9ebfc913453f98ba6912a8c42684fd742cc1 introduced a regression in the homebrew module where it won't load.

Originally the bail-out code was:
```zsh
if [[ "$OSTYPE" != (darwin|linux)* ]]; then
  return 1
fi
```

The regression is:
```zsh
if ! is-darwin || ! is-linux; then
  return 1
fi
```

Fix:
```zsh
if ! is-darwin && ! is-linux; then
  return 1
fi
```

srijanshetty:

Fixes #1815

## Proposed Changes
  - Correct use of DeMorgan's law

### DIFF
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -6,7 +6,7 @@
 #
 
 # Return if requirements are not found.
-if ! is-darwin || ! is-linux; then
+if ! is-darwin && ! is-linux; then
   return 1
 fi
 


belak:
Good catch! Thanks for submitting this!

PR #1819, opened 2020-04-08
don't dowload whole history of powerlevel9k and powerlevel10k submodules

vbauerster:
Cloning prezto's submodules takes considerable amount of time. This is especially noticeable when setting up environment on remote host over SSH with slow connection. Moreover it's really annoying as you realize that time spent for stuff that you never use.
With this PR git will fetch only required shallow commit, as described [here](https://stackoverflow.com/questions/30129920/git-submodule-without-extra-weight/38895397#38895397), thus saving time over SSH with slow connection.
I selected only `powerlevel` submodules because they're pretty fat.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,6 +28,8 @@
 [submodule "modules/prompt/external/powerlevel9k"]
 	path = modules/prompt/external/powerlevel9k
 	url = https://github.com/bhilburn/powerlevel9k.git
+	shallow = true
 [submodule "modules/prompt/external/powerlevel10k"]
 	path = modules/prompt/external/powerlevel10k
 	url = https://github.com/romkatv/powerlevel10k.git
+	shallow = true


belak:
Makes sense. Thanks!

jeffwidman:
IMO we should be shallow-cloning _all_ submodules.

vbauerster:
@jeffwidman Unfortunately some submodules have their own submodules which hardly ever use `shallow = true`. For example [powerlevel9k](https://github.com/Powerlevel9k/powerlevel9k) has [shunit2](https://github.com/kward/shunit2) submodule dependency which whole history is about 69 MiB. So `git submodule update --init --recursive` will still fetch whole history of sub-submodules.

jeffwidman:
@romkatv ☝️

romkatv:
Enabling shallow cloning of powerlevel10k is a good idea. Powerlevel10k documentation lists 12 installation options and only two of them pull full git history. With this PR there will be just one (zimfw doesn't have an option to enable shallow cloning).

PR #1825, opened 2020-04-17
Add Elapsed time module

Salamandar:
Add a module Elapsed time, that prints the run time of commands, just like the Paradox theme, but for any theme !

The corresponding code could be removed from the Paradox theme.

### DIFF
--- a/modules/elapsed-time/init.zsh
+++ b/modules/elapsed-time/init.zsh
@@ -1,0 +1,36 @@
+#
+# Provides elapsed time information
+#
+# Authors:
+#   Salamandar <felix@piedallu.me>
+#
+
+_elapsed_time_start_time=$SECONDS
+
+_elapsed_time_min_seconds=5
+
+function elapsed-time-start {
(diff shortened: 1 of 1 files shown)


Salamandar:
This code could be updated, i.e to change the granularity (milliseconds) or to add a function that does not printf, but returns the string to be printed by the theme.

belak:
Thanks for submitting this! Unfortunately, we're not currently accepting modules in prezto-core. However, if you want to re-submit this to [prezto-contrib](https://github.com/belak/prezto-contrib) we'd be happy to accept it there.

Salamandar:
OK, gonna do that. :)

indrajitr:
Actually, after thinking a bit about it, this functionality is after all just a zsh hook and probably doesn’t require a full blown module. 

Instead it could be refactored out into a common function (like `prompt-pwd`) and used by other themes.

Salamandar:
Can you point to me an example ? I could do a PR too.
Also, I don't know if you saw it, but my PR was accepted is prezto-contrib.

PR #1826, opened 2020-04-22
homebrew: load helper dependency

zmb3:
f4ca9ebfc913453f98ba6912a8c42684fd742cc1 introduced the `is-darwin` and `is-linux` helpers, but called them without first loading the module.

Updates #1815


srijanshetty:
Yup, missed the load of module.

PR #1827, opened 2020-04-23
Matching submodule name with path

bXV0ZQ:
`pure` submodule name doesn't match its path in `.gitmodules`

`gSx` alias (`git-submodule-remove`) fails due to this unusual configuration.

## Proposed Changes

  - Rename submodule to match its path as all others

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,7 +16,7 @@
 [submodule "modules/prompt/external/agnoster"]
 	path = modules/prompt/external/agnoster
 	url = https://github.com/agnoster/agnoster-zsh-theme.git
-[submodule "modules/prompt/functions/pure"]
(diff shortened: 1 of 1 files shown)


indrajitr:
Interesting situation. If we really have to update the path, we have to instruct users to do a `git submodule sync` (maybe with `--recursive`).
Additionally, `zprezto-update` will have to be updated.

jeffwidman:
We should probably go ahead and bit the bullet on this one, especially since we've got a few other deprecations/migrations in-flight.

@bXV0ZQ do you want to add the changes to `zprezto-update` suggested by @indrajitr ?

indrajitr:
Cherry-picked commits from @bXV0ZQ and raised updated PR: #1915.

PR #1830, opened 2020-04-26
Do not source history-substring-search if already loaded

amynbe:
Split history-substring-search/init.zsh in 2 files to allow reuse outside Prezto

I'm using the Zinit plugin manager and would like to use this module for the sane default keybindings it provides. However, I can't so so because init.zsh tries to source history-substring-search from an external git repo which makes the script fail.

## Proposed Changes
Split history-substring-search/init.zsh in 2 files : init.zsh and configure.zsh

init.zsh now only contains code to source dependencies, namely the editor module, the history-substring-search external repo and configure.zsh

configure.zsh can then be used in a standalone manner by other plugin managers.


indrajitr:
Refactoring things like these are cool. However, consistency is more important as we'll all agree :)

@amynbe, the approach you are taking is `<module>/init.zsh` would just be the entry point which would call other `*.zsh` for functions, configurations, etc.

This hasn't been the case so far with prezto. If we take this route, we should consider doing this consistently across all modules.

What say @belak, @johnpneumann, @jeffwidman ?

srijanshetty:
I agree with @indrajitr. Also @amynbe it would make more sense to create a separate plugin for your usecase as you're trying to use a bundled prezto plugin with a separate plugin infrastructure. There won't be any gurantees that your plugin might again break in the future?

Maintaining it separating will benefit a lot for you.

amynbe:
Fair enough, and thanks to both of you for kindly review this PR, but in my opinion, given this module has last been updated 3 years ago, I don't expect any breaking changes to happen in the near future. 

Zinit plugin manager offers this "snippet loading" feature where you can just load scripts straight from OMZ, Prezto, or any git repo, and it's working great for many modules so far. The convenience outweighs the risk IMO: after all, breaking changes can happen even between plugins maintained on separate repositories.

I'm willing to take the risk anyway, rather than copy and maintain the same code elsewhere that someone already kindly wrote, shared and maintained on a popular repository.

I don't see that consistency of code convention is a strong argument against a refactoring that is quite safe and easy to understand. This approach with a single `init.zsh` containing both the `source` directive and the additional Prezto config, I don't think it's really a convention, just people have been doing this by convenience since the time before plugin managers became popular. So if the need to split other modules arises in the future, why not do it on a case-by-case basis?

srijanshetty:
Fair enough @amynbe. You arguments make perfect sense. I did not know the internal functioning of zinit package manager. I'll go through your PR once, I'll still wait for some other maintainers to comment though.

jeffwidman:
I don't feel strongly either way on this one... it's interesting idea, and I see the value, but it is an additional layer of indirection and we would want to move toward doing this for all modules... 

So I am somewhat open to this, with two caveats:
1. Understanding that while this is essentially creating an API that may be used outside of the `prezto` framework, that is a completely unofficial API that could break at any time. Ie, don't yell at us if we forget and later changes break things... just submit a PR and remind us. 
2. here this layer of indirection is simple... there may be times where its not so simple, and not worth the obfuscation, so even if we move toward this module, likely not _every_ module will get it


I am also curious to hear @johnpneumann and @belak 's thoughts....

belak:
My preferred solution would be to only load zsh-history-substring-search if it's not already loaded (or provide an option to override it). This seems like it would fit both use cases - simpler to maintain and allowing loading outside prezto.

I believe the check would just be `if $(( ! $+functions[history-substring-search-up] )); then source files; fi` (or something similar).

amynbe:
Commit added as per @belak 's suggestion. It works well too.
Thanks for the suggestion Kaleb.

indrajitr:
@amynbe Can you please squash the two commits into a single one?

belak:
@indrajitr We can squash on merge - that's what I usually do :)

belak:
Thanks for submitting this!

PR #1842, opened 2020-05-26
Make .zcomp* location configurable

### LINKED ISSUE #1841: Make .zcomp* location configurable
Would you consider a PR to make the `.zcomp*` location configurable like `HISTFILE`?

The motive is my [zsh-bootstrap](https://github.com/texastoland/zsh-bootstrap) project that installs Prezto with zgen using Homebrew. It overrides `ZDOTDIR` with a set of rc files that delegate to Prezto's  runcoms and checks overrides in `XDG_CONFIG_HOME`. The idea is one-click install, easier customization, and more standardized integration with other tools.

It already works great locally. The only problem is Homebrew installs packages in `/usr/local` land. I already source from `XDG_CONFIG_HOME` and can move `HISTFILE` there too. The only roadblock is the cache files are hardcoded to `${ZDOTDIR:-HOME}`. I don't foresee any backwards compatibility issues. Too novel a change?

The only relevant discussion I could find was https://github.com/sorin-ionescu/prezto/issues/255#issuecomment-8215163.

texastoland:
Closes #1841

## Proposed Changes

Look up `zcompdump` and `zcompcache` <ins>and all other cache files</ins> in `XDG_CACHE_HOME` ~if set, otherwise use the zsh default~. I provided additional detail in the linked issue and have tested as thoroughly as I can locally.

### DIFF
--- a/modules/command-not-found/init.zsh
+++ b/modules/command-not-found/init.zsh
@@ -21,7 +21,7 @@ elif (( $+commands[brew] )); then
     "$(brew --repository 2> /dev/null)"/Library/Taps/*/*/cmd/brew-command-not-found-init(|.rb)(.N)
   )
   if (( $#cnf_command )); then
-    cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/zsh/prezto-brew-command-not-found-cache.zsh"
+    cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/brew-command-not-found-cache.zsh"
 
     if [[ "${${(@o)cnf_command}[1]}" -nt "$cache_file" \
           || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
--- a/modules/command-not-found/init.zsh
+++ b/modules/command-not-found/init.zsh
@@ -21,11 +21,12 @@ elif (( $+commands[brew] )); then
     "$(brew --repository 2> /dev/null)"/Library/Taps/*/*/cmd/brew-command-not-found-init(|.rb)(.N)
   )
   if (( $#cnf_command )); then
-    cache_file="${TMPDIR:-/tmp}/prezto-brew-command-not-found-cache.$UID.zsh"
+    cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/zsh/prezto-brew-command-not-found-cache.zsh"
 
     if [[ "${${(@o)cnf_command}[1]}" -nt "$cache_file" \
           || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
           || ! -s "$cache_file" ]]; then
+      mkdir -p "$cache_file:h"
       # brew command-not-found-init is slow; cache its output.
       brew command-not-found-init >! "$cache_file" 2> /dev/null
     fi
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -33,15 +33,15 @@ unsetopt FLOW_CONTROL      # Disable start/stop characters in shell editor.
 # shell is opened each day.
 autoload -Uz compinit
 _comp_path="${XDG_CACHE_HOME:-$HOME/.cache}/zsh/zcompdump"
-_comp_files=($_comp_path(Nm-20))
-if (( $#_comp_files )); then
+# #q expands globs in conditional expressions
+if [[ $_comp_path(#qNmh-20) ]]; then
   # -C (skip function check) implies -i (skip security check).
   compinit -C -d "$_comp_path"
 else
   mkdir -p "$_comp_path:h"
   compinit -i -d "$_comp_path"
 fi
-unset _comp_path _comp_files
+unset _comp_path
 
 #
 # Styles
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -32,8 +32,7 @@ unsetopt FLOW_CONTROL      # Disable start/stop characters in shell editor.
 # cache time of 20 hours, so it should almost always regenerate the first time a
 # shell is opened each day.
 autoload -Uz compinit
-# Respect XDG_CACHE_HOME if set, otherwise use zsh default.
-_comp_path="${${XDG_CACHE_HOME:+$XDG_CACHE_HOME/zcompdump}:-${ZDOTDIR:-$HOME}/.zcompdump}"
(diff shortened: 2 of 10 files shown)


belak:
I'm not opposed to this, but is there a reason for avoiding ZDOTDIR?

texastoland:
I explained a bit the linked issue. The short version is I made Prezto Homebrew-installable. I commander `ZDOTDIR` and delegate user configuration to `XDG` paths as seen in https://github.com/texastoland/zsh-bootstrap/commit/7a593cb945cb2ecbc63ce23a29d3de8a7f84e2a7. That way `ZDOTDIR`, Prezto, and zgen can be source from `/usr/local`.

belak:
I'm fine accepting this as-is, but I think we also have some separate caching for a number of the modules. It looks like we use the `TMPDIR` for those... what are your opinions? Would it make sense to move those as well?

texastoland:
Wow I didn't consider `TMPDIR`. It appears in 8 modules besides external prompts. All usages separate user caches by `UID` instead of random names using `mktemp`. I found a [really thorough article](http://0pointer.net/blog/projects/tmp.html):

> 4. You need a place to put cache files which should be persistent and run unprivileged: use $XDG_CACHE_HOME.
> 5. Nothing of the above applies and you need to place a small file that needs no persistency: use $TMPDIR with a fallback on /tmp. And use mkstemp(), and mkdtemp() and nothing homegrown.

I think these all fall squarely into `XDG_CACHE_HOME`. I agree consistency is more important. Changing them would be a straightforward substitution:

```diff
- "${TMPDIR:-/tmp}/blabla-$UID"
+ "${XDG_CACHE_HOME:-/$HOME/.cache}/blabla"
```

Feedback wanted:
1. Should the completions still default to `ZDOTDIR`?
2. I just added a missing `mkdir -p` or the caches never got created. Should we do that just once on init?

texastoland:
@belak Done but I don't use these modules for a sanity test. I documented the regex I used to update the paths in the commit message and just checked if `mkdir -p` was needed.

texastoland:
@belak 🚀

texastoland:
@belak Everything seems to be working locally 🚀

texastoland:
I'm in the process of switching to OMZ. I noticed it uses `ZSH_CACHE_DIR` pervasively. I could implement that here. They default to a folder inside their package which seems like a common choice among zsh tools. I still think `XDG_CONFIG_HOME` is more useful.

belak:
Sorry I haven't had the time to take a closer look at this lately - I've got quite a bit going on in my personal life and at work.

The main reason I haven't moved on this yet is that I don't know how it will work with nix (or even if it will) and don't have a setup to test with. At this point, I'm tempted to just push it through and fix any issues that come up, but that's not nice to our users.

Hoping to get to this soon, but unfortunately I can't make any promises just yet.

texastoland:
No rush! I don't know nix at all. From a quick search, XDG dirs have been configurable since NixOS/nixpkgs#67389. Take your time, though. I'd switch your local branch and make sure your `~./cache` looks fine after a while. 

Also note there was an inconspicuous bugfix:

https://github.com/texastoland/prezto/blob/46159bd527ff25a58a2e633fcbc9074423256808/modules/completion/init.zsh#L37.

A missing `h` modifier caused the cache to live 20 days instead of 20 hours since it was implemented three years ago.

belak:
Looks good to me then. Thanks again for doing the legwork on this!

texastoland:
I'm considering rewriting `zsh-help`. It doesn't work on Mac. I fixed the bugs and noticed it's replaceable with a 1-liner (or few). I'll try to sen it next week!

hotpxl:
Is there any reason to use $HOME instead of $TMPDIR for stuff like `ssh-agent.sock`?

For me personally I work on a cluster environment where the home directory is mounted on a NFS, so it's definitely not doing the right thing anymore. The sock file should be local to each host.

And in general, we shouldn't rely on $HOME for storing local temporary files, right?

belak:
It's a challenge balancing different systems and setups. We defaulted to an approximation of `XDG_CACHE_DIR` which is `~/.cache` by default. I understand why it's breaking for you, but unfortunately don't have a great solution. `TMPDIR` isn't standard (one of the reasons for moving away from it), `XDG_CACHE_DIR` is more of a standard, but isn't always in a convenient place.

Do you have any ideas on how to make this better?

texastoland:
I think it'd be reasonable to implement `ZSH_CACHE_DIR` like OMZ. @hotpxl One problem with `TMP_DIR` can be seen in the diff. It potentially exposes user data and by convention appends `UID` to prevent collisions. I'm willing to PR if @belak is interested.

belak:
I honestly don't know what the correct solution here is... `/tmp` isn't ideal because it potentially leaks information, `$HOME/anything` isn't ideal in the case of shared home directories, at least for things like sockets.

I agree that using `$HOME` for things like the ssh-agent is not expected behavior. Does it make sense to put everything in `$TMPDIR` with a UID suffix somewhere? If everything is updated to use `$TMPDIR` similar to how it did before, then it would at least be possible to override with $HOME/.cache if a user wants to.

texastoland:
I think the XDG directories are the right default. I just propose making it configurable so someone could set `ZSH_CACHE_DIR=$TMPDIR/prezto.$UID`. How many of the changes produce unexpected behavior? I'd be happy to look for a solution.

hotpxl:
> I think the XDG directories are the right default. I just propose making it configurable so someone could set `ZSH_CACHE_DIR=$TMPDIR/prezto.$UID`. How many of the changes produce unexpected behavior? I'd be happy to look for a solution.

This makes sense. I'm all for making this configurable. But with the current setup I'm just setting `XDG_CACHE_HOME` since none of my other applications use it.

hotpxl:
I'd vote for a more explicit design:

We declare a single cache directory variable like `ZSH_CACHE_DIR` in `zprofile` or `zpreztorc` (using `zstyle`), which defaults to `${XDG_CACHE_HOME:-$HOME/.cache}`. And then everything else just references this `$ZSH_CACHE_DIR`, instead of having each function fall back to `$HOME/.cache` separately. This seems more manageable and easy to reason about.

texastoland:
@hotpxl I'm about to travel for my birthday, but I'll tag you in a PR later!

hotpxl:
> @hotpxl I'm about to travel for my birthday, but I'll tag you in a PR later!

Thanks, and happy birthday!

PR #1848, opened 2020-06-01
Ask pyenv whether pip exists if installed

### LINKED ISSUE #1846: On Ubuntu 20.04 with pyenv, the system version has only pip2/pip3 and the Python module fails
### Description
On Ubuntu 20.04, a globally installed `pip` command no longer points to Python 2's pip. Instead, there is only `pip2` and `pip3`. However, after installing pyenv, a shim is installed for `pip`. Running `pip` with the system interpreter selected will fail with a message to that effect. However, the command does technically exist so the check `$commands[pip]` will return true.

This causes [the pip command detection in the Python module](https://github.com/sorin-ionescu/prezto/blob/cb69d2d3e471726a76ac3e4e484deac52a2abf56/modules/python/init.zsh#L154) to select the wrong pip command.

### Expected behavior

The Python module works without errors and pip completion is loaded.

### Actual behavior

This error message:
```
pyenv: pip: command not found

The `pip' command exists in these Python versions:
  [other versions]

Note: See 'pyenv help global' for tips on allowing both
      python2 and python3 to be found.
```

And pip shell completion is not loaded.

### Steps to Reproduce

1. Install pyenv: `curl https://pyenv.run | bash`
2. Install any Python version through pyenv: `pyenv install 3.8.3`
3. Install shims: `pyenv rehash`
4. Set the system interpreter as default: `pyenv global system`
5. Restart zsh: `exec zsh`

### Discussion

We coulde change the glob pattern so `pip3` and `pip2` get preference. However, this seems just as bad in the opposite direction.

If pyenv is installed, `pyenv which` could be used instead as a test: `pyenv which pip >/dev/null` returns false. Pyenv will set `PYENV_ROOT` (if the user did not set it), so we know when to use it.

Would a PR be welcome to fix this?

### Versions

  - Prezto commit: cb69d2d3e471726a76ac3e4e484deac52a2abf56
  - ZSH version: 5.8 (x86_64-ubuntu-linux-gnu)
  - OS information: Ubuntu 20.04

jcassee:
Fixes #1846

## Proposed Changes

  - Check pyenv for the existence of pip if it is installed.

Pyenv will install shims for commands that exist in any interpreter, even if it is not the current one. This means that a command may technically exist, but when executed will tell the user to try a
different interpreter. The original check for pip in the Python module can fail for this reason, in particular on Ubuntu 20.04.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -147,18 +147,25 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
 fi
 
 # Load PIP completion.
-if (( $#commands[(i)pip(|[23])] )); then
-  cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/pip-cache.zsh"
-
-  # Detect and use one available from among 'pip', 'pip2', 'pip3' variants
+# Detect and use one available from among 'pip', 'pip2', 'pip3' variants
+if [[ -n "$PYENV_ROOT" ]]; then
+  for pip in pip{,2,3}; do
+    pip_command="$(pyenv which "$pip" 2>/dev/null)"
+    [[ -n "$pip_command" ]] && break
+  done
+  unset pip
+else
   pip_command="$commands[(i)pip(|[23])]"
+fi
+if [[ -n "$pip_command" ]]; then
+  cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/pip-cache.zsh"
 
   if [[ "$pip_command" -nt "$cache_file" \
         || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
         || ! -s "$cache_file" ]]; then
     mkdir -p "$cache_file:h"
     # pip is slow; cache its output. And also support 'pip2', 'pip3' variants
-    $pip_command completion --zsh \
+    "$pip_command" completion --zsh \
       | sed -e "s/\(compctl -K [-_[:alnum:]]* pip\).*/\1{,2,3}{,.{0..9}}/" \
       >! "$cache_file" \
       2> /dev/null
@@ -166,8 +173,9 @@ if (( $#commands[(i)pip(|[23])] )); then
 
   source "$cache_file"
 
-  unset cache_file pip_command
+  unset cache_file
 fi
+unset pip_command
 
 # Load conda into the shell session, if requested
 zstyle -T ':prezto:module:python' conda-init


modules/python/init.zsh
```diff
@@ -147,12 +147,16 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
 fi
 
 # Load PIP completion.
-if (( $#commands[(i)pip(|[23])] )); then
+for pip_command in pip{,2,3}; do
+  if [[ "$PYENV_ROOT" ]]; then
+    [[ "$(pyenv which "$pip_command" 2>/dev/null)" ]] && break
+  else
+    [[ "$(which "$pip_command" 2>/dev/null)" ]] && break
```
jcassee, line 154: Using `$command` is nice, but using `which` gives it a nice symmetry.

srijanshetty: I like the symmetry as well, but would want some more comments on the change before accepting.

srijanshetty:
@belak: Could you review this change as well. Lgtm and is needed after python2 EOL

modules/python/init.zsh
```diff
@@ -147,12 +147,16 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
 fi
 
 # Load PIP completion.
-if (( $#commands[(i)pip(|[23])] )); then
+for pip_command in pip{,2,3}; do
+  if [[ "$PYENV_ROOT" ]]; then
```
belak, line 151: This should be using `-n` to check if it's of a non-zero length.

modules/python/init.zsh
```diff
@@ -147,12 +147,16 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) && \
...
+    [[ "$(pyenv which "$pip_command" 2>/dev/null)" ]] && break
+  else
+    [[ "$(which "$pip_command" 2>/dev/null)" ]] && break
+  fi
+done
+if [[ "$pip_command" ]]; then
```
belak, line 157: This should also be using `-n`

modules/python/init.zsh
belak: Also, there is a downside to using `which`. If any of these are overridden to be functions or aliases, the cache check below `if [[ "$pip_command" -nt "$cache_file" ...]]` will fail.

jcassee: > Also, there is a downside to using which. If any of these are overridden to be functions or aliases, the cache check below if [[ "$pip_command" -nt "$cache_file" ...]] will fail.

Interestingly, the result of calling `which` is never used. So whether or not it is an alias, `[[ "$pip_command" -nt "$cache_file" ...]]` _always_ fails! This is an even bigger bug.

jcassee:
@belak @srijanshetty I rebased to resolve the conflict introduced in #1842. (I would really like GitHub to notify me when that happens...) Could you give it another review?

belak:
Looks good to me. Thanks!

jcassee:
Thanks for the review and merge @belak  and @srijanshetty

PR #1850, opened 2020-06-06
Allow custom TMUX config path

pschmitt:
## Proposed Changes

  - Right now the TMUX plugin only makes use of the default TMUX config path (`~/.tmux.conf`)
  - This PR allows users to set the path to their tmux config.

Example:

```sh
zstyle ':prezto:module:tmux:config' path "$XDG_CONFIG_HOME/tmux.conf"
```

### DIFF
--- a/modules/tmux/README.md
+++ b/modules/tmux/README.md
@@ -53,6 +53,16 @@ zstyle ':prezto:module:tmux:iterm' integrate 'yes'
 
 Read [iTerm2 and tmux Integration][7] for more information.
 
+#### Custom Config File Path
+
+By default this plugin will use tmux's default config path (`$HOME/.tmux.conf`).
+If your config is stored elsewhere (in `XDG_CONFIG_HOME` for example), you can
+set the path using:
+
+```sh
+zstyle ':prezto:module:tmux:config' path "$XDG_CONFIG_HOME/tmux.conf"
+```
+
 Aliases
 -------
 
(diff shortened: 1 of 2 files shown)


srijanshetty:
Thanks for the change @pschmitt. Though I'm divided on the utility of the changeset. 

I use tmux and the default has location supported by tmux is `~/.tmux.conf`. The usual workflow does not involve changing the location of the conf file and one can always alias `tmux=tmux -f path` which should solve the problem.

pschmitt:
Yes an alias could work, I personally don't "trust" aliases all that much and prefer to avoid using them.

Having the possibility to use a zstyle instruction at the module level is cleaner in my opinion.
That's obviously debatable but here's how I see it: to have the tmux module behave if you don't use the default config path ([which will change btw in 3.2](https://github.com/tmux/tmux/blob/0d8ba2e57ff88582bf1ffeddd523e113f0138001/CHANGES#L133-L134)) you have to do 2 things:
1. setup an alias *before* the module get sourced
2. configure the module via zstyle

Just doing 2. is better imho.

srijanshetty:
> Try $XDG_CONFIG_HOME/tmux/tmux.conf as well as ~/.config/tmux/tmux.conf for
  configuration file (the search paths are in TMUX_CONF in Makefile.am).

Did not know about this. Thanks for pointing it out :smile:

> setup an alias before the module get sourced

The temporal coupling is just ugly, I agree.

> Just doing 2. is better imho.

No doubt, 2 is the better implementation.

I'm just wondering if this is changeset is needed at all, and even if we do make this change, always using `tmux -f $tmux_config` in the changeset is a ticking time-bomb and people will forget to use it. Probably makes sense to define a local variable which would be more defensive than this.

Just want more opinions on this changeset.

PR #1853, opened 2020-06-19
Change git alias `glg`

hidekuro:
I created this alias because the `--all` option cannot be overridden.
If possible, I'd like to change the existing `glg` (has `--all` option) to `glga` or `glG` and make this alias as `glg`.


belak:
I would be fine with reversing the aliases (make --all glG). If you want to make that change, I can merge it in with this.

srijanshetty:
@belak what's the policy with addition of new aliases? Do we look at them case by case?

My personal muscle memory doesn't really gel well with prezto aliases so I have my own set which is mostly confined to git based aliases and shortening some common commands I use.

belak:
@srijanshetty It's mostly case-by-case... I try to look at them logically since I generally don't use them either. In this case, it makes more sense to make the base alias without `--all` and have the modifier add it. We've done other aliases similar to this in the past.

hidekuro:
@belak I agree with you. It might be simpler to remove `--all` from `glg`.

hidekuro:
update issue subject.

## Issue Descriptions

remove `--all` flag to be able to control revision to displaying.

hidekuro:
Is there anything else I should do?
Should I withdraw my PR?

belak:
This looks good to me - sorry, I've just got a ton of real life stuff going on at the moment.

hidekuro:
@belak Thank you for your time!

PR #1857, opened 2020-07-19
Fix README.md

ekrecker:
I changed `instructed` to `instruct`


srijanshetty:
It should be `be instructed` instead of `instruct`. So, `in case of conflict you will be instructed to go into....`

ekrecker:
Okay.
I revised it again.

belak:
LGTM! Thanks!

PR #1858, opened 2020-07-19
Add ^_ mapping to undo in viins mode

ardeidae:
This one line addition allows us to cancel a tab completion in viins as ^_ in emacs mode.
That's all.


belak:
Does this line up with what vim uses? If it does, it seems like a worthwhile change.

PR #1860, opened 2020-07-20
Add ^_ mapping to undo in viins mode

ardeidae:
My initial PR wasn't good. So I closed it. So, this is the good version.

This one line addition allows to cancel a tab completion in viins as ^_ in emacs mode.
That's all.

This mapping is essential for a vi mapping. Thank you for considering it.

### DIFF
--- a/modules/editor/init.zsh
+++ b/modules/editor/init.zsh
@@ -318,6 +318,7 @@ bindkey -M vicmd "$key_info[Control]X$key_info[Control]E" edit-command-line
 
 # Undo/Redo
 bindkey -M vicmd "u" undo
+bindkey -M viins "$key_info[Control]_" undo
 bindkey -M vicmd "$key_info[Control]R" redo
 
 if (( $+widgets[history-incremental-pattern-search-backward] )); then


belak:
Whoops, I commented on the old PR, but I'll repeat here:

Is this the keybind vim uses for this? If yes, it seems like a worthwhile change.

ardeidae:
As far as I know, there is no undo mapping in vim insert mode. I just checked the vim documentation, and I didn't find anything.
There is `i_ctrl-w` to undo the last typed word, but it does not undo a variable expansion, for example.

So, I guess taking the same mapping as emacs is a good choice.

ardeidae:
Hi,
Do you have some news about this pull request?
Matthieu

belak:
I finally figured out what was bugging me! It seems like emacs has multiple keys bound to `undo`. I was used to using `C-/`, but the bind you mentioned `C-_` is also used, as well as `C-x u` and a few others.

As far as I'm aware `C-/` is more standard, but I'd accept this if it added both binds.

ardeidae:
OK, I will update this request this evening.

Le jeu. 30 juil. 2020 à 20:24, Kaleb Elwert <notifications@github.com> a
écrit :

ardeidae:
The `C-/` mapping seems to be already defined in vins mode.
So I think you should accept this PR as it is to add `C-_`.
Having the two mappings maybe a good alternative for different keyboard layouts.
I didn't know `C-/`, but on french keyboard, it could be better, but for others layouts, maybe the other will be preferred.

belak:
I don't think we bind it in prezto - maybe that's by zsh itself. I guess this is fine for now. Thanks for submitting this!

ardeidae:
Thanks you :-)

PR #1861, opened 2020-07-27
Add check for empty WORKON_HOME

oohlaf:
## Proposed Changes
When $WORKON_HOME is not set and $ENV_NAME is equal to
$PROJECT_ROOT/.venv which contains bin/activate then
the code wrongly assumes workon is available.

It should execute the second branch to source bin/activate.


belak:
Good catch! LGTM!

PR #1879, opened 2021-01-08
cask aliases update

### LINKED ISSUE #1866: caskl return error 1
### Description
`caskl` tries to run new way to list casks, but an error code is returned.

### Expected behavior
Listing of installed casks

### Actual behavior
```
Error: Calling brew cask list is deprecated! Use brew list --cask instead.
-> [1]
```

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

### Versions

  - Prezto commit: 0a405494cae54e208ff550c1ba83380ed115410a
  - ZSH version: zsh 5.7.1
  - OS information: macOS 10.15.6

Rawmanski:

Fixes #1866 

## Proposed Changes

  - deprecated cask aliases
  - 
  -

### DIFF
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -39,14 +39,11 @@ alias brewu='brew upgrade'
 alias brewx='brew uninstall'
 
 # Homebrew Cask
-alias cask='brew cask'
-alias caskc='hb_deprecated brew cask cleanup'
-alias caskC='hb_deprecated brew cask cleanup'
-alias caski='brew cask install'
-alias caskl='brew cask list'
-alias casko='brew cask outdated'
-alias casks='hb_deprecated brew cask search'
-alias caskx='brew cask uninstall'
+alias caski='brew install --cask'
+alias caskl='brew list --cask'
+alias casko='brew outdated --outdate'
+alias casks='brew search --cask'
+alias caskx='brew uninstall --cask'
 
 function hb_deprecated {
   local cmd="${@[3]}"
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -41,7 +41,7 @@ alias brewx='brew uninstall'
 # Homebrew Cask
 alias caski='brew install --cask'
 alias caskl='brew list --cask'
-alias casko='brew outdated --outdate'
+alias casko='brew outdated --cask'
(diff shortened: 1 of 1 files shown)


Rawmanski:
Updated cask aliases, because old ones are deprecated. Hence `brew cask install` isn't working anymore, instead use `brew install --cask`.

modules/homebrew/init.zsh
```diff
@@ -39,14 +39,11 @@ alias brewu='brew upgrade'
...
-alias casko='brew cask outdated'
-alias casks='hb_deprecated brew cask search'
-alias caskx='brew cask uninstall'
+alias caski='brew install --cask'
+alias caskl='brew list --cask'
+alias casko='brew outdated --cask'
+alias casks='brew search --cask'
+alias caskx='brew uninstall --cask'
 
 function hb_deprecated {
```
jeffwidman, line 48: Shouldn't this also be removed since it's no longer used??

indrajitr: Well, kinda :) Actually, we shouldn't be abruptly dropping alias without a migration warning. This is in my backlog. Will cherry-pick the commit and apply the necessary changes.

jeffwidman: While I am not opposed to a migration warning, I am also _completely_ fine with dropping aliases without a migration warning. I _really_ don't want the expectation to be that we have to do a formal migration for everything, especially more minor things like simple aliases.

So far we've kept the focus that this is a low-maintenance thin shell wrapper targeting power users, so I think it's fine to have an assumption of "read the commit log when you upgrade" so as long as we clearly document backwards breaking changes it in the commit title, then I think that's sufficient. We aren't that active of a project, so as long as you update once or twice a year it should take < 10 mins to skim the commit log.

Again, I am totally not opposed if you want to go to the additional work of adding migration warnings, as long as it doesn't become an expectation/requirement except for core functionality changes (and even then, I'm often still fine with "the migration message is the commit history").

indrajitr: Okay @jeffwidman, compelling arguments! 👍🏽

indrajitr:
Created #1905 cherry-picking commits from @Rawmanski.

indrajitr:
Closed via #1905

Issue #1880, opened 2021-01-08
compinit not always getting the benefit of `-C` option


mattmc3:
### Description
I was tuning my ZSH startup performance and noticed that compaudit kept coming up when I run zprof. That seemed strange, since Prezto should only be running `compint -C` (skip function check) most of the time, and just running `compinit` once daily. I believe this is due to my completions not changing, and thus `compinit` runs, but never changes the modified time on zcompdump, so it just runs again and again, never getting the benefit of `compinit -C`

## Proposed Changes

The fix seems simple if I'm not mistaken on what happens. `touch zcompdump` after running `compinit -i`.

### Expected behavior

If I have a 2 day old zcompdump, I expect `compinit` to do a full run, but then get the benefit of `compinit -C` for the rest of the day.

### Actual behavior

The full `compint` runs because my completions never change, and my zcompdump never gets modified, so my compinit always takes the long way around. 

### Steps to Reproduce

```zsh
zcompdump="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/zcompdump"
touch -a -m -t 01061701.01 $zcompdump
```

Then, `zmodload zsh/zprof` at the top and `zprof` at the bottom of your .zshrc and open new ZSH sessions. `compaudit` shows every time, and the file dates never update.

### Versions

  - Prezto commit: latest
  - ZSH version: zsh 5.8 (arm-apple-darwin20.2.0)
  - OS information: macOS Big Sur 11.1

mattmc3:
Resolved via this PR: https://github.com/sorin-ionescu/prezto/pull/1903

PR #1881, opened 2021-01-08
Ensure compinit updates its modified date

mattmc3:
`compinit -i` doesn't always change zcompdump, which means that the `compinit -C` condition fails to run until the completions change. When we `compinit -i`, we should touch the file too to solve this.

See: https://github.com/sorin-ionescu/prezto/issues/1880

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -40,6 +40,7 @@ if [[ $_comp_path(#qNmh-20) ]]; then
 else
   mkdir -p "$_comp_path:h"
   compinit -i -d "$_comp_path"
+  touch "$_comp_path"
 fi
 unset _comp_path
 


modules/completion/init.zsh
```diff
@@ -40,6 +40,7 @@ if [[ $_comp_path(#qNmh-20) ]]; then
 else
   mkdir -p "$_comp_path:h"
   compinit -i -d "$_comp_path"
+  touch "$_comp_path"
```
jeffwidman, line 43: Could you add a clarifying comment in the code  of _why_ we need this? Nothing fancy, but it's helpful when you see things that seem useless in teh code to have a quick comment explaining why it's not useless...

jeffwidman (approved):
This sounds reasonable to me, but I haven't done a deep dive on it. 

Any thoughts @belak @indrajitr @johnpneumann ?

indrajitr:
Good one @mattmc3. Thank you!
Cherry-picked your commit and applied with additional comment (for our future self) in #1903 :)

PR #1885, opened 2021-02-14
Update history-substring-search submodule to v1.0.2

ytang:
The reference to `history-substring-search` is over four years old. Let's update it to the latest version.

## Proposed Changes

  - Update submodule `history-substring-search` to v1.0.2


jeffwidman (approved):
Thx!

PR #1888, opened 2021-03-01
osx module mand fix

Herobs:
> The {query} has to be percent-escaped in macOS 11.

Just replace `:` to `%3A` work fine.

But add function like this one I find on stackoverflow to encode the entire query string would be more robust.
```bash
urldecode() {
    # urldecode <string>

    local url_encoded="${1//+/ }"
    printf '%b' "${url_encoded//%/\\x}"
}
```


indrajitr:
Merged. Thanks @Herobs!

PR #1895, opened 2021-04-26
prompt: Remove Powerlevel9k

indrajitr:
## Remove Powerlevel9k

Powerlevel9k has been marked deprecated and unmaintained by the author.
Powerlevel10k is now the recommended replacement.

See: https://github.com/Powerlevel9k/powerlevel9k#powerlevel9k-is-deprecated-and-now-unmaintained-please-use-powerlevel10k

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -25,10 +25,6 @@
 [submodule "modules/prompt/external/async"]
 	path = modules/prompt/external/async
 	url = https://github.com/mafredri/zsh-async.git
-[submodule "modules/prompt/external/powerlevel9k"]
-	path = modules/prompt/external/powerlevel9k
-	url = https://github.com/bhilburn/powerlevel9k.git
-	shallow = true
 [submodule "modules/prompt/external/powerlevel10k"]
 	path = modules/prompt/external/powerlevel10k
 	url = https://github.com/romkatv/powerlevel10k.git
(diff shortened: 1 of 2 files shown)


belak:
I seem to remember that p10k is backwards compatible - would it work to point the p9k setup to p10k?

CC @romkatv

romkatv:
> I seem to remember that p10k is backwards compatible

Mostly. See https://github.com/romkatv/powerlevel10k#does-powerlevel10k-always-render-exactly-the-same-prompt-as-powerlevel9k-given-the-same-config.

> would it work to point the p9k setup to p10k so we don't break users of p9k when they upgrade?

This may cause some grief to the existing prezto users who are still on powerlevel9k and whose config isn't 100% compatible with powerlevel10k. If they are fine with the prompt they have, I'd say leave them be. It's not like maintaining the powerlevel9k submodule requires a lot of ongoing work for prezto folks.

indrajitr:
This is the kind of discussion I was hoping to have :)

So we are looking are three options:

* Keep everything _as-is_. Safe in the near term, but doesn't help in the long run. Granted this isn't much work though.
* Apply under-the-hood linking from p9k to p10k. Will mostly work, but adds an element of surprise when it doesn't.
* Take a one-time hit and avoid bloat in prezto with (upstream) unsupported code and communicate to the prezto user what exactly changed.

This PR attempts the last one. This does add a one-time annoyance, but a reasonably easy one to mitigate. What say?

NB: I missed out on committing change in `prompt/init.sh` which alerts the user about the change. It has now been updated.

belak:
This seems reasonable. Good idea to start looking to drop support.

jeffwidman:
I'm +1 on just killing it. Prezto has historically taken a fairly low-level / blunt approach to migrations/deprecations. IE, your existing won't break, but when you update, you're expected to at least skim the commit log to see what the changes are.

In general we try to notify and _not_ be the first folks to drop something, but given that powerlevel10k has been around a while I think we can say we've given plenty of time. This one is a pretty simple migration, so I'm personally in favor of just dropping the bloat and moving forward.

PR #1898, opened 2021-04-29
Allow loading more specific 'run-help' function

indrajitr:
Allow loading more specific 'run-help' function from $fpath. This allows automatically looking up specific sub-command helper if available instead of the static default ('man').

See: 
- https://github.com/zsh-users/zsh/blob/ccc9cff9e244725ed604fd1ac20e4958339e3885/Functions/Misc/run-help#L3-L8
- https://stackoverflow.com/a/32293317


jeffwidman (approved):
Nifty!

belak (approved):

PR #1900, opened 2021-04-30
command-not-found: Update docs for Homebrew

indrajitr:
Update documentation for Homebrew to indicate that a recent version of 
Homebrew does not need to tap `command-not-found` Homebrew repository.

See: Homebrew/brew#11137

### DIFF
--- a/modules/command-not-found/README.md
+++ b/modules/command-not-found/README.md
@@ -7,8 +7,9 @@ the proper install command.
 
 Debian and Arch Linux based distributions use the [`command-not-found`][1] tool.
 
-macOS uses Homebrew's [`command-not-found` clone][2]. Note that you also need to
-[follow the instructions][3] to tap the `command-not-found` homebrew repository.
+macOS uses Homebrew's [`command-not-found` clone][2]. Note that unless you have
+a recent version of Homebrew installed, you might also need to tap the
+`command-not-found` Homebrew repository [following the instructions][3].
 
 
 Authors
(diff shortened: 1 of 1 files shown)


jeffwidman (commented):
I'd actually omit any mention of how to install... this code is unlikely to get touched soon, and anyone looking at this within a few months will probably be running a newer version of homebrew as it auto-updates itself IIRC. Anyone who hits problems will almost certainly go look up the repo and find the install instructions, but I suspect the number of folks impacted will be very small.

Don't forget to also remove the footnote `[3]` if you remove this.

indrajitr:
> I'd actually omit any mention of how to install... this code is unlikely to get touched soon, and anyone looking at this within a few months will probably be running a newer version of homebrew as it auto-updates itself IIRC. Anyone who hits problems will almost certainly go look up the repo and find the install instructions, but I suspect the number of folks impacted will be very small.

Yes, I considered removing the how-to install part altogether. But then thought of keeping it around for a short while for that small subset of people who updated prezto (user environment) but didn't update homebrew (system environment) and then clean it up.

> Don't forget to also remove the footnote [3] if you remove this.
Indeed :)

jeffwidman (approved):
looks fine. I'm still +1 on simply removing it but also fine if you want to do it in two steps over a longer migration period.

jeffwidman:
Sorry I wasnt' clear... you should merge this once you're comfortable with it!

indrajitr:
> looks fine. I'm still +1 on simply removing it but also fine if you want to do it in two steps over a longer migration period.

Right, the only reason I wanted to keep it around for a while is because this specific Homebrew change is quite recent (about 2 weeks). And for folks with _managed_ macOS, updates to `/usr/local/homebrew` (or `/opt/homebrew`) might depend on external factors (MDM, `sudo` access etc.).

PR #1901, opened 2021-04-30
general: Remove redundant usage of `cat`

indrajitr:
In general, the substitution ‘$(cat foo)’ may be replaced by the
equivalent but faster ‘$(<foo)’.

### DIFF
--- a/modules/archive/functions/archive
+++ b/modules/archive/functions/archive
@@ -11,7 +11,7 @@
 local archive_name path_to_archive _gzip_bin _bzip2_bin _xz_bin
 
 if (( $# < 2 )); then
-  cat >&2 <<EOF
+  >&2 <<EOF
 usage: $0 [archive_name.zip] [/path/to/include/into/archive ...]
 
 Where 'archive.zip' uses any of the following extensions:
--- a/modules/archive/functions/lsarchive
+++ b/modules/archive/functions/lsarchive
@@ -10,7 +10,7 @@
 local verbose
 
 if (( $# == 0 )); then
-  cat >&2 <<EOF
+  >&2 <<EOF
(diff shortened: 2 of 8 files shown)


jeffwidman (commented):
Code looks fine if `cat` is truly redundant here, but I'm too ignorant here to feel comfortable approving... would like the other maintainers to take a look.

indrajitr:
@belak any view?

belak:
I held off on this one before because I wanted to test it myself - it might be a little while until I can do that

belak:
This is an interesting PR, to be sure - I didn't know this was possible. It mostly makes sense, but it's not the direction my brain would go for these.

I like the change in modules/python/init.zsh - that's a fairly common convention when reading from files which I've seen in multiple places before.

I have no strong opinions on the heredocs - they work, as far as I can tell, they're solid, and will probably perform marginally better. However, if it was purely my decision, I probably wouldn't push for that replacement - it's a bit less readable and doesn't follow with common conventions, especially as seen in other shells. That hasn't stopped us before, at least when there's a zsh convenience, but for me in this specific case it wouldn't be worth it.

I'm a fan of the replacement in modules/python/init.zsh, and if you feel strongly about the heredoc change or think it adds enough value, you're definitely welcome to merge it as-is.

indrajitr:
Yes, readability does take a hit, I admit. But Zsh "hieroglyphs" (will all esoteric single-character flags and switches) probably don't focus on readability anyway I guess :) 

One probable reason we don't see this convention around is probably because of POSIX compliance. Besides, pure Zsh optimized interactive shell is a niche space.

I'll hold off merging this just yet. Let's see if this PR can evolve a bit.

PR #1906, opened 2021-05-02
history: Move HISTFILE path to more conventional one

### LINKED ISSUE #1766: HISTFILE defaults to .zhistory rather than .zsh_history
### Description

When migrating from oh-my-zsh I lost my history and didn't initially understand why.

oh-my-zsh uses `.zsh_history` for history, see [lib/history.zsh](https://github.com/ohmyzsh/ohmyzsh/blob/master/lib/history.zsh#L29)

### Expected behavior

By default, without configuration, I should be able to continue using my existing history file `.zsh_history`.

### Actual behavior

The prezto history modules defaults `HISTFILE` to `.zhistory`

### Versions

  - Prezto commit: 2d6205e71404704eecc4f402e5b09d7cbd19bab7
  - ZSH version: zsh 5.7.1 (x86_64-apple-darwin18.2.0)
  - OS information: macOS 10.14.6

indrajitr:
Change default `HISTFILE` path from `~/.zhistory` to more conventional `~/.zsh_history`. This aligns better with Debian variants, macOS and Oh My Zsh.
 
This would provide an easier onboarding experience to users coming to Prezto from standard (framework free) zsh or from Oh My Zsh.

For existing Prezto users, we attempt to automatically rename old `HISTFILE` to new one iff the old one exists and the new one doesn't. However, if both old and new `HISTFILE` exist and the old one is more recent, we just alert the user about what changed with a suggested way to accommodate the change.

Closes #1766.

### DIFF
--- a/modules/history/README.md
+++ b/modules/history/README.md
@@ -3,10 +3,13 @@ History
 
 Sets [history][1] options and defines history aliases.
 
-**Note:** If you are migrating from [Oh My Zsh][2] and are not on a recent
-version of macOS and want to keep your history, you will either need to set
-`HISTFILE` manually to `$HOME/.zsh_history` or rename `~/.zsh_history` to
-`~/.zhistory`.
+**Note:** Default path of `HISTFILE` has changed from `${ZDOTDIR:-$HOME}/.zhistory`
+to `${ZDOTDIR:-$HOME}/.zsh_history`. The file will be automatically renamed if
+possible (when the new one doesn't exist). Otherwise, if you want to preserve
+previous history, you will need to move them from `${ZDOTDIR:-$HOME}/.zhistory`
+to `${ZDOTDIR:-$HOME}/.zsh_history`.
+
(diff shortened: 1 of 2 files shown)


jeffwidman (approved):
The auto-rename is a very clean way of handling. Has the potential to break other scripts etc, but I suspect not too many people script access to their shell history. And I'd (slightly) prefer to auto-rename than not... if someone really doesn't want it, they can rename it right back, it's not that hard.

Before merging, can you wait a day or two for @belak to also take a look in case he has any input?

indrajitr:
> The auto-rename is a very clean way of handling. Has the potential to break other scripts etc, but I suspect not too many people script access to their shell history. And I'd (slightly) prefer to auto-rename than not... if someone really doesn't want it, they can rename it right back, it's not that hard.

As you have noticed, auto-renaming is very conservative. It does so under exactly one condition - when the old HISTFILE exists and the new one doesn't exist. But point taken.

> Before merging, can you wait a day or two for @belak to also take a look in case he has any input?

Of course.

belak (approved):
Looks good to me at a quick glance. Thanks for doing this!

PR #1908, opened 2021-05-02
completion: Add completion for keg-only brewed curl when available

indrajitr:
As curl bundled with macOS does not include `_curl` definitions, one way to make `_curl` available is to install curl from Homebrew.

However, since curl from Hombrew is _keg-only_, its function path needs to be explicitly included.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -14,6 +14,13 @@ fi
 # Add zsh-completions to $fpath.
 fpath=("${0:h}/external/src" $fpath)
 
+# Add completion for keg-only brewed curl when available.
+if (( $+commands[brew] && ! $+functions[_curl] )) && \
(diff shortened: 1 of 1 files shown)


belak:
Cool workaround. My only worry is that any call to brew is fairly expensive.

indrajitr:
`brew --prefix` is [just shell](https://github.com/Homebrew/brew/blob/eae355810873ceb6972f14f7f46c508a7e0309d5/Library/Homebrew/brew.sh#L66), and [intentionally so](https://github.com/Homebrew/brew/blob/eae355810873ceb6972f14f7f46c508a7e0309d5/Library/Homebrew/brew.sh#L2-L3). We should be fine.

belak:
Nice, glad to see that's been improved since I looked at this last (admittedly quite a while ago)

PR #1910, opened 2021-05-03
node: Fixup nvm path detection for Homebrew based nvm

indrajitr:
Prefer using `brew --prefix nvm` instead of just `brew --prefix` and actually use the variable `nvm_prefix` once detected and set up.


jeffwidman (approved):
This looks reasonable to me, and I trust you know what you're doing... I'm not really a `node` guy so didn't play with it myself to verify. But I'm comfortable with merging and if there are unexpected problems, I'm sure someone will submit a PR. 😀 
🚢

PR #1912, opened 2021-05-03
rsync: Check for new rsync before applying newer options

indrajitr:
The newer options for extended attributes or file-flags got reliable only after rsync v3.1.
Also, recommend Homebrew or MacPorts rsync for macOS.


jeffwidman (approved):

jeffwidman:
I don't have full context here on all the history of `rsync`, but these changes look reasonable to me.

PR #1915, opened 2021-05-05
prompt: Fix git submodule path for 'pure' with actual path

indrajitr:
In addition to fixing git submodule path for 'pure' with the actual path, we also apply submodules' path change in doc and `zprezto-update` function.
 
Accordingly, updated instruction and `zprezto-update` function to synchronizes submodules' remote URL configuration setting to the updated value automatically.

---
**WARNING:** This will require synchronizing submodules' remote URL configuration setting to the value specified in `.gitmodules` by doing: `git submodule sync --recursive` in your Prezto location.

---
Supersedes and resolves #1827

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,7 +16,7 @@
 [submodule "modules/prompt/external/agnoster"]
 	path = modules/prompt/external/agnoster
 	url = https://github.com/agnoster/agnoster-zsh-theme.git
-[submodule "modules/prompt/functions/pure"]
+[submodule "modules/prompt/external/pure"]
 	path = modules/prompt/external/pure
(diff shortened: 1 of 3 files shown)


jeffwidman (approved):

jeffwidman:
Let's have @belak take a look too in case there's something we both overlooked

belak:
Thanks for all the work you've been putting in!

I believe this makes sense, feel free to merge it.

Just a note - I don't have a ton of bandwidth to review this week because I'm moving this Saturday, but I'll try to poke my head in a little bit. Don't feel like you need to wait for me on everything, especially doc changes. :)

PR #1917, opened 2021-05-06
utility: Detect GNU 'ls' before applying the relevant tweaks

indrajitr:
Detect `ls` provided by GNU Core Utilities upfront before applying the relevant settings and aliases.

The revised logic takes the following aspects into consideration:
- `is-callable 'dircolors'` is not a sufficient test for verifying if the `ls` in scope is _also_ provided by GNU Core Utilities. `ls` can continue to be the one provided by BSD Core Utilities because of its precedence in `$path`.
- The `ls` in scope anyway can be either GNU Utils provided or BSD Utils provided depending on either `$path` precedence, soft link (via `ln -s`), or wrapper function (via 'gnu-utility' module).
- So instead of '_guessing_' the possible one, just detect if it is from GNU Utils and apply the relevant settings and aliases.
- Note that GNU prefixed `ls` (`gls`) is intentionally not honored in this case to avoid possible conflict/shadowing with other `gls` callable (can happen if 'git' module is used, for example). Besides, honoring `gls` would imply we honor other commands (like `mv`, `rm` etc.) to be consistent. We have a module *`gnu-utils`* dedicated for that after all.

Also closes #1752.

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -2,6 +2,11 @@
 
 Defines general aliases and functions.
 
+**Note:** Some of the utilities configured in this module might be provided via
+GNU utilities with incompatible arguments on non-GNU systems. In such cases,
+using [*`gnu-utility`*][1] module is recommended and it must be loaded
+**before** this module.
+
 ## Settings
 
 ### Highlighting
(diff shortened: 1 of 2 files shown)


indrajitr:
Note:
- This is an alternative version of #1786.
- On approval, this closes #1786 and also closes #1752.

PR #1919, opened 2021-05-07
completion: Update zsh-completions submodule to 0.33.0

indrajitr:
zsh-completions 0.33.0 released

See: https://github.com/zsh-users/zsh-completions/releases/tag/0.33.0


belak:
:shipit: 

Unless there's a reason to be more careful, I often do a quick audit of the changes and just push updates to external repos - a PR probably isn't always necessary.

belak (approved):

PR #1923, opened 2021-05-10
utility: Miscellaneous fixes and tweaks

indrajitr:
## Proposed Changes

Miscellaneous fixes and tweaks in utility module:

- `wdiff`: Use `git` fallback (if present) even when color is off and `wdiff` is not present.
- `dut`: Skip additional call to `grep` and use Zsh native mechanism to detect GNU version of `du`.
- `make`: Simplify conditional block.
- general: Remove redundant `function` clause as per Prezto convention.

Note: While this is a single PR for multiple changes (for convenience of approvers), the changes have been split into separate commits (for convenience of `git bisect`-ing and `git revert`-ing in future if necessary).

### DIFF
--- a/modules/utility/functions/diff
+++ b/modules/utility/functions/diff
@@ -5,13 +5,13 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
-function diff {
-  if zstyle -t ':prezto:module:utility:diff' color \
-    && (( $+commands[colordiff] )); then
-      command colordiff "$@"
-  else
-    command diff "$@"
-  fi
-}
+# function diff {
 
-diff "$@"
+if zstyle -t ':prezto:module:utility:diff' color \
+      && (( $+commands[colordiff] )); then
+  command colordiff "$@"
+else
+  command diff "$@"
+fi
+
+# }
--- a/modules/utility/functions/dut
+++ b/modules/utility/functions/dut
@@ -6,22 +6,22 @@
 #   Sorin Ionescu <sorin.ionescu@gmail.com>
 #
 
-function dut {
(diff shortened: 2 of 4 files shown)


modules/utility/functions/diff
```diff
@@ -5,13 +5,13 @@
...
 
-function diff {
-  if zstyle -t ':prezto:module:utility:diff' color \
-    && (( $+commands[colordiff] )); then
-      command colordiff "$@"
-  else
-    command diff "$@"
-  fi
-}
+# function diff {
```
blanorama, line 8: friendly asked: why committing commented code? or do I oversee something?

indrajitr: Good question @blanorama! 
In this case, the `function diff {}` wrapper is unnecessary as you recognize. It was added here for consistency with the convention that we follow in Prezto. Please have a look at the [original PR](https://github.com/sorin-ionescu/prezto/pull/1366) by @belak regarding the motivation behind this.

blanorama: ahh, okay, so in general it’s there for the autoload mechanism to work, which is using grep among others; and in this particular case it would not be needed but for consistency we stick to this…
Do I get it right?

indrajitr: Actually, it is there just for consistency (documentation, easy to grep, etc.) only. The line is just a comment, it has no other significance for Zsh runtime :)

Autoloading will work regardless of the line `# function diff {` because the file `diff` exists in a directory (`modules/utility/functions`) listed for Zsh to scans for function (specified via `fpath`).

blanorama: Got it, thanks for clarification. 👍

PR #1929, opened 2021-05-17
update gSu

hidekuro:
## Proposed Changes

Remove `origin master` from gSu.
It safe for branches not named "master".

  - Avoid the error "fatal: couldn't find remote ref master" in a repository which the default branch is not "master".
  - And with this method, we can add any remote and branch name after it such as "origin master" or "upstream main".

### DIFF
--- a/modules/git/alias.zsh
+++ b/modules/git/alias.zsh
@@ -253,7 +253,7 @@ if ! zstyle -t ':prezto:module:git:alias' skip 'yes'; then
   alias gSl='git submodule status'
   alias gSm='git-submodule-move'
   alias gSs='git submodule sync'
-  alias gSu='git submodule foreach git pull'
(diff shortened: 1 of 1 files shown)


indrajitr:
Does it help if you do: `git submodule update --recursive --remote` in your case?
Admittedly, it might not work if the submodule branch isn't set properly though.

hidekuro:
@indrajitr It seems better. It works fine in my case. :)
I will change to the method you suggested.

indrajitr:
Thank you @hidekuro! Your changes have been cherry-picked ([with additional comments](https://github.com/sorin-ionescu/prezto/pull/1930/commits/a7033d9deabe236cf80ee477e1d7fe4cae38935a)) in #1930.

hidekuro:
The proposal has been cherry-picked into #1930, so I close this PR.

PR #1931, opened 2021-05-21
python: Optimize pip completion for lazy loading and update pyenv initialization

indrajitr:
## Fixes:
Move `pip` to separate completion definition delegating to pip to do all the completion work.

Since the completion is loaded lazily on-demand, it avoids the performance overhead during Zsh initialization.

_Implementation note:_
The helper function `_pip_completion` implementation is based on the official pip completion function (which can be generated with `pip completion --zsh`) adhering to the newer compsys style (see `man zshcompsys` for more details).

Update pyenv initialization flow per pyenv 2.0 recommendation.

Also, prefer looking up `$functions` for fully initialized `pyenv`.

Closes #1670

### DIFF
--- a/modules/python/README.md
+++ b/modules/python/README.md
@@ -2,6 +2,9 @@
 
 Enables local Python and local Python package installation.
 
+This module must be loaded _before_ the _`completion`_ module so that the
+provided completion definitions are loaded.
+
 ## Settings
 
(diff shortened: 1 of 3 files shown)


jeffwidman:
Great work @indrajitr !! 2* 👍

PR #1935, opened 2021-05-24
Fix broken source commands in node module

ehrenkret:
Placing `--no-use` inside the quoted string results in source
attempting to load a file which literally ends with ` --no-use`
instead of passing --no-use as an argument.


indrajitr:
Good catch, thank you @ehrenkret!

PR #1938, opened 2021-05-27
Update init.zsh

FirasSadiyah:
With the pyenv 2.0.0 upgrade, `pyenv init` -no longer sets PATH, therefore we need to use `eval "$(pyenv init --path)"` instead of `eval "$(pyenv init -)"`
Tested on macOS Big Sur 11.4 using zsh 5.8 and pyenv 2.0.0 installed via HomeBrew.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -13,11 +13,11 @@ pmodload 'helper'
 # Load manually installed pyenv into the path
 if [[ -s "${PYENV_ROOT:=$HOME/.pyenv}/bin/pyenv" ]]; then
   path=("${PYENV_ROOT}/bin" $path)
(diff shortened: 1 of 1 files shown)


indrajitr:
Thank you @FirasSadiyah for the PR! ✨ 👍🏽 

Wondering if we need just `eval "$(pyenv init --path)"` instead `eval "$(pyenv init -)"` or if we need _both_ `eval "$(pyenv init --path)"` and `eval "$(pyenv init -)"`.

Meanwhile, there is a separate PR (#1931) with this fix and some other ones. Would be great if you can please give it a try.

indrajitr:
Closed via #1931.

PR #1939, opened 2021-06-01
Add IntelliJ to the list of exceptions

mpeterson:
### Proposed Changes

Jetbrains IntelliJ adds an env variable `__CFBundleIdentifier` with the value `com.jetbrains.intellij` indicating that is the terminal triggered from here.

This patch adds this as an exception not to start tmux

### DIFF
--- a/modules/tmux/init.zsh
+++ b/modules/tmux/init.zsh
@@ -23,7 +23,7 @@ if ([[ "$TERM_PROGRAM" = 'iTerm.app' ]] && \
   _tmux_iterm_integration='-CC'
 fi
 
-if [[ -z "$TMUX" && -z "$EMACS" && -z "$VIM" && -z "$INSIDE_EMACS" && "$TERM_PROGRAM" != "vscode" ]] && ( \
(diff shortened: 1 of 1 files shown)


indrajitr:
LGTM for macOS, does IntelliJ set the same variable on Windows and Linux?

tblaschke:
Instead of using the bundleIdentifier, we could use the $TERMINAL_EMULATOR variable. See #1958

mahinshaw:
It might be better to use `"$__CFBundleIdentifier" != com.jetbrains.*`. This allows for all the tools under the jetbrains domain to load without starting tmux.

catflyflyfly:
> Instead of using the bundleIdentifier, we could use the $TERMINAL_EMULATOR variable. See #1958

Come here to say that your solution works. Thank you! I'll add more details about my finding on your issue.

jeffwidman:
Closing as https://github.com/sorin-ionescu/prezto/pull/1958 was merged which should address this.

There's some suggested improvements in the comments ☝️ , please feel free to open a PR implementing those as we're open to improvements.

Issue #1949, opened 2021-08-03
Virtualenvwrapper stopped working after d686d


jeffwidman:
Before d686da3c039a1ed0ae271860d271ec966f59c16a, virtualenvwrapper shell scripts were sourced properly, the `workon` etc commands were on my path, etc.

After that commit, the shell script doesn't get loaded for me at all. 

I'm still debugging what specifically changed, but clearly it is something in that commit.

belak:
I don't have prezto installed at the moment, but I'm fairly sure it's that a workaround I added was removed: https://github.com/sorin-ionescu/prezto/commit/d686da3c039a1ed0ae271860d271ec966f59c16a#diff-414db2cc4cb0c69d0fea2d6304ceb9921bf9d0d96f8d47f545e19bfa217dc39dL120-R119

belak:
Are you seeing the same problem as in https://github.com/sorin-ionescu/prezto/issues/1416, or is it a separate issue?

jeffwidman:
It's this line: https://github.com/sorin-ionescu/prezto/commit/d686da3c039a1ed0ae271860d271ec966f59c16a#diff-414db2cc4cb0c69d0fea2d6304ceb9921bf9d0d96f8d47f545e19bfa217dc39dR37

Adding basic `echo` statements and the new version doesn't catch python so it returns early. Without that `return`, it works as expected.

The change is a little cryptic to me, so trying to figure it out.

jeffwidman:
> Are you seeing the same problem as in #1416, or is it a separate issue?
> If it is that issue, this patch, should re-add that hack...

 I suspected the same bit of code at first, but it's a separate issue.

jeffwidman:
Non-working: 
* `$#commands[(i)python[23]#]`
* `$commands[(i)python[23]]`
Working: 
* `$+commands[(i)python[23]#]`
* `$#commands[(i)python[23]]`
* `$+commands[(i)python[23]]`

So it's related to the `#`. I'm not sure what they do though... checking the docs, but if anyone knows off top of their head, feel free to chime in.

jeffwidman:
I did some reading, and still not totally clear, but that may be because it's an hour past normal bed time. I'll re-read in the morning. @indrajitr if you happy to know, I'd appreciate any pointers.

[...]

PR #1951, opened 2021-08-03
Patch python detection to unbreak virtualenvwrapper

jeffwidman:
I discovered that my virtualenvwrapper scripts were no longer getting
loaded. After `git bisect` + further debugging, I traced it down to the
first change in this commit.

I also noticed the second change had similar formatting, so I modified
it as well, although I'm not sure that's it's broken.

Note: I researched quite a bit what the `#` is doing here, and I still
don't understand it. All I know is that it worked previously, then it
broke, and now it worked again. So putting up the commit/PR for
discussion.

Fix https://github.com/sorin-ionescu/prezto/issues/1949

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -44,7 +44,7 @@ fi
 unset local_pyenv
 
 # Return if requirements are not found.
-if (( ! $#commands[(i)python[23]#] && ! $+functions[pyenv] && ! $+commands[conda] )); then
+if (( ! $+commands[(i)python[23]] && ! $+functions[pyenv] && ! $+commands[conda] )); then
   return 1
 fi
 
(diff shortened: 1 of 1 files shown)


modules/python/init.zsh
```diff
@@ -44,7 +44,7 @@ fi
 unset local_pyenv
 
 # Return if requirements are not found.
-if (( ! $#commands[(i)python[23]#] && ! $+functions[pyenv] && ! $+commands[conda] )); then
+if (( ! $+commands[(i)python[23]] && ! $+functions[pyenv] && ! $+commands[conda] )); then
```
jeffwidman, line 47: This is the broken line... there are several ways to fix: https://github.com/sorin-ionescu/prezto/issues/1949#issuecomment-891561833

modules/python/init.zsh
```diff
@@ -125,7 +125,7 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
   else
     # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper if 'python' is
     # available in '$path'.
-    if (( ! $+VIRTUALENVWRAPPER_PYTHON )) && (( $#commands[(i)python[23]#] )); then
+    if (( ! $+VIRTUALENVWRAPPER_PYTHON )) && (( $+commands[(i)python[23]] )); then
```
jeffwidman, line 128: This line appeared to still work for me, but given that the syntax above was broken, I'd feel more comfortable reverting this one as well, at least until we understand what exactly it's trying to get at.

jeffwidman:
This virtualenvwrapper functionality is pretty popular with python users, so we def want it to work smoothly out of the box for prezto. 

So while I'm fine letting this PR float for a few days, but after that, if we don't get further feedback we should merge this. 

And then if someone chimes in later (looking at you @indrajitr 😀 ), than we can go back and tweak further.

jeffwidman:
See also discussion in https://github.com/sorin-ionescu/prezto/issues/1949#issuecomment-902188772

That's probably a better way to solve this, so I'm going to close this PR.

PR #1958, opened 2021-10-02
Add exception for JetBrains terminal emulation

tblaschke:
## Proposed Changes

Do not autostart tmux if we start a terminal inside PyCharm or IntelliJ. Should also work for other JetBrains products.

### DIFF
--- a/modules/tmux/init.zsh
+++ b/modules/tmux/init.zsh
@@ -23,7 +23,7 @@ if ([[ "$TERM_PROGRAM" = 'iTerm.app' ]] && \
   _tmux_iterm_integration='-CC'
 fi
 
-if [[ -z "$TMUX" && -z "$EMACS" && -z "$VIM" && -z "$INSIDE_EMACS" && "$TERM_PROGRAM" != "vscode" ]] && ( \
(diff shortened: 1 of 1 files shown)


catflyflyfly:
tmux autostart will mess with JetBrains products (and other IDE as well) on 2 occasions

1. When you start a new terminal from the terminal pane

2. When you start-up the IDE from GUI, ex: Dock & Application for macOS.

   This will ping a notification 
   > IntelliJ IDEA has failed to load the environment from '/bin/zsh'. Integration with tools that rely on environment variables may work incorrectly. [Learn More](https://jb.gg/shell-env)

Solutions:

1. `&& "$TERMINAL_EMULATOR" != "JetBrains-JediTerm"` will handle the terminal pane part.

2. `&& -z "$INTELLIJ_ENVIRONMENT_READER"` will handle the start-up from GUI part.

Note:

This issue occur to me on VSCode start-up as well. This should be an another issue, but TL;DR I solve that with `&& "$__CFBundleIdentifier" != "com.microsoft.VSCode"`.

indrajitr:
Let's roll with this till we arrive at a more ideal solution.

PR #1959, opened 2021-10-28
Allow for explicit path to`rbenv` and `rvm`

johndbritton:
This adds both `$RBENV_PATH` and `$RVM_PATH` as search locations for those version managers in the `ruby` module.

_Motivation_

Homebrew recently moved the default install location for all of it's binaries from `/usr/local/bin` to `/opt/local/bin`. Due to the way Homebrew symlinks binaries (under version folders), setting `$RBENV_ROOT` is not sufficient.

An alternative approach would be to detect `rbenv` using `which` and searching the path.

### DIFF
--- a/modules/ruby/init.zsh
+++ b/modules/ruby/init.zsh
@@ -8,8 +8,8 @@
 #
 
 # Possible lookup locations for manually installed rbenv and rvm.
-local_rbenv_paths=({$RBENV_ROOT,{$XDG_CONFIG_HOME/,$HOME/.}rbenv}/bin/rbenv(N))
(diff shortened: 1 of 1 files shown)


indrajitr:
Nice idea @johndbritton! However, we would probably want to have a more generalized approach to handle the variation of Homebrew paths for Arm and Intel Macs.

johndbritton:
@indrajitr I can add explicit search paths for homebrew but thought this was more likely to get merged as a generic solution.

If you confirm explicitly looking at the homebrew default locations is acceptable I will update.

johndbritton:
Closing in favor of #1997

PR #1962, opened 2021-11-04
Tmux option for separation of SSH and local session autostarts

zbirenbaum:
## Proposed Changes

  - Allow the user to set a variable in .zpreztorc to create different sessions for local and remote connections on auto-start
  - Allow the user to independently name the auto-started local and remote connections
  - Update README.md to represent the above functionality

## Motivation
Currently, there is only a single global configuration for the session auto-started by the tmux module.  It can be extraordinarily inconvenient to be working on a desktop and laptop, and by logging in through SSH on the laptop have the session detach on the desktop. This is especially true in cases when a command is left running in the foreground of one session intentionally, only to have it pop up in your face on logging in to another. This seems like far more convenient and intuitive behavior.

While I realize there are numerous ways to work around this built into tmux, it is extremely convenient to not have to go through the extra steps. I have confirmed that the behavior when none of my added options are set is identical to before, and I maintained the coding style present in the rest of init.zsh, so I can think of no reason not to add in this functionality. 

## Usage
If you would like to create separate background sessions by default for SSH connections and local connections, you may do so by adding the following line to _`${ZDOTDIR:-$HOME}/.zpreztorc`_:

```sh
zstyle ':prezto:module:tmux:auto-start' separate 'yes'
```

If separation is enabled and names are not provided, the sessions will be named _prezto-local_ or _prezto-remote_ respectively. If a name is provided for one of the two connection types but not the other, only the unassigned name will be set to the default. The names of these sessions are customizable in a similar manner to above: 

```sh
zstyle ':prezto:module:tmux:session:local' name '<YOUR DEFAULT LOCAL SESSION NAME>'
zstyle ':prezto:module:tmux:session:remote' name '<YOUR DEFAULT REMOTE SESSION NAME>'
```


zoqaeski:
I've modified the tmux module as well to add a few functions that allow you to spawn a new session from an arbitrary directory, naming the session after the starting directory. This pull request would work well with that.

PR #1964, opened 2021-11-11
Update docker aliases to use Compose V2 #1963

### LINKED ISSUE #1963: docker aliases should update to use Compose V2
### Description

Currently, `dkc` aliases in docker module is using Compose V1.
It should update to use Compose V2.

### Versions

  - Prezto commit: 166cbe2

hidekuro:
Fixes #1963

## Proposed Changes

  - use `docker compose` instead of `docker-compose`


hidekuro:
no ones here.

belak:
Hey, I'm here but getting ready to go on paternity leave so I've had other priorities.

Merges to this repo tend to happen in groups, when the maintainers have time, so I wish you hadn't closed this PR (and deleted the branch) after less than a month - next time, feel free to ping us before you give up so easily. I'm generally around on Libera as belak if you want to ping me directly - I won't always be able to answer right away but I try to respond.

indrajitr:
Looks good. However, I am worried about backward compatibility. Possible to check for the presence of `docker-compose` before resorting to `docker compose`?

hidekuro:
@indrajitr 
It is possible. for example:

```bash
(docker compose version &>/dev/null) && COMPOSE_CMD="docker compose" || COMPOSE_CMD="docker-compose"
alias dkc="$COMPOSE_CMD"
...
unset COMPOSE_CMD
```

But I don't think it's good to continue to support older versions, so I'm won't do that.

hidekuro:
no one here, after all.

belak:
Unfortunately we do have to support backwards compatibility - officially we support back to the latest stable Debian. We can't currently afford to always be at the bleeding edge - there's too much additional work to support this.

I really do appreciate the time you took to submit this, but your attitude ("no ones here", and "no ones here after all") was not appreciated. We are still a team of volunteers at the end of the day. We're people too.

hidekuro:
> Unfortunately we do have to support backwards compatibility - officially we support back to the latest stable Debian. We can't currently afford to always be at the bleeding edge - there's too much additional work to support this.

If I had heard that view earlier, there would have been something I could have done.

hidekuro:
There is an easy way to get my goals without this PR:

Append `alias docker-compose='docker compose'` in end of `~/.zprofile`.
So, forget about this PR.

hidekuro:
@belak Thanks for spending your time on this PR.
I'm sorry I couldn't make a better contribution.

Regards.

PR #1965, opened 2021-11-12
Correct typo in installation instructions

AndrewKvalheim:
The installation instructions reference `${$HOME}`, which fails to evaluate with:

```console
zsh: bad substitution
```


belak (approved):

belak:
Looks good to me! Thanks for catching that!

PR #1968, opened 2021-11-29
Merge Upstream Master

baberthal:
Merge upstream/master and fix conflicts


modules/git/alias.zsh
```diff
@@ -257,9 +219,7 @@ if ! zstyle -t ':prezto:module:git:alias' skip; then
 
   # Tag (t)
   alias gt='git tag'
-  alias gtl='git tag --list'
```
baberthal, line 260: Make sure I have all the right aliases in here

aetimmes:
This should probably be closed.

PR #1979, opened 2022-02-01
Cache output of 'brew --prefix'

### LINKED ISSUE #1978: Calling 'brew --prefix' is slow in the completions module - let's speed it up!
### Description

While profiling my prezto config, I noticed that the completions module was taking a bit longer than I felt it should (fractions of seconds, not whole seconds, but still it's slower than I wanted). After profiling, I determined that a subshell call to `$(brew --prefix)` was the culprit. I'm not the first to notice performance related to `brew --prefix`: https://github.com/Homebrew/brew/issues/3327

### Expected behavior

Since the completions module is one of the core Prezto modules and is enabled by default, I'd like to see it be as speedy as possible.

### Actual behavior

It's not speedy.

### Steps to Reproduce

1. On a Mac, install homebrew
2. Use brew to install curl (brew install curl)
3. Prezto now slows down because now the completions for curl are in a location managed by homebrew

### Versions

  - Prezto commit: latest (7884fb2271d131bb492fafc4eba1c537a821092b)
  - ZSH version: zsh 5.8 (x86_64-apple-darwin20.1.0) 
  - OS information: MacOS 12.2

### Potential solution

Current code:

```zsh
# Add completion for keg-only brewed curl when available.
if (( $+commands[brew] )) \
      && [[ -d "${curl_prefix::="$(brew --prefix 2> /dev/null)"/opt/curl}" ]]; then
  fpath=($curl_prefix/share/zsh/site-functions $fpath)
fi
unset curl_prefix
```

Proposed change (add a cache file):

```zsh
# Add completion for keg-only brewed curl when available.
  # 'brew --prefix' is slow; cache its output
  cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/brew-cache.zsh"
  if [[ "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" || ! -s "$cache_file" ]]; then
    mkdir -p "$cache_file:h"
    echo "brew_prefix=$(brew --prefix 2> /dev/null)" >! "$cache_file"
  fi
  source "$cache_file"
  if [[ -d "${curl_site_functions::=$brew_prefix/opt/curl/share/zsh/site-functions}" ]]; then
    fpath=($curl_site_functions $fpath)
  fi
  unset cache_file brew_prefix curl_site_functions
fi
```

I'm open to discuss other solutions or implementations, but this implementation is similar to what the Perl module does so I modeled off that. I'm currently using it as a .patch file in my personal $ZDOTDIR and this sped up loads by more than 1/10th of a second.

mattmc3:

Fixes #1978

## Proposed Changes

  - Add a cache file for homebrew's `brew --prefix` slow command

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -15,11 +15,19 @@ fi
 fpath=(${0:h}/external/src $fpath)
 
 # Add completion for keg-only brewed curl when available.
-if (( $+commands[brew] )) \
-      && [[ -d "${curl_prefix::="$(brew --prefix 2> /dev/null)"/opt/curl}" ]]; then
-  fpath=($curl_prefix/share/zsh/site-functions $fpath)
+if (( $+commands[brew] )); then
+  # 'brew --prefix' is slow; cache its output
+  cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/brew-cache.zsh"
+  if [[ "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" || ! -s "$cache_file" ]]; then
+    mkdir -p "$cache_file:h"
+    echo "brew_prefix=$(brew --prefix 2> /dev/null)" >! "$cache_file"
+  fi
+  source "$cache_file"
+  if [[ -d "${curl_site_functions::=$brew_prefix/opt/curl/share/zsh/site-functions}" ]]; then
+    fpath=($curl_site_functions $fpath)
+  fi
+  unset cache_file brew_prefix curl_site_functions
 fi
-unset curl_prefix
 
 #
 # Options


indrajitr:
An alternative implementation is #1996 which bypasses all cache management and relies on native zsh expansion and glob qualifiers.  Would be great to have feedback @mattmc3.

mattmc3:
Hmm... `${commands[brew]:h:h}`. That's really clever. My Intel Mac reports `/usr/local` which is correct. When I get the chance, I'll test that my M1 Mac reports `/opt/homebrew`. If that works, this is clearly the preferable implementation compared to introducing caching.

indrajitr:
Thank you very much for verifying @mattmc3. Much appreciated.

mattmc3:
Unfortunately, this doesn't work on M1 installs of homebrew.

```
$ echo ${commands[brew]:h:h}
/usr/local
$ brew --prefix
/opt/homebrew
```

indrajitr:
> After a little more digging, it's because I have symlinked `/opt/homebrew/bin` to `/usr/local/bin`. This is a common thing and not a bug, but I think we can easily fix this with the `:A` absolute-path-follow-symlinks expansion (ie: `echo ${commands[brew]:A:h:h}`).

Yes, you are absolutely right @mattmc3. We should support having `brew` symlinked to arbitrary location. Besides, looking at Homebrew sources, the absolute-path approach would be the more stable path. This is also necessary for handling all `brew --foo` sub-calls. I'll update #1996 in a while.

PR #1981, opened 2022-02-04
Fix crash on open caused by incorrect conditional in python module

### LINKED ISSUE #1977: Setting up Python modules causes zsh to crash.
### Description

When I try to load a Python module with zpreztorc's `zstyle ':prezto:load' pmodule` setting, zsh starts and crashes immediately.

Note that I'm using "powerlevel10k" as the theme.

### Expected behavior

To be able to use Python modules in prezto.

### Actual behavior

1. When using zsh as login shell and setting Python modules to load and starting iterm2, iterm2 terminates after a few seconds.
2. When I start zsh from bash, I get the zsh prompt, but then I get the bash prompt
3. I checked the log from the console of my Mac and found no logs that looked like that.
4. I tried reinstalling gdbm from Homebrew, but that didn't make any difference at all.
5. I tried switching the configuration with zpreztorc and found that zsh started when I disabled loading Python modules.

### Steps to Reproduce

1. Install Prezto.
2. Open zpreztorc and configure it to load powerlevel10k and the Python module.
3. Make sure that zsh crashes.

### Versions

  - Prezto commit: 7884fb2271d131bb492fafc4eba1c537a821092b
  - powerlevel10k version: v1.16.0
  - ZSH version: 5.8 (x86_64-apple-darwin21.0)
  - OS information: macOS Monterey version 12.2

zbirenbaum:
Fixes #1977

The logic currently requires the virtualenv-init plugin to be present in order to properly load the virtualenvwrapper plugin. The virtualenvwrapper plugin does not rely on virtualenv-init though, and the docs claim to load virtualenvwrapper by default, so this is improper behavior. The code begins an 'else' statement if virtualenv-init is not present in which will inevitably discover the script in your pyenv plugins dir and attempt to source it, which will always cause an irrecoverable crash on shell open.

I would call this a fairly major issue. If someone with only one user configured on a linux system with no direct root login capability (as is often default for many distros) would be completely unable to get into their system without mounting it from a boot usb to get rid of the plugin if they closed their current shell after enabling it.

My fix just adds a 'fi' to the proper location, moves up the pyenv_unset plugins, and formats the affected areas to match the scheme of the rest of the file, as it is no longer nested. 

This behavior was already present in the commit referenced by #1949, so this likely will not fix their issue, although, if they had the wrong commit number as the cause, it may.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -114,33 +114,34 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
     pyenv_plugins=(${(@oM)${(f)"$(pyenv commands --no-sh 2> /dev/null)"}:#virtualenv*})
   fi
 
+  # Optionally activate 'virtualenv' plugin when available.
   if (( $pyenv_plugins[(i)virtualenv-init] <= $#pyenv_plugins )); then
-    # Enable 'virtualenv' with 'pyenv'.
     eval "$(pyenv virtualenv-init - zsh)"
-
-    # Optionally activate 'virtualenvwrapper' plugin when available.
-    if (( $pyenv_plugins[(i)virtualenvwrapper(_lazy|)] <= $#pyenv_plugins )); then
-      pyenv "$pyenv_plugins[(R)virtualenvwrapper(_lazy|)]"
-    fi
-  else
-    # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper if 'python' is
-    # available in '$path'.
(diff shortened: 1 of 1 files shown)


zbirenbaum:
Since this is a rather severe bug that is addressed (nearly locked me out of my system) and I now have permission, I am going to go ahead and merge it.

jeffwidman:
This fix was premature, it results in not correctly sourcing `virtualenvwrapper` when `pyenv` isn't installed.

This is what I think https://github.com/sorin-ionescu/prezto/issues/1949#issuecomment-1145702592 was getting at, and I'm experiencing it myself.

I'm heading out the door now, but will look at this later tonight for how to fix this in a way that works both with/without `pyenv` / `virtualenv-init` and without crashing anything.

PR #1982, opened 2022-02-05
Add `-` alias for `cd -`

Enzime:
Taken with love from `oh-my-zsh`

### DIFF
--- a/modules/directory/init.zsh
+++ b/modules/directory/init.zsh
@@ -25,5 +25,6 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
 # Aliases
 #
 
+alias -- -='cd -'
 alias d='dirs -v'
 for index ({1..9}) alias "$index"="cd +${index}"; unset index


Enzime:
@belak @indrajitr any thoughts on this PR?

indrajitr:
I like it 🤣. But I want to have @belak, @jeffwidman and others weigh in to see if this aligns with general aesthetics of prezto.

Enzime:
@sorin-ionescu @indrajitr any thoughts on this PR?

sorin-ionescu:
Using dashes for an alias might cause problems. Are there any known conflicts?

Enzime:
I think this should be low risk as `ohmyzsh` has had this alias for 13 years now and I'm not aware of any issues.

This alias should only have an effect when someone uses a single dash as a command i.e.
```
$ -
$ - apple
```

The only conflict I'm aware of is a `zsh` builtin for `-` which does the following:

``The  command  is  executed  with  a `-' prepended to its argv[0] string.``

jeffwidman (approved):
👍 from me.

belak:
Looks good to me! Thanks!

PR #639, opened 2014-07-14
Show git dirty/clean status through branch color

mavidser:
I think it's a nifty enhancement, showing clean/dirty status on the left portion of the screen.

PR #1990, opened 2022-03-07
Allow users to set a dirty-branch format

jboulter11:
## Proposed Changes

- Implements dirty-branch format which overrides regular branch format when your git status is dirty.  Primarily, this is useful for changing the color of the branch name.
- Taken directly from #639 which implemented this, but also included a change to the sorin prompt.  That change is not present in this PR.
- I have personally had this change applied locally for years.

Example usage, taken from my own prompt:
```sh
zstyle ':prezto:module:git:info:branch' format '%%B%F{183}@%b%f%%b'
zstyle ':prezto:module:git:info:dirty-branch' format '%%B%F{225}@%b%f%%b'
```

Clean:
Screen Shot 2022-03-06 at 4 04 32 PM
Dirty:
Screen Shot 2022-03-06 at 4 04 43 PM

### DIFF
--- a/modules/git/functions/git-info
+++ b/modules/git/functions/git-info
@@ -417,6 +417,11 @@ function git-info {
   if (( dirty > 0 )); then
     zstyle -s ':prezto:module:git:info:dirty' format 'dirty_format'
     zformat -f dirty_formatted "$dirty_format" "D:$dirty"
+    # Overwrite branch format to use dirty-branch format
+    zstyle -s ':prezto:module:git:info:dirty-branch' format 'branch_format'
+    if [[ -n "$branch" && -n "$branch_format" ]]; then
+      zformat -f branch_formatted "$branch_format" "b:$branch"
+    fi
   else
     zstyle -s ':prezto:module:git:info:clean' format 'clean_formatted'
   fi


belak:
It's been a while since I've looked at the git-info formatting code... as far as I can tell, we don't do anything like this anywhere else in git-info (overriding a prompt variable depending on another one). It still may be worthwhile though.

~~I *think* the example you gave can be done without this patch, by setting the dirty format to just include the color format.~~ I took another look at this and it doesn't seem like it.

So, I think this would be worthwhile, I just worry about the ballooning complexity if there are other use cases where people want to change other parts of the prompt based on git statuses. That being said, I'm not sure if there's a good way to account for that, so I guess we could just take things as they come.

Sorry, that was a bit of a ramble - the TL;DR is that I'm inclined to accept and merge this, but I'd be curious what the other maintainers think.

jboulter11:
@belak thanks for your feedback.  I agree that having these fields interact could be a slippery slope, but I think this one is pretty straightforward and the functionality is easy to understand and reason about.

If we are waiting on others, should we should give them a tag to get their attention?

jboulter11:
@belak since no one has spoken up at this point, think we can merge this one?

belak:
That works! Thanks for the reminder ping!

jboulter11:
Thank you! :)

PR #1992, opened 2022-03-13
Only export env variables in zprofile if unset

akanter:
As per [zsh documentation](https://zsh.sourceforge.io/Intro/intro_3.html) environment variables should be expected to be in `.zshenv` and not be overridden in `.zprofile`. This change modified EDITOR, VISUAL, PAGER, and LESS to only set prezto defaults if they were not previously set in the loading order.

See:
https://github.com/nix-community/home-manager/issues/2739
https://github.com/nix-community/home-manager/issues/2751

### DIFF
--- a/runcoms/zprofile
+++ b/runcoms/zprofile
@@ -9,7 +9,7 @@
 # Browser
 #
 
-if [[ "$OSTYPE" == darwin* ]]; then
+if [[ -z "$BROWSER" && "$OSTYPE" == darwin* ]]; then
   export BROWSER='open'
 fi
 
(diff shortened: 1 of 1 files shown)


pinpox:
Took me for ever to find where EDITOR was being overriden in my config.  Big +1 on this!

indrajitr (approved):
LGTM.
Minor nit: Possible to wrap `export BROWSER='open'` inside similar condition as well for the sake of consistency?

akanter:
Thanks for the review, @indrajitr ! I modified the diff to do the same for BROWSER as well as LESSOPEN. Should be good to merge now.

PR #1994, opened 2022-03-18
fix: use inflating binaries instead of deflating in unarchive function

blanorama:
Fixes: 1

## Proposed Changes

  - when the system is checked for the different (un)-archiving binaries, in this function it should be checked for binaries inflating the package instead of deflating, since we want to unarchive


jeffwidman (approved):

jeffwidman:
Thanks!

PR #1997, opened 2022-04-01
zprofile: Expand the default list of well known paths

### LINKED ISSUE #1988: Default to including the new homebrew path for apple silicon (m1) macs
On an Apple Silicon / M1 mac, Homebrew now installs binaries to `/opt/homebrew/{bin,sbin}` instead of `/usr/local/{bin,sbin}`.

Given the popularity of homebrew, IMO we should update Prezto to default to including this new path.

This could be done at several different abstraction layers:
* the `homebrew` module
* hardcoding the path (as demonstrated here: https://github.com/sorin-ionescu/prezto/pull/1974/files)

The downside of adding it to the homebrew module is that it's only loaded during an interactive session: https://github.com/sorin-ionescu/prezto/blob/20641f3447e9caef169c8210ff540921419ce448/runcoms/zshrc#L8-L11

Whereas I have a few places in my runcoms where I set env vars using something like:

```shell
export VIRTUALENVWRAPPER_PYTHON=$(brew --prefix)/bin/python3
```

So having `brew` as part of the path for all sessions and loaded early is preferable.

Ideally we only add the new `/opt/homebrew` location if the machine is actually apple Silicon, but that's a trivial thing since if it's an intel-based box, that folder is unlikely to exist.

Given the above, it may be simplest to go with the solution in https://github.com/sorin-ionescu/prezto/pull/1974/files and be done with it.

Thoughts from others?

indrajitr:
## Proposed Changes

Two additional sets of paths are now added to the default list of well known paths: `$HOME/{bin,sbin}` and `/opt/{homebrew,local}/{bin,sbin}`.

- `$HOME/{bin,sbin}`: Most users have custom scripts in `$HOME/bin` anyway, we might as well honor those. `$HOME/sbin` is not really common, but we can keep it for consistency.
- `/opt/{homebrew,local}/{bin,sbin}`: With Homebrew changing default installation location in macOS on Apple Silicon which will eventually become ubiquitous, we have a good reason to add these paths by default. While at it, we also honor MacPorts installation.

In all cases, we add them _iff_ the paths actually exist, not otherwise. This has the side effect of a newly installed program not available immediately in the `$path` in a mint fresh system until the shell is reloaded. But that's a minor inconvenience to keep the global `$path` from getting unnecessarily bloated.

Fixes #1988, #1959


jeffwidman (approved):

ickc:
I think this behavior is a bit intrusive.

Paths under `$HOME` & `/opt` are non-defaults and users should be setting them by themselves, especially the order of these paths.

Are there any chance you would consider reverting this, or consider making this configurable?

Thanks.

belak:
The /opt/homebrew paths is the default for Homebrew on Apple Silicon devices so we're definitely keeping that. Is there a reason you're against ~/bin and ~/sbin? It's a common pattern to put scripts in there and if it doesn't exist, it shouldn't cause any issues.

Has this change caused problems for you? Do you have another reason for being opposed to this change?

In general if you have an advanced use case not supported by the provided runcoms, you're encouraged to copy the runcoms and modify them to fit your needs.

ickc:
The diff is

```
path=(
-   /usr/local/{bin,sbin}
+   $HOME/{,s}bin(N)
+   /opt/{homebrew,local}/{,s}bin(N)
+   /usr/local/{,s}bin(N)
   $path
 )
```

This is a few things happening at the same time, 

1. ordering: previously only `/usr/local` is added, prepending is already opinionated, but now there are much more going on, some might prefer the order given, some might want the one in `$HOME` to be appended instead for reasons such as security.
2. the consideration of addition of `/opt/homebrew` vs. the others has subtle differences, so discussing this one at a time:
    - For homebrew usage, the Intel homebrew was in `/usr/local` which is already in the `PATH`, a deliberate choice by their maintainers, but to some it is a mistake. The departure to `/opt/homebrew` arguably is the answer to the critics. Now prezto might want this to be in people's `PATH` automatically again, which is opinionated. I'm not going to argue against it, but it would be better for it to be configurable.
    - Other paths, notably that from macports, `/opt/local` hasn't been changed for years, and why should prezto all of a sudden do the smart thing for the end users to auto discover them? We may as well throw in default paths of all other package manager as well... My problem is not that being smart is not ok, but then for people having both homebrew and macports, this is going to be problematic. The end users should decide in which order they want that, and even for more advanced cases they would dynamically load either homebrew or macports but not both.

(Yes, the change does causes problems for me as I'm using both homebrew and macports and also have a dynamical way to load either both or only one of them. This change break mine partially because zprofile is loaded before zshrc. But nevermind about my specific situation because I already fixed that in my end.)

TL;DR: The main problem I have with this change is because it is too intrusive and it is breaking things for me unexpectedly. I guess a framework like this should be opinionated, but the PATH is a really sensitive matter (which can change people's environment in subtle ways) and if not doing the smart and opinionated thing has been working for people for years, why do that now? I guess the simplest remedy may be to just give people an option that the path would never be modified.

indrajitr:
> ordering: previously only /usr/local is added, prepending is already opinionated, but now there are much more going on, some might prefer the order given, some might want the one in $HOME to be appended instead for reasons such as security.

Ordering is a side effect, I admit. Is there a recommended best practice that can we can referred to? If there is a compelling reason to rearrangement, I'd be all for it.

> ... Now prezto might want this to be in people's `PATH` automatically again, which is opinionated. I'm not going to argue against it, but it would be better for it to be configurable.

Yes, it is indeed opinionated and intentional - to make to work for _most users_, not _all users_. Ability to conveniently customize/configure `$PATH` with a set of sane default (for _most users_) is the goal. Is there a way we can make this happen for `$PATH` in zsh?

> Other paths, notably that from macports, `/opt/local` hasn't been changed for years, and why should prezto all of a sudden do the smart thing for the end users to auto discover them?

Not honoring the macports defaults, was a feature gap IMHO.

> We may as well throw in default paths of all other package manager as well... 

Fair point :) However, other package managers in Linux/*BSD world (`pkg`, `pacman`, `apt`, `apk`, `rpm`/`yum`) follow a well defined spec/convention. That's not really the case in macOS world (`homebrew`, `macports` are _"community creation"_ - not from Apple). I guess, that's why they end up getting _special treatment_.

> My problem is not that being smart is not ok, but then for people having both homebrew and macports, this is going to be problematic. The end users should decide in which order they want that, and even for more advanced cases they would dynamically load either homebrew or macports but not both.

Indeed, these are advanced scenarios that require [customization](https://github.com/sorin-ionescu/prezto#customization).

> ... I guess the simplest remedy may be to just give people an option that the path would never be modified.

As stated above, if we can achieve this in a nice way in prezto/zsh it would be fantastic. Recommendation(s), or even better, a PR or two will be a great start to the discussion.

PR #1998, opened 2022-04-05
fix: extend deflating flag for pixz in unarchive function

blanorama:
Fixes 1

## Proposed Changes

  - when the system is checked for the different (un)-archiving binaries, in this function the `pixz` binary should be extended with the deflating flag
  - (missed that one in my former pull request)


belak:
LGTM! Thanks!

PR #1999, opened 2022-04-07
feat(module/history): add full configurability

Shea690901:
While previusly configurable, it was inconsistent with other
configuration options and it was missing configurability of in memory /
on disc history size.

Signed-off-by: Shea690901 <ginny690901@hotmail.de>

### DIFF
--- a/modules/history/README.md
+++ b/modules/history/README.md
@@ -38,6 +38,35 @@ Alternately, you can set `HISTFILE` manually to _`${ZDOTDIR:-$HOME}/.zhistory`_.
 
 - `history-stat` lists the ten most used commands
 
+## Settings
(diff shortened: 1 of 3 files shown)


indrajitr:
While `zstyle` definitions are usually supposed to be honored dynamically (without requiring `zsh` to be reloaded), we do use it to set global variables (which require `zsh` reload) anyway. Thus, we might support this configurability anyway.

@Shea690901, thank you for the PR! Can you please also update `zpreztorc` as well (with commented-out options) for visibility to the users?

Shea690901:
Done!

indrajitr:
Thank you @Shea690901!

PR #2001, opened 2022-04-08
feat(module/ssh): add globbing for ssh-add param

### LINKED ISSUE #2000: suboptimal ssh-agent configuration
### Description
While it's nice to add different keys via `zstyle ':prezto:module:ssh:load' identities 'id_rsa' 'id_rsa2' 'id_github'`, normally each secret key `key` is accompanied  by a public key `key.pub`. As such the easiest way would be to use `zstyle ':prezto:module:ssh:load' identities '*.pub(:r)'`
### Expected behavior

File globbing to happen and all of my key files added.

### Actual behavior

No key get's added.

### Steps to Reproduce

1. generate more then one key, stored in separate files
2. use above configuration using globbing
3. terminate current session (ssh-agent has to die)
4. start new session
5. use `ssh-add -l` to check that not even password-less keys got added
6. 
### Versions

  - Prezto commit: current
  - ZSH version: 5.8.1
  - OS information: independent on OS

Shea690901:
fixes #2000

Signed-off-by: Shea690901 <ginny690901@hotmail.de>

### DIFF
--- a/modules/ssh/init.zsh
+++ b/modules/ssh/init.zsh
@@ -52,9 +52,9 @@ if ssh-add -l 2>&1 | grep -q 'The agent has no identities'; then
   # program specified by SSH_ASKPASS and open an X11 window to read the
   # passphrase.
   if [[ -n "$DISPLAY" && -x "$SSH_ASKPASS" ]]; then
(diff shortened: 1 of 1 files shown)


indrajitr (approved):
Nice!

PR #2003, opened 2022-05-16
Add option to disable http-serve in utility module

dnguyenzd:

Fixes #

## Proposed Changes

In my machine (MacBook Pro, Monterey), prezto init takes 0.4-0.5s. Here's logs that I put before and after prezto init in zshrc:

```
21:17:08.873183000 before prezto init
21:17:09.342031000 after prezto init
```

It's fast but I want to get it down further. After some debugging, I narrow down to `utility` module, or more specifically the setting up of `http-serve` alias in this module (I think it's due to the call to python to check version number).

Not all people need this alias, so added a toggle to allow disabling `http-serve` entirely. After disabling it, the loading is down to less than 0.2s

```
21:27:22.248647000 before prezto init
21:27:22.407725000 after prezto init
```

### DIFF
--- a/modules/utility/README.md
+++ b/modules/utility/README.md
@@ -181,6 +181,12 @@ In addition, the following aliases have been added:
 
 - `http-serve` serves a directory via HTTP.
 
+To disable `http-serve`, add the following to _`${ZDOTDIR:-$HOME}/.zpreztorc`_.
+
+```sh
+zstyle ':prezto:module:utility' http-serve 'no'
+```
+
 ## Functions
 
 ### General
--- a/modules/utility/init.zsh
+++ b/modules/utility/init.zsh
@@ -194,14 +194,16 @@ fi
 # Miscellaneous
 
 # Serves a directory via HTTP.
-if (( $#commands[(i)python(|[23])] )); then
-  autoload -Uz is-at-least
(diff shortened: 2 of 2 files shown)


indrajitr:
Interesting idea. But we need to figure out a way to reduce the performance cost even for those who need to use http serve.

dnguyenzd:
to be honest, I don't feel that alias belongs to `utility` module. 

It assumes that all people who have python will want to do http serve. What if users prefer nodejs to serve http a directory? It's probably more suitable for `python` module, or removed altogether and let users decide what they like to use

For me, I want to use `utility` module for its other aliases and integration with GNU Core Utilities. I have python installed but I don't use python in day-to-day work

jeffwidman:
@dnguyenzd now that https://github.com/sorin-ionescu/prezto/pull/2014 was merged, how much time improvement does this PR actually add?

My guess is very little, in which case we're probably fine closing this rather than adding another option, but maybe I'm wrong.

The other alternative is to drop this alias altogether... I can't think of a time I wanted this, and even if I did, I doubt I'd remember these aliases exist so would probably end up cobbling something together manually.

belak:
> now that #2014 was merged, how much time improvement does this PR actually add?

This is a good point - I think this can be closed and if it's still too slow on startup we can move all of it to a function so it can be done on call to http-serve

PR #2009, opened 2022-07-08
Fix duplicate vcs config

furusax0621:


furusax0621:
Sorry, I made a mistake.

PR #2021, opened 2022-10-25
Set `EXTENDED_GLOB` in the python module

jeffwidman:
This is needed to find python. 

It is often already set because a user has the directory or completion module loaded before this, but that's not always true.

See extensive debugging / further explanation here: https://github.com/sorin-ionescu/prezto/issues/1949

This does not fully resolve that issue, as there's another enhancement that I'll put up as a separate PR.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -8,6 +8,12 @@
 #   Indrajit Raychaudhuri <irc@indrajit.com>
 #
 
+#
+# Options
+#
+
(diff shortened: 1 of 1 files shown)


belak:
I’m just merging this - I thought we already made this change!

jeffwidman:
Thanks!

jeffwidman:
We agreed on it, but no one ever shipped it, and now I need to work on some python, so have some extra motivation :-D

Issue #2022, opened 2022-10-26
virtualenvwrapper broke after d840f0fc7bb9e604fedef40eff4ed53f4c3c60f6


jeffwidman:
Sourcing `virtualenvwrapper` broke for me after d840f0fc7bb9e604fedef40eff4ed53f4c3c60f6 / https://github.com/sorin-ionescu/prezto/pull/1981.

I do not have `pyenv` installed, just plain ol' `virtualenvwrapper`.

Looking at `git blame`, there's a _lot_ of prior art here around trying to support all the permutations of `pyenv` and various plugins:
* https://github.com/sorin-ionescu/prezto/issues/1413
* https://github.com/sorin-ionescu/prezto/pull/1414
* https://github.com/sorin-ionescu/prezto/pull/1433
* https://github.com/sorin-ionescu/prezto/pull/1434

Given all the above, I suspect the easiest thing to do may be to revert d840f0fc7bb9e604fedef40eff4ed53f4c3c60f6 and then try again to very surgically support the case where `pyenv` + `virtualenvwrapper` are present, but `virtualenv-init` is missing.

PR #2023, opened 2022-10-26
Support `virtualenvwrapper` with / without `pyenv` `virtualenv-init` or `virtualenvwrapper` plugins

jeffwidman:
The desired logic is:

For the `pyenv` plugins `virtualenv-init` and `virtualenvwrapper`:
1. If either plugin is present, activate it
2. If `virtualenvwrapper` plugin is not present, then [fallback to standard `virtualenvwrapper`](https://github.com/sorin-ionescu/prezto/pull/1414#issuecomment-320306421).
3. If `virtualenvwrapper` plugin is present, then don't fallback to
   standard `virtualenvwrapper`, [regardless of whether `virtualenv-init` is present](https://github.com/sorin-ionescu/prezto/pull/1981#issue-1123766676).

Previously, if the `virtualenv` command was present but `pyenv` was
missing, then the fallback wouldn't be hit. This bug was introduced by
https://github.com/sorin-ionescu/prezto/pull/1981/ which ensured that
the `pyenv` `virtualenvwrapper` plugin was activated if present,
regardless of the presence of the `virtualenv-init` plugin.

As an optimization, the check for the `pyenv` plugins are skipped if
`pyenv` itself isn't found.

Since we only want to fallback if the `pyenv` `virtualenvwrapper` plugin
is missing, but that's buried within the `pyenv` logic and we also need
to handle when `pyenv` itself is missing, this switches to using a flag
variable.

I also renamed the `virtualenv_sources` var to
`virtualenvwrapper_sources` as `virtualenv` is distinct from
`virtualenvwrapper`, so using one name for a var that is really about
the other is confusing.

Looking at `git blame`, there's a _lot_ of prior art here around trying
to support all the permutations of `pyenv` and various plugins:
* https://github.com/sorin-ionescu/prezto/issues/1413
* https://github.com/sorin-ionescu/prezto/pull/1414
* https://github.com/sorin-ionescu/prezto/pull/1433
* https://github.com/sorin-ionescu/prezto/pull/1434

So we need to be extremely careful to continue to support all these permutations.

Fix https://github.com/sorin-ionescu/prezto/issues/2022

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
   # look for plugins of interest. Scanning shell '$path' isn't enough as they
   # can exist in 'pyenv' synthesized paths (e.g., '~/.pyenv/plugins') instead.
   local -a pyenv_plugins
+  local pyenv_virtualenvwrapper_plugin_found
   if (( $+commands[pyenv] )); then
     pyenv_plugins=(${(@oM)${(f)"$(pyenv commands --no-sh 2> /dev/null)"}:#virtualenv*})
+
+    # Optionally activate 'virtualenv-init' plugin when available.
+    if (( $pyenv_plugins[(i)virtualenv-init] <= $#pyenv_plugins )); then
+      eval "$(pyenv virtualenv-init - zsh)"
+    fi
+
+    # Optionally activate 'virtualenvwrapper' plugin when available.
+    if (( $pyenv_plugins[(i)virtualenvwrapper(_lazy|)] <= $#pyenv_plugins )); then
+      pyenv "$pyenv_plugins[(R)virtualenvwrapper(_lazy|)]"
+      pyenv_virtualenvwrapper_plugin_found="true"
+    fi
+
+    unset pyenv_plugins
   fi
 
-  # Optionally activate 'virtualenv' plugin when available.
-  if (( $pyenv_plugins[(i)virtualenv-init] <= $#pyenv_plugins )); then
-    eval "$(pyenv virtualenv-init - zsh)"
+  if [[ $pyenv_virtualenvwrapper_plugin_found != "true" ]]; then
+    # Fallback to standard 'virtualenvwrapper' if 'python' is available in '$path'.
+    if (( ! $+VIRTUALENVWRAPPER_PYTHON )) && (( $#commands[(i)python[23]#] )); then
+      VIRTUALENVWRAPPER_PYTHON=$commands[(i)python[23]#]
+    fi
+
+    virtualenvwrapper_sources=(
+      ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
+      /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
+    )
+    if (( $#virtualenvwrapper_sources )); then
+      source "$virtualenvwrapper_sources[1]"
+    fi
+
+    unset virtualenvwrapper_sources
   fi
 
-  # Optionally activate 'virtualenvwrapper' plugin when available.
-  if (( $pyenv_plugins[(i)virtualenvwrapper(_lazy|)] <= $#pyenv_plugins )); then
-    pyenv "$pyenv_plugins[(R)virtualenvwrapper(_lazy|)]"
-  fi
-  
-  unset pyenv_plugins
-  
-else
-  # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper if 'python' is
-  # available in '$path'.
-  if (( ! $+VIRTUALENVWRAPPER_PYTHON )) && (( $#commands[(i)python[23]#] )); then
-    VIRTUALENVWRAPPER_PYTHON=$commands[(i)python[23]#]
-  fi
-
-  virtenv_sources=(
-    ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
-    /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
-  )
-  if (( $#virtenv_sources )); then
-    source "$virtenv_sources[1]"
-  fi
-
-  unset virtenv_sources
+  unset pyenv_virtualenvwrapper_plugin_found
 fi
 
 # Load conda into the shell session, if requested.


modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
...
-else
-  # Fallback to 'virtualenvwrapper' without 'pyenv' wrapper if 'python' is
-  # available in '$path'.
-  if (( ! $+VIRTUALENVWRAPPER_PYTHON )) && (( $#commands[(i)python[23]#] )); then
-    VIRTUALENVWRAPPER_PYTHON=$commands[(i)python[23]#]
+    # Optionally activate 'virtualenvwrapper' plugin when available.
+    if (( $pyenv_plugins[(i)virtualenvwrapper(_lazy|)] <= $#pyenv_plugins )); then
+      pyenv "$pyenv_plugins[(R)virtualenvwrapper(_lazy|)]"
+      $pyenv_virtualenvwrapper_plugin_found=true
+    fi
```
jeffwidman, line 132: The only change in this block is indenting within the `pyenv` check + setting the `pyenv_virtualenvwrapper_plugin_found` var.

modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
...
+
+    virtualenvwrapper_sources=(
+      ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
+      /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
+    )
+    if (( $#virtualenvwrapper_sources )); then
+      source "$virtualenvwrapper_sources[1]"
+    fi
+
+    unset virtualenvwrapper_sources
```
jeffwidman, line 151: The only change in all this block of logic is the rename of `virtenv_sources` to `virtualenvwrapper_sources`

modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
...
+    unset pyenv_plugins
   fi
 
-  virtenv_sources=(
-    ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
-    /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
-  )
-  if (( $#virtenv_sources )); then
-    source "$virtenv_sources[1]"
+  if [[ !$pyenv_virtualenvwrapper_plugin_found ]]; then
```
jeffwidman, line 137: This changes from an `else` to checking the flag variable... that way if the `virtualenv` command was found but `pyenv` was missing we still fall back to trying to activate standard `virtualenvwrapper`.

modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
   # look for plugins of interest. Scanning shell '$path' isn't enough as they
   # can exist in 'pyenv' synthesized paths (e.g., '~/.pyenv/plugins') instead.
   local -a pyenv_plugins
+  local pyenv_virtualenvwrapper_plugin_found=false
```
jeffwidman, line 119: I am actually weak on the `zsh` foo, so unclear if this needs to be designated as a `local` or not since it's `unset` at the end of the block.

I am a little confused because `pyenv_plugins` is marked as a local`, but `virtenv_sources` wasn't marked as a local...

modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
...
+      VIRTUALENVWRAPPER_PYTHON=$commands[(i)python[23]#]
+    fi
+
+    virtualenvwrapper_sources=(
+      ${(@Ov)commands[(I)virtualenvwrapper(_lazy|).sh]}
+      /usr/share/virtualenvwrapper/virtualenvwrapper(_lazy|).sh(OnN)
+    )
+    if (( $#virtualenvwrapper_sources )); then
+      source "$virtualenvwrapper_sources[1]"
+    fi
```
jeffwidman, line 149: The only change in all this block of logic is the rename of `virtenv_sources` to `virtualenvwrapper_sources`

modules/python/init.zsh
```diff
@@ -116,38 +116,42 @@ if (( $+VIRTUALENVWRAPPER_VIRTUALENV || $+commands[virtualenv] )) \
   # look for plugins of interest. Scanning shell '$path' isn't enough as they
   # can exist in 'pyenv' synthesized paths (e.g., '~/.pyenv/plugins') instead.
   local -a pyenv_plugins
+  local pyenv_virtualenvwrapper_plugin_found
```
jeffwidman, line 119: Testing revealed this line isn't necessary... but I was unclear if it's more idiomatic to declare the var since otherwise it's only set deep in nested `if`, but then tested at a higher level of scope.

Also unclear if when declaring it's best to declare it as a `local` or not... I don't understand why `pyenv_plugins` is declared as a `local` but `virtenv_sources` isn't...

jeffwidman:
@zbirenbaum @diraol @zachwhaley since you were all involved in prior PR's around this section of code logic, can you take this for a test drive to ensure I'm not accidentally breaking something in your use cases with this.

Also @indrajitr / @belak I'd appreciate some sanity check reviews.

modules/python/init.zsh
belak: I think pyenv_plugins is declared locally so we can specify that it's an array. It seems reasonable to also declare pyenv_virtualenvwrapper_plugin_found up here.

belak (approved):
I really like this change - I think it's a more formalized version of the logic we wanted in the first place and starts removing some of the magic/inconsistencies.

It's way easier to say "pyenv's virtualenv plugin will be used when available, otherwise we fall back to the plain virtualenvwrapper".

zbirenbaum:
I like this too. My change was a quick one to prevent people from getting locked out of their systems since that’s a really bad bug to have. This is a better formalization.

jeffwidman:
Thanks all!

I'm going to merge given the positive reception. If we find further bugs/unhandled edge cases, I'm happy to take another look.

PR #2029, opened 2022-12-23
Fix zcompile race condition

### LINKED ISSUE #2028: zcompile race condition
### Description

This is a follow-up to the ticket, which I think was prematurely closed: https://github.com/sorin-ionescu/prezto/issues/318

### Expected behavior

I run iTerm (others run tmux/screen), so multiple zsh will be open at the same time.

I expect there to be no error/warning

### Actual behavior

I almost always end up with:
```
/Users/huyz/.zprezto/runcoms/zlogin:zcompile:13: can't write zwc file: /Users/huyz/.cache/prezto/zcompdump.zwc
```
Looks like a race condition.

### Steps to Reproduce

1. Set up iTerm/tmux/screen to start multiple zsh shells simultaenously
2. Start iTerm/tmux/screen
3. Repeat if necessary

### Suggested solution

A simple fix like what oh-my-zsh has implemented for the same problem: https://github.com/ohmyzsh/ohmyzsh/pull/11345
```zsh
if command mkdir "${ZSH_COMPDUMP}.lock" 2>/dev/null; then
  zcompile "$zcompdump"
  command rmdir "${ZSH_COMPDUMP}.lock" 
fi
```

### Versions

  - Prezto commit: e3a9583f3370e11a0da1414d3f335eac40c1e922
  - ZSH version: 5.8.1
  - OS information: macOS 12.6.2 (Monterey, ARM)

huyz:
Fixes #2028

## Proposed Changes

If several login shells are started at the same time, only one should be running `zcompile`.  This is the same fix as https://github.com/ohmyzsh/ohmyzsh/pull/11345

### DIFF
--- a/runcoms/zlogin
+++ b/runcoms/zlogin
@@ -10,7 +10,10 @@
   # Compile the completion dump to increase startup speed.
   zcompdump="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/zcompdump"
   if [[ -s "$zcompdump" && (! -s "${zcompdump}.zwc" || "$zcompdump" -nt "${zcompdump}.zwc") ]]; then
-    zcompile "$zcompdump"
+    if command mkdir "${zcompdump}.zwc.lock" 2>/dev/null; then
+      zcompile "$zcompdump"
+      command rmdir  "${zcompdump}.zwc.lock" 2>/dev/null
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
Thanks!

PR #2032, opened 2023-01-29
Fix osx module manp function

mattmc3:
Opening man pages using Preview.app using `manp` broke in Ventura. There's now a new way to do it. This PR updates the `manp` function in the osx module. See [article](https://scriptingosx.com/2022/11/on-viewing-man-pages-ventura-update/).

### DIFF
--- a/modules/osx/functions/manp
+++ b/modules/osx/functions/manp
@@ -9,7 +9,7 @@ function manp {
   local page
   if (( $# > 0 )); then
     for page in "$@"; do
-      man -t "$page" | open -f -a Preview
+      mandoc -T pdf "$(/usr/bin/man -w $page)" | open -fa Preview
     done
   else
     print 'What manual page do you want?' >&2


jeffwidman (approved):
Thanks!

Will this work for all mac versions or is it only going to work in Ventura? I'm probably 👍 on shipping even if only works on Ventura, as most folks who would update to new versions of `prezto` will also update to new versions of MacOS, and this probably isn't a commonly used feature... but be good to know how much breakage we're adding here...

mattmc3:
I'm not sure which version mandoc began shipping with, but it's not new in Ventura. I don't run older versions of macOS to test fully, but I did run a quick GitHub actions verification using a macos-11 workflow (BigSur, 2020) with the command `MANPAGER=cat man mandoc | grep PDF`. It returned `PDF-1.1 output may be generated by -T pdf.  See PostScript Output for -O`, which is the same output as what I get in Ventura. `mandoc` is also installable via homebrew. I did read a couple places where earlier versions required rosetta2 on Apple Silicon, but if you were an early adopter of M1 chips that was par for the course and pretty seamless.

If people report a problem, ping me and I'll add an `if else` wrapper to add back the old command for earlier versions.

jeffwidman:
I think we're good, no need to add an `if else`, if someone else cares enough they can go to the trouble of adding a PR. 👍

PR #2033, opened 2023-01-31
Add GNU awk to gnu-utility module

jbsilva:
## Proposed Changes

Add GNU awk (gawk) to gnu-utility module.

I needed this when using [gawk](https://www.gnu.org/software/gawk/) downloaded from [Homebrew](https://formulae.brew.sh/formula/gawk).


jeffwidman (approved):

jeffwidman:
Makes sense, thanks!

PR #2037, opened 2023-04-13
Add zstyle to skip aliases for additional modules

### LINKED ISSUE #2036: Add zstyle to skip aliases in more modules beyond just git
### Description

Similar to #310, It would be nice if Prezto supported a `zstyle` customization like the one in the git module that allows skipping the creation of aliases.

```zsh
# skip emacs alias creation
zstyle ':prezto:module:emacs:alias' skip 'yes'

# or, skip all alias creation in supported modules
zstyle ':prezto:module:*:alias' skip 'yes'
```

This is helpful for modules like homebrew, for example, where it's nice to have the environment variable functionality but not necessarily the aliases. Aliases are highly opinionated shortcuts and can be numerous in any given module, and they are not simple to undo without a large series of `unalias` commands in a users' config. A `zstyle` would solve this, and it's not a new concept since the `git` module already does this.

This feature only makes sense for modules that have a purpose beyond just creating aliases. So, for example, `dnf`, `rsync`, etc would not need this `zstyle` unless someday their purpose expands.

This can be implemented with a simple `if` block similar to how the `git` module already does it. PR incoming.

mattmc3:

Fixes #2036 

## Proposed Changes

Simply adds an optional `zstyle` `if` block to allow skipping alias creation for the following modules:

  - directory
  - emacs
  - fasd
  - history
  - homebrew
  - macports
  - node
  - perl
  - python
  - rails
  - ruby
  - screen
  - tmux

Be sure to view the diff with whitespace ignored (`?w=1`). It is much simpler to see that the only change here is the addition of an `if` block. https://github.com/mattmc3/prezto/commit/a7d6a1a5553d32700c77924a9f0e947c35e61bd9?w=1

### DIFF
--- a/modules/directory/init.zsh
+++ b/modules/directory/init.zsh
@@ -25,6 +25,8 @@ unsetopt CLOBBER            # Do not overwrite existing files with > and >>.
 # Aliases
 #
 
-alias -- -='cd -'
-alias d='dirs -v'
-for index ({1..9}) alias "$index"="cd +${index}"; unset index
+if ! zstyle -t ':prezto:module:directory:alias' skip; then
+  alias -- -='cd -'
+  alias d='dirs -v'
+  for index ({1..9}) alias "$index"="cd +${index}"; unset index
+fi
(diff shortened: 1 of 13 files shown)


mattmc3:
Simplified diff link for this PR: https://github.com/sorin-ionescu/prezto/pull/2037/files?w=1

belak:
I really like this change - thanks for submitting it!

PR #2041, opened 2023-04-27
Remove couple of deprecation warnings

indrajitr:
These deprecations are close to 2 years old now and about time to clean them up.

### DIFF
--- a/modules/history/init.zsh
+++ b/modules/history/init.zsh
@@ -42,28 +42,3 @@ if ! zstyle -t ':prezto:module:history:alias' skip; then
   # Lists the ten most used commands.
   alias history-stat="history 0 | awk '{print \$2}' | sort | uniq -c | sort -n -r | head"
 fi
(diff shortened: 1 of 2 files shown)


jeffwidman (commented):
Code looks good, but TBH, 2 years isn't that long in the world of shell scripts... many folks setup their system and then rarely update their dotfiles more than once every few years... so I think I'd float this one open another 1-2 years before merging?

indrajitr:
@jeffwidman, your past self wanted _just kill it_ back then https://github.com/sorin-ionescu/prezto/pull/1895#issuecomment-828335494 😄 -- and all for good reasons that @belak and I agreed with.

That said, I don't mind waiting out longer if we really must.

jeffwidman:
Ah good call, I agree with my reasoning in my past self.

jeffwidman (approved):

PR #2042, opened 2023-04-27
python: Expand Python command pattern per convention

indrajitr:
Use `(i)python[0-9.]#` as subscript for commands associative array, following the [convention](https://github.com/zsh-users/zsh/blob/858b8de3d70fe76a3637c281bc2c8e3a6d961a2c/Completion/Unix/Command/_python#L1) of detection python more exhaustively.

This is also a follow-up to the previous [discussion](https://github.com/sorin-ionescu/prezto/issues/1949#issuecomment-1291224049) where @jeffwidman wanted me to address this.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -50,7 +50,7 @@ fi
 unset local_pyenv
 
 # Return if requirements are not found.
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
I haven't tested it locally, but it looks like a straightforward change to me. If it does turn out to break unexpectedly should be an easy fix, so I'm 👍 as long as you've tested it locally for you.

PR #2046, opened 2023-05-05
python: Remove pip completion from prezto local

indrajitr:
pip completion is bundled with Zsh 5.9. Remove it from prezto local to avoid shadowing the bundled completion.

This is similar to the policy that [zsh-users/zsh-completions](https://github.com/zsh-users/zsh-completions/blob/master/CONTRIBUTING.md#how-to-contribute-to-zsh-completions) follows. Zsh 5.9 is about a year old now and is available in most desktop distributions. So most users with an updated environment will benefit from the updated pip completion.

_Note:_
We can probably inject a completion helper conditionally for older zsh versions, but would prefer to avoid the clutter.


jeffwidman (approved):
SGTM!

PR #2048, opened 2023-05-19
history-substring-search: add fuzzy and unique

dllud:
Adds 2 useful options available on the currently linked version (`v1.0.2`) of history-substring-search: fuzzy search and unique results (removes duplicates).

### DIFF
--- a/modules/history-substring-search/README.md
+++ b/modules/history-substring-search/README.md
@@ -65,6 +65,24 @@ _`${ZDOTDIR:-$HOME}/.zpreztorc`_:
 zstyle ':prezto:module:history-substring-search' globbing-flags ''
 ```
 
+### Fuzzy search
+
+To enable fuzzy search add the following line to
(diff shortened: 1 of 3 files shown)


indrajitr:
Nice!

indrajitr (approved):

jeffwidman:
Would it be a better out-of-the-box user experience to enable fuzzy search by default?

Unique I'm 👍 on adding the option, not so sure about enabling by default...

dllud:
> Would it be a better out-of-the-box user experience to enable fuzzy search by default?

I don't think so. It drastically changes the results you get when using spaces.
Currently users are expecting a search for `ls -` to return the previous ls commands where arguments were used. With fuzzy search enabled they would get that plus all commands where `ls` was followed by any directory that had a `-` in it's name.

indrajitr:
I am applying them as-is. Minor tweaks to follow.

PR #2063, opened 2023-10-05
Only call `pyenv init` once

johndbritton:
The `python` module calls `pyenv init` twice on each startup.

The bug was introduced here: https://github.com/sorin-ionescu/prezto/commit/afe59b293b5d91353f58a4c786a560242bb43415#diff-414db2cc4cb0c69d0fea2d6304ceb9921bf9d0d96f8d47f545e19bfa217dc39dL20

The change to the logic dropped an `else` resulting in `pyenv init` being called twice.

### DIFF
--- a/modules/python/init.zsh
+++ b/modules/python/init.zsh
@@ -29,9 +29,9 @@ if [[ -s "${local_pyenv::=${PYENV_ROOT:-$HOME/.pyenv}/bin/pyenv}" ]] \
   autoload -Uz is-at-least
   if is-at-least 2 ${"$(pyenv --version 2>&1)"[(w)2]}; then
     eval "$(pyenv init --path zsh)"
+  else
+    eval "$(pyenv init - zsh)"
   fi
-
-  eval "$(pyenv init - zsh)"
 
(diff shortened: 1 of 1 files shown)


belak:
That commit seems to imply that both are needed. @indrajitr would probably be the best person to take a look.

johndbritton:
That commit ends up loading `pyenv` twice resulting in duplicate entries in `$PATH`. I think my change is correct, but there might be another preferred way to resolve. It definitely should not be running twice though.

johndbritton:
@indrajitr Do you have any suggested changes, or do you think this is ok?

indrajitr:
@johndbritton, thanks for bringing it up. It was originally added back in #1931 to work around the quirk of pyenv (pyenv/pyenv#1906).

I think this can now be simplified assuming you have an updated pyenv installed. I'll take a look into this shortly.

johndbritton:
Great, thanks. I think you're right that it's no longer necessary.

indrajitr:
cherry-picked [here](https://github.com/sorin-ionescu/prezto/commit/457ade4f014285d47fec229c5210af752f5be7e4). Thanks @johndbritton!

PR #2064, opened 2023-10-18
add zstyle to disable alias sl=ls

abizer:
sl is the command alias for [`sapling`](https://github.com/facebook/sapling) so this alias conflicts with it. This PR adds an option to disable the alias.

## Proposed Changes
  - `zstyle -T ':prezto:module:utility' slls 'no'` disables `alias sl=ls`

PR #2066, opened 2023-11-06
remove alias sl='ls'

abizer:
replacement for https://github.com/sorin-ionescu/prezto/pull/2064 that removes this alias entirely


jeffwidman (approved):
Thanks @abizer !

This is technically a breaking change, but given that it's removing an alias intended as a typo correction, I think this is fine to ship as-is. Anyone writing code programmatically targeting typo corrections should have their head examined! 😁

jeffwidman:
Thanks @abizer !

PR #2076, opened 2024-02-28
Include Homebrew completion path

vboginskey:
## Proposed Changes

This PR adds the Homebrew completion directory to `fpath` to enable Zsh to automatically pick up completions from Homebrew-installed packages.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -14,14 +14,18 @@ fi
 # Add zsh-completions to $fpath.
 fpath=(${0:h}/external/src $fpath)
 
-# Add completion for keg-only brewed curl on macOS when available.
+# Add Homebrew completions, including keg-only brewed curl, on MacOS when available.
 if (( $+commands[brew] )); then
   brew_prefix=${HOMEBREW_PREFIX:-${HOMEBREW_REPOSITORY:-$commands[brew]:A:h:h}}
   # $HOMEBREW_PREFIX defaults to $HOMEBREW_REPOSITORY but is explicitly set to
   # /usr/local when $HOMEBREW_REPOSITORY is /usr/local/Homebrew.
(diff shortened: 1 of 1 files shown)


vboginskey:
@indrajitr thoughts?

indrajitr:
@vboginskey, apologies for the delayed response.

There are a few things to consider here:
1. Zsh `fpath` is not necessarily about completion functions _only_. They are also about other Zsh internal functions and helpers.
2. So even if we have to inject additional `fpath` members, they have to be at a more generalized place and they have to be customizable (which would be typically in [_`$ZDOTDIR/zprofile`_](https://github.com/sorin-ionescu/prezto/blob/master/runcoms/zprofile)).
3. That said, site function directories are supposed to go _hand in hand_ with the actual Zsh instance being used - so that stock Zsh functions and helpers are consistent with the Zsh instance being used.

Summary:
- If you are using macOS native instance of Zsh from `/bin/zsh`, you'd automatically have `/usr/share/zsh/site-functions` and `/usr/share/zsh/5.9/functions` in your `$fpath`.
- If you are using Homebrew instance of Zsh from `/opt/homebrew/bin/zsh`, you'd automatically have `/opt/homebrew/share/zsh/site-functions` and `/opt/homebrew/Cellar/zsh/5.9/share/zsh/functions` in your `$fpath`.
- These don't need to be specified explicitly because they are Zsh build-time config and are baked into the corresponding Zsh installation.

You have two options from here:
1. Explicitly add Homebrew zsh package function paths (`/opt/homebrew/share/zsh/site-functions` and `/opt/homebrew/Cellar/zsh/5.9/share/zsh/functions` in your `$fpath`) in your _`$ZDOTDIR/zprofile`_ and brace for some inconsistencies with built-in Zsh functions.
2. Switch your default Zsh shell to use Homebrew Zsh (from `/opt/homebrew/bin/zsh`) instead of macOS native Zsh (from `/bin/zsh`).

vboginskey:
That all makes sense, thank you for the detailed response 🙏 .

PR #2078, opened 2024-03-12
Adjust case-insensitive completion to use [:upper:] and [:lower:]

### LINKED ISSUE #2053: Completion chooses incorrect prefix among filenames
### Description

When tab-completing the name of a file with the `completion` module enabled, prezto sometimes pre-fills a prefix that is not universal among all options in the directory.

### Expected behavior

Attempting to tab-complete filenames should list and cycle through all candidate files in the directory.

### Actual behavior

Tab-completion auto-fills the first character of an arbitrary file in the directory. Pressing tab again completes the name of that file, rather than cycling through other files in the directory.

See [here](https://asciinema.org/a/597337). Notice that, despite `foo/aaaa` and `foo/bbbb` being present, attempting to autocomplete `ls foo/` first gives `foo/B` (a prefix for which *no* file exists), then `foo/bbbb`, skipping `foo/aaaa` entirely.

### Steps to Reproduce

It is inconsistent on my system (that is, it does not occur in every folder), but I can seem to reproduce it consistently when exactly two ambiguous results are present.

1. Create an empty folder `foo`
2. Create files `aaa` and `bbb` in folder
3. `ls foo/<TAB>`

I believe this is related to prezto and not completion in general because using regular tab completion with `zsh -f` correctly lists both `foo/aaa` and `foo/bbb` when prompted.

### Versions

  - Prezto commit: c9c716e9c11938a1aff2f30230d4dc1da38dc564
  - ZSH version: zsh 5.9 (x86_64-redhat-linux-gnu)
  - OS information: Fedora release 38 (thirty-eight)

CT075:
This fixes #2053.

Previously, in cases where exactly two completion arguments were present, one would match `{a-zA-Z}` and the other would match `{A-Za-z}`, thus causing a first-character "match". The [docs](https://zsh.sourceforge.io/Doc/Release/Completion-Widgets.html#Completion-Matching-Control) recommend using `[:lower:]` and `[:upper:]` over `a-z` as well.

### DIFF
--- a/modules/completion/init.zsh
+++ b/modules/completion/init.zsh
@@ -84,7 +84,7 @@ if zstyle -t ':prezto:module:completion:*' case-sensitive; then
   zstyle ':completion:*' matcher-list 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
   setopt CASE_GLOB
 else
-  zstyle ':completion:*' matcher-list 'm:{a-zA-Z}={A-Za-z}' 'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
+  zstyle ':completion:*' matcher-list 'm:{[:lower:]}={[:upper:]}' 'm:{[:upper:]}={[:lower:]}'  'r:|[._-]=* r:|=*' 'l:|=* r:|=*'
   unsetopt CASE_GLOB
 fi
 


indrajitr:
Thanks @CT075!

PR #2080, opened 2024-03-31
Update history-substring-search

hotpxl:
This updates history-substring-search module, specifically to include [this fix](https://github.com/zsh-users/zsh-history-substring-search/commit/da5f0673b9a464c702f042a8a18d8ce2d08ac6e8)

Since z-sy-h was updated and it's now using a widget, it may override history-substring-search highlight. This history-substring-search fix resolves this problem.


indrajitr:
Looks like zsh-users/zsh-history-substring-search doesn't do tagged releases anymore. We have to switch `modules/history-substring-search/external` to track zsh-users/zsh-history-substring-search@master for now because zsh-users/zsh-history-substring-search/@v1.1.0 is quite old.

PR #2083, opened 2024-04-04
Allow overriding _ssh_agent_sock for symmetry to _ssh_agent_env

hotpxl:
`_ssh_agent_env` can be overridden by another module, this change proposes the same treatment to `_ssh_agent_sock`.

Nothing is overridden by default so it shouldn't affect any existing user.

Specifically, I was looking into overriding where the SSH files are stored, without modifying `XDG_CACHE_HOME` (because a bunch of other stuff uses that), and overriding `_ssh_agent_env` and `_ssh_agent_sock` seems like the cleanest.


indrajitr:
Thank you @hotpxl for the recent changes!

hotpxl:
Thank you for the reviews!

PR #2085, opened 2024-04-05
Cache "brew shellenv" results

hotpxl:
This could sometimes be slow (as with any external command execution) but the result should almost never change.

### DIFF
--- a/modules/homebrew/init.zsh
+++ b/modules/homebrew/init.zsh
@@ -22,18 +22,15 @@ fi
 # variables as they are already handled in standard zsh configuration.
 if (( $+commands[brew] )); then
   cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/brew-shellenv-cache.zsh"
-  if [[ "${commands[brew]}" -nt "$cache_file" \
+  if [[ "$commands[brew]" -nt "$cache_file" \
       || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
-      || ( ! -n ${cache_file}(#qNmw-1) ) ]]; then
-    # Rebuild cache if either
-    # 1. brew is newer than the cache file
-    # 2. .zpreztorc is newer than the cache file
-    # 3. cache file doesn't exist
-    # 4. cache file is older than a week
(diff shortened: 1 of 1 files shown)


modules/homebrew/init.zsh
```diff
@@ -21,7 +21,20 @@ fi
 # Load 'HOMEBREW_' prefixed variables only. Avoid loading 'PATH' related
 # variables as they are already handled in standard zsh configuration.
 if (( $+commands[brew] )); then
-  eval "${(@M)${(f)"$(brew shellenv 2> /dev/null)"}:#export HOMEBREW*}"
+  cache_file="${XDG_CACHE_HOME:-$HOME/.cache}/prezto/brew-shellenv-cache.zsh"
+  if [[ "${commands[brew]}" -nt "$cache_file" \
+      || "${ZDOTDIR:-$HOME}/.zpreztorc" -nt "$cache_file" \
+      || ( ! -n ${cache_file}(#qNmw-1) ) ]]; then
```
indrajitr, line 27: This will require `EXTENDED_GLOB` to be set. This might not be the case when one enables this module. We have burned ourselves in the [past](https://github.com/sorin-ionescu/prezto/issues/1949#issuecomment-902188772) :)

Do you think we can ignore checking for the age of the `cache_file`?

hotpxl: Thanks for pointing this out. I can either do `LOCAL_OPTIONS` + `EXTENDED_GLOB`, or I can remove the check for the age. I don't really need the age myself, but I think of it as some kind of an escape hatch - just in case we missed some dependency and brew's shellenv does get updated, we'll eventually pick it up. What do you think?

indrajitr:
@hotpxl, are you okay with [this](https://github.com/sorin-ionescu/prezto/pull/2085/commits/32884e045d364cf6e43fca8f2847c4831facc50d)?

hotpxl:
> @hotpxl, are you okay with [this](https://github.com/sorin-ionescu/prezto/pull/2085/commits/32884e045d364cf6e43fca8f2847c4831facc50d)?

I'm OK with this, thanks!

PR #2086, opened 2024-04-15
Removing Fig from README.md

lnlyssg:
## Proposed Changes

Fig is sunsetting as per https://fig.io/blog/post/fig-is-sunsetting and CodeWhisperer does not support plugins so I've removed the Fig section from the Readme


indrajitr:
Thanks @lnlyssg!

PR #2090, opened 2024-06-03
Submodules Bumps

diraol:
Just some submodules bumps :)


indrajitr:
Thanks @diraol!
Regarding zsh-completion, I agree with switching to master branch.

PR #2100, opened 2024-12-15
Add Volta Support

adamavenir:


modules/node/init.zsh
```diff
@@ -64,4 +68,12 @@ if ! zstyle -t ':prezto:module:node:alias' skip; then
   alias npmci='npm ci'
   alias npmcit='npm cit'
   alias npmit='npm it'
-fi
+
+  # Add Volta-specific aliases if Volta is installed
+  if (( $+commands[volta] )); then
+    alias vl='volta list'
+    alias vla='volta list all'
+    alias vi='volta install'
```
indrajitr, line 76: Almost every *nix out there has `vi` as editor command. Also chances are many users just type `vi` without thinking and expect the editor to show up on the screen.

adamavenir: lol yes I am actually a vim user and just wasn't thinking 😆

PR #2104, opened 2025-01-17
Use @whjvenyl's fork of fasd

tagatac:
## Proposed Changes
https://github.com/clvv/fasd is archived and has not been changed in over 9 years. The fork at https://github.com/whjvenyl/fasd is responsive to PRs and, notably, supports case-insensitive matching.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -21,7 +21,7 @@
 	url = https://github.com/sindresorhus/pure.git
 [submodule "modules/fasd/external"]
 	path = modules/fasd/external
-	url = https://github.com/clvv/fasd.git
+	url = https://github.com/whjvenyl/fasd.git
 [submodule "modules/prompt/external/async"]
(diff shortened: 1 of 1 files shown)


indrajitr:
All for it in general 👍🏽 
One niggling aspect of the forked repo (the reason I have been on the fence on this PR) is the absence of any recent stable releases or tags. Following a main/master branch of a submodule is inherently less stable - we try to avoid that (although we had to end up making exceptions to some of the well-known submodules in the past).

indrajitr:
Wonderful! @tagatac would have time to update the PR with linking the submodule to the tag? I'll get to it later this week otherwise.

tagatac:
Done. I think. I updated the submodule commit to the release commit. I don't see a way to link to the tag symbolically. LMK if I'm missing something 🙏 Thanks!

PR #2117, opened 2025-11-25
Fix typo in README

vojtechsebo:


jeffwidman:
Thanks! But `sane` is the intended word here.

PR #2122, opened 2026-01-23
fix diff function: conform to GNU diff behaviour for scripting

### LINKED ISSUE #2121: [bug] diff function breaks compatibility with GNU diff
There are two major issues with this function:

- It misses the return value of the original diff. GNU diff returns value 0 for matching files and 1 for non-matching files, which enables it's use in conditional statements. The function here always returns 0.
- It's always colored even if the output is not a TTY, which leads to spectacular bugs when it's output is piped elsewhere.

lesf0:
Fixes #2121

## Proposed Changes

  - This change changes the behaviour of diff function to conform to GNU diff: return code indicates comparison result, output is not colored when used within a pipeline.

### DIFF
--- a/modules/utility/functions/diff
+++ b/modules/utility/functions/diff
@@ -10,7 +10,7 @@
 if zstyle -t ':prezto:module:utility:diff' color \
       && [[ -t 1 ]] \
       && (( $+commands[colordiff] )); then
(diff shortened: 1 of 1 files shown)


indrajitr:
Thanks @lesf0 !

PR #2142, opened 2026-04-24
python: fix venv display in pure

mandre:
The magic value changed from 12 to 20 in pure v1.27.0 via https://github.com/sindresorhus/pure/commit/987ae71a976226d2ccf01dd4df49483ef683d73awhich broke displaying the python venv.


indrajitr:
Thanks @mandre !

